Issue #3, opened 2018-02-14
NextCloud


SalocinHB:

927589452:
So I could write this for a mug?

wernermo:
We don’t have mugs. Yet. We will see what we can do.

PR #44, opened 2018-03-30
added basic ownCloud setup manual

spacecodeit:
First try to contribute to the lab.

Please note:
There is no ownCloud logo with transparent background and it is not allowed to invert the image: https://owncloud.org/trademark/

PR #45, opened 2018-03-30
#44 with feature branch (Adding Nextcloud)

spacecodeit:
I forgot to use a feature branch in my first PR. #44


luto:
I'm assuming NextCloud and OwnCloud are still incompatible, so this won't fix #3. Are we happy with OwnCloud or should there be two guides: one for NextCloud, one for OwnCloud?

luto:
Also, thank you very much for the guide! :) :tada:

spacecodeit:
I did a quick test, the steps needed for NextCloud are the same as for ownCloud.

As I know it is also possible to migrate from ownCloud to NextCloud. After a quick research I can't find anything about the other way. The documentation for ownCloud to NextCloud can be found here:
https://docs.nextcloud.com/server/13/admin_manual/maintenance/migrating_owncloud.html

tocsinDE:
> Are we happy with OwnCloud or should there be two guides: one for NextCloud, one for OwnCloud?

For me it's like Gogs vs. Gitea: NextCloud seems to be developed more actively than OwnCloud and gained a lot of traction in the past months, so +1 for NextCloud.

nichtmax:
According to our support, users ask _way more_ often for [Nextcloud](https://wiki.uberspace.de/cool:owncloud). Owncloud seems to be more enterprisey.

spacecodeit:
> According to our support, users ask way more often for Nextcloud. Owncloud seems to be more enterprisey.

If you wan't I can change my guide to Nextcloud, as I mentioned above, the steps seams to be (nearly) the same.

ezzra:
Yes, Nextcloud is much more "sexy" and community driven. But maybe there is no reason why there shouldn't be an ownCloud manual as well? 
I think in the end, there will be several differences in personalization, updates, links etc. that should be reasonable to use separate manuals and not put them together.

nichtmax:
I’ll give my 2 cents tomorrow.

nichtmax:
There is no reason to have both guides, we'd like to have NextCloud first though or both at the same time. @spacecodeit would you like to change this one to NextCloud or should we wait for the NextCloud guide or do you want to clone it and we release both?

spacecodeit:
I will change to NextCloud in the next days.

source/guide_nextcloud.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Another thing you can configure to get a better performance is enabling Memcaching.
+
+To enable Memcaching (APCu) you only need to add the following line to your /var/www/virtual/$USER/html/config/config.php:
+
+::
+
+ 'memcache.local' => '\OC\Memcache\APCu',
+
+
```
nichtmax: The admin interface complains about 
```
The "Strict-Transport-Security" HTTP header is not set to at least "15552000" seconds. For enhanced security, it is recommended to enable HSTS as described in the security tips.
```
Please add a hint that we can't change that at the moment and link to [the manual](https://manual.uberspace.de/en/web-security.html).

source/guide_nextcloud.rst
```diff
@@ -0,0 +1,114 @@
...
+============
+
+``cd`` to your `document root`_, then download the latest release of the Nextcloud and extract it:
+
+.. note:: The link to the lastest version can be found at Nextcloud's `download page <https://nextcloud.com/install/#instructions-server>`_.
+
+::
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
+ [isabell@stardust html]$ curl https://download.nextcloud.com/server/releases/nextcloud-42.23.1.tar.bz2 | tar -xjf - --strip-components=1
```
nichtmax: Can you change that to 

```
+ [isabell@stardust html]$ curl https://download.nextcloud.com/server/releases/latest.tar.bz2 | tar -xjf - --strip-components=1
```

please?

nichtmax:
Can you please add the [versions](https://github.com/Uberspace/lab/blob/master/source/README.md#boilerplate) you tested with?

PR #52, opened 2018-04-03
proxy redirect to secure https connection

ezzra:
Is there any reason, why not all the proxy redirects should be used to redirect to secure https as well? Since TLS is now available by default on U7


SalocinHB:
No need, all HTTP connections are already redirected by nginx

ezzra:
ah ok, great, its not mentioned in the manual, is it? Maybe here https://manual.uberspace.de/en/web-security.html

SalocinHB:
It’s stated in the info box at the very start of that article: “We enforce HTTPS”

PR #53, opened 2018-04-03
guide for jingo wiki

ezzra:
I don't know if you have any relevance guidelines for tools. This one is a not very much known but still used by some people according to the forks and stars on [Github](https://github.com/claudioc/jingo), I personally like to use it because of its simplicity.

### DIFF
--- a/source/guide_jingo.rst
+++ b/source/guide_jingo.rst
@@ -209,7 +209,7 @@ Be sure to set any random signs for the option ``secret`` within the ``config.ya
 Updates
 =======
 
-.. note:: Check the Github_ repository regularly to stay informed about the newest version.
+.. note:: Check the update feed_ regularly to stay informed about the newest version.
 
 
 If there is a new version available, you can get the code using Git:
@@ -254,3 +254,4 @@ In the end you need to restart the service daemon, so the new code is used by th
 .. _npm: https://manual.uberspace.de/en/lang-nodejs.html#npm
 .. _supervisord: https://manual.uberspace.de/en/daemons-supervisord.html
 .. _domains: https://manual.uberspace.de/en/web-domains.html
+.. _feed: https://github.com/claudioc/jingo/releases.atom
--- a/source/guide_jingo.rst
+++ b/source/guide_jingo.rst
@@ -129,10 +129,10 @@ For the initial config here, the most simple way is to set up a local account th
 .. code-block:: console
 
   [isabell@stardust ~]$ ~/jingo/jingo --hash-string MySuperSecretPassword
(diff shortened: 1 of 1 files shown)


source/guide_jingo.rst
```diff
@@ -0,0 +1,256 @@
...
+.. include:: includes/web-domain-list.rst
+
+Installation
+============
+
+First get the Jingo code from Github_:
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ git clone https://github.com/claudioc/jingo ~/jingo
```
ezzra: I am unsure if the git clone shut be done on master, or on the last release number like in the etherpad guide. Because here the releases are settled in tags, there would be necessary a bit more irritating git commands. 
The Jingo maintainer advises to pull from master anyway.

nichtmax: If the Jingo maintainer advises to pull from master, that's fine.

source/guide_jingo.rst
```diff
@@ -0,0 +1,257 @@
+.. highlight:: console
+
+.. sidebar:: About
+
+  This guide has been tested with Uberspace 7.1.1 and Jingo 1.8.5
```
nichtmax: Could you move that to the bottom? We updated [the rules](https://github.com/Uberspace/lab/blob/master/source/README.md). Sorry, we're still experimenting :)

source/guide_jingo.rst
```diff
@@ -0,0 +1,257 @@
...
+    baseUrl: ''
+  [...]
+
+You will also need to set up an account to login. You can choose multiple ways for authentication like *Github*, *Google*, *Ldap*.
+For the initial config here, the most simple way is to set up a local account that is stored in the config file. You first need to choose a password and create a hash:
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ ~/jingo/jingo --hash-string MySuperSecretPassword
+  1531379b51fd883c2eee31a7fa95ae9a94e0e1bf
```
nichtmax: Haven't tested it but if that's the hash for `MySuperSecretPassword` could you please replace it with random data? If you already did: nevermind ☺️ We want to make sure that the password is *not* `MySuperSecretPassword`

source/guide_jingo.rst
ezzra: ok did so, whats the best way to keep track of rule changes like this? Maybe you can set up an issue that can be subscribed?

source/guide_jingo.rst
ezzra: ok, I hope now its made really sure ;)

source/guide_jingo.rst
nichtmax: Will do.

PR #63, opened 2018-04-05
Lychee improvement

ezzra:
Pulling from Github can be done simpler. Also added security improvements to hide the .git folders.

### DIFF
--- a/source/guide_lychee.rst
+++ b/source/guide_lychee.rst
@@ -80,7 +80,7 @@ If a new version is available, ``cd`` to your Lychee folder and do a simple ``gi
 .. _Lychee: https://lychee.electerious.com/
 .. _GitHub: https://github.com/electerious/Lychee/releases
 .. _Feed: https://github.com/electerious/Lychee/releases.atom
-.. _PHP: http://www.php.net/
+.. _PHP: https://manual.uberspace.de/en/lang-php.html
 .. _credentials: https://manual.uberspace.de/en/database-mysql.html#login-credentials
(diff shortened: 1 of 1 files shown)


source/guide_lychee.rst
```diff
@@ -65,25 +61,37 @@ You will need to enter the following information:
 
 For the last step you have to enter the username/password you want to use for the Lychee user.
 
+Best practices
+==============
+
+Security
+--------
+
+To make the ``.git`` folders inaccessible from the web, add the following line to the end of the ``~/html/.htaccess`` file:
```
luto: The `.git` directory should not contain any private information. Is there any harm in exposing it?

jeyca: Indeed, its just a security measure I use to take when (php) tools do not have a specific public webfolder. Because this is a public repository the threat is tiny, just the possibility that someone accidentally commits private data, corrects the mistake but leaves it in the git history. Should not happen but I personally don't like to allow access to developing parts.

However, I can remove this commit if you want to keep it simple. Because its apart from the installation process and just for security hardening, I thought it might fit for others too.

ezzra: Indeed, its just a security measure I use to take when (php) tools do not have a specific public webfolder. Because this is a public repository the threat is tiny, just the possibility that someone accidentally commits private data, corrects the mistake but leaves it in the git history. Should not happen but I personally don't like to allow access to developing parts.

However, I can remove this commit if you want to keep it simple. Because its apart from the installation process and just for security hardening, I thought it might fit for others too.

luto:
Thank you! :)

PR #76, opened 2018-04-13
Add notice to Gitea about pubkey ssh login.

M4a1x:
Public key logins via ssh will get rejected if you don't use the
built-in ssh server. Putty fails with cryptic "Server refused to
allocate pty". Password login still works as expected.

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -152,8 +152,7 @@ Finishing installation
 
 Point your browser to your gitea URL and finish the installation.
 
-.. note:: Gitea prevents you from logging in with private keys if you
- don't use the built-in ssh server. Login via password still works as expected.
+.. note:: Using the same public key for Gitea and your Uberspace will prevent you from logging in with this key via SSH. Login via password still works as expected, you can use the Dashboard_ to manage your keys.
 
 Updates
 =======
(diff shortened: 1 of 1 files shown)


SalocinHB:
I'm sure you can break a lot of things by using settings other than the ones recommended in this guide. I don't think there's a need to mention every possible config option that could cause troubles. Stick to the guide and it'll work.

luto:
That's only partly true. if you use the same SSH-Key for git and normal SSH key, gitea is likely to lock you out, yes. That's not a deviation from the default config or this guide.

source/guide_gitea.rst
```diff
@@ -152,6 +152,9 @@ Finishing installation
 
 Point your browser to your gitea URL and finish the installation.
 
+.. note:: Gitea prevents you from logging in with private keys if you
+ don't use the built-in ssh server. Login via password still works as expected.
```
luto: Opening ports is not yet possible on uberspace 7, so the built-in server is not an option. I think we should nevertheless warn people about using the same key for SSH and git. Enabling both ssh and git using the same key would be awesome, but I don't think that's doable.

M4a1x: Ah so that's the issue. I just followed the guide, added my ssh key (same as Uberspace) to Gitea and was locked out. I spent some time figuring out why putty wouldn't let me in. That's why I thought I'd update this notice.

So maybe something more along the lines of this?
```
Using the same public key for Gitea and your Uberspace will prevent you 
from logging in with this key via SSH. Login via password still works as expected. 
```

SalocinHB: Maybe include a hint that `authorized_keys` can be edited via the Dashboard if this happens?

nichtmax: > Using the same public key for Gitea and your Uberspace will prevent you from logging in with this key via SSH. Login via password still works as expected, you can use the [Dashboard](https://uberspace.de/dashboard/authentication) to manage your keys.

PR #86, opened 2018-04-20
Small Nextcloud update

splitt3r:
I updated the awesome tutorial by @spacecodeit so the logo fits in and there was one wrong Nextcloud wording left.

**Before**:

before

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -9,7 +9,7 @@
 Nextcloud
 #########
 
-Nextcloud_ is an open source cloud solution written in PHP and distributed unter the AGPLv3 licence.
+Nextcloud_ is an open source cloud solution written in PHP and distributed unter the AGPLv3 licence. You can find further infromation on how to secure and optimise your setup at the offical `docs page <https://docs.nextcloud.com>`_.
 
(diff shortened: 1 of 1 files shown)


splitt3r:
I think it would also be a great addition to give a hint about `'htaccess.IgnoreFrontController' => false,`. I think a lot of people change that setting.

source/guide_nextcloud.rst
```diff
@@ -9,7 +9,7 @@
 Nextcloud
 #########
 
-Nextcloud_ is an open source cloud solution written in PHP and distributed unter the AGPLv3 licence.
+Nextcloud_ is an open source cloud solution written in PHP and distributed unter the AGPLv3 licence. You can find further infromation on how to secure and optimise your setup at the offical `docs page <https://docs.nextcloud.com>`_.
```
SalocinHB: Good idea, but the link should probably be at the end of the article

splitt3r: Ok. I will have a look at resizing the logo so this text can be moved to the end oft the tutorial.

luto: the logo overlapping the box below is a general problem, which shouldn't be solved on a per-guide basis, imho.

luto:
I included the Next**C**loud fix manually. Since the main purpose of the additional link seems to be to fix the logo, I'm closing this for now. See #89.

Issue #89, opened 2018-04-21
logo overlaps content, if first paragraph is too short


luto:
![image](https://user-images.githubusercontent.com/2158203/39085683-a295ba66-4586-11e8-83af-40562dc0df12.png)

See https://github.com/Uberspace/lab/pull/86, https://github.com/Uberspace/lab/commit/cb4a88064d8cba553de50c7cc697e9d6aeee4168

nichtmax:
Does not seem to work. Did you check before merging @SalocinHB?

SalocinHB:
Of course. Works fine here.

nichtmax:
@brutus Seems to work only sometimes, sometimes after a reload the Script does not work anymore.
![pasted_image_at_2018_05_03_02_55_pm](https://user-images.githubusercontent.com/988418/39577857-bf503054-4ee2-11e8-877e-42bbcd58dcb5.png)

brutus:
Works here too… but I can make it bail, if I force reload the site (CTRL+SHIFT+R). After a _normal_ reload (CTRL+R) it works again :man_shrugging:

nichtmax:
Other way around here with Safari 😁

nichtmax:
fixed 💪

PR #101, opened 2018-04-26
Feature/fix overlapping logo

brutus:
Add some margin to the first HR to push content down further, if the logo laps over that content.

Aims to close: https://github.com/Uberspace/lab/issues/89

### DIFF
--- a/source/_static/js/custom.js
+++ b/source/_static/js/custom.js
@@ -1,0 +1,32 @@
+var app = {
+
+    check_logo_height: function() {
+        var logo = $('div[itemprop="articleBody"] > .sidebar')
+        var hr = $('div[itemprop="articleBody"] > .section > hr')
+        var logo_end = 0
+        var hr_position = 0
+        if (logo.length) {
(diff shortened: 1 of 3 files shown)


source/_static/js/custom.js
```diff
@@ -0,0 +1,31 @@
...
+        if (logo_end > hr_position) {
+            var margin = parseInt(hr.css('margin-bottom'), 10)
+            var offset = Math.ceil(logo_end - hr_position)
+            var new_margin = margin + offset
+            console.log( '[logo] pushing first break down to ' + margin + ' + ' + offset + ' → ' + new_margin);
+            hr.css('margin-bottom', new_margin + 'px');
+        }
+    },
+
+    inti: function() {
```
luto: init?

brutus: I put the `check_logo_height` function in an `app` object… to preserve some global namespace. If needed, new functions and other stuff can be added to the `app` object.  Calling `app.init()` does all the stuff that should happens automatically on each page load.

PR #106, opened 2018-05-03
wrap logo–check in jQuery ready function

brutus:
I can't reproduce https://github.com/Uberspace/lab/issues/89 locally, but it seems, that the function call fails (since `UberLab launched…` is always printed to the console). So… I wrapped the check in a ready function :crossed_fingers:


nichtmax:
This does the trick. Thanks @brutus ✌️

PR #112, opened 2018-05-08
Fix Nextcloud cron script path and log script output

coderkun:
The current guide for Nextcloud uses a wrong path for the cron script under `$HOME` instead of the installation directory `/var/www/virtual/$USER/html/`. Commit 654abaa fixes this.

Additionally commit dd70a51 logs the output of the cron script to `$HOME/logs/nextcloud-cron.log` which might be useful for maintenance and debugging.

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -76,7 +76,7 @@ Add the following cronjob to your crontab_:
 
 ::
 
- *  *  *  *  * php -f $HOME/html/cron.php
+ *  *  *  *  * php -f /var/www/virtual/$USER/html/cron.php
 
 Memcaching
 ----------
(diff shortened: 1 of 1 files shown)


SalocinHB:
Hmm, the path isn't wrong. $HOME/html is by default a symlink to /var/www/virtual/$USER/html. These guides always assume a fresh Uberspace.

SalocinHB:
You're right, that doesn't hurt. 

Another thing: Is there actually any output from cron.php? I can't remember ever receiving a mail from cron related to this cronjob …

coderkun:
That is a good question. It looks like the script itself does not log anything. The only log output would be errors then if the script does not run correctly.

PR #114, opened 2018-05-08
Update guide_nextcloud.rst

dertuxmalwieder:
Gotcha ... gefunden beim Basteln mit nextCloud auf einem eigenen Server.


SalocinHB:
`Zend OPcache` is always loaded by deafult

dertuxmalwieder:
nextCloud does not care, it specifically wants the opcache module. I tried that.

See also: https://github.com/nextcloud/server/issues/9410

dertuxmalwieder:
Ah. Embarrassing typo. m(

PR #120, opened 2018-05-11
Added guide for Matomo

GodMod:
A guide for the Analytics Tool Matomo (former known as Piwik).


source/guide_matomo.rst
```diff
@@ -0,0 +1,107 @@
...
+.. sidebar:: Logo
+
+  .. image:: _static/images/matomo.png
+      :align: center
+
+#########
+Matomo
+#########
+
+Matomo_ (former known as Piwik) is an open source website tracking tool (like Google Analytics) written in PHP and distributed under the GNU General Public License v3.0 licence.
```
nichtmax: Maybe describe in one or two sentences why it makes sense to do the tracking by oneself and not by a third party. GDPR comes to mind.

source/guide_matomo.rst
```diff
@@ -0,0 +1,107 @@
...
+ [isabell@stardust ~]$ cd ~/html/
+ [isabell@stardust ~]$ mkdir matomo
+ [isabell@stardust ~]$ cd matomo/
+ [isabell@stardust matomo]$
+
+If you want to install matomo into your `document root`_, just navigate with ``cd`` to your `document root`_.
+
+Now download the latest version and extract it:
+
+.. note:: The link to the lastest version can be found at Matomo's `download page <https://matomo.org/download/>`_.
```
nichtmax: The link is always the same so it makes no sense to send the user to the download page. No need to mention it then ☺️

source/guide_matomo.rst
```diff
@@ -0,0 +1,107 @@
...
+ [isabell@stardust matomo]$
+
+This will create a ``piwik`` folder containing the files and directories. Now we will move the files from the ``piwik`` folder to the parent folder.
+
+::
+
+ [isabell@stardust matomo]$ cd piwik/ && mv * .. && cd .. && rm piwik -rf
+ [isabell@stardust matomo]$
+
+Now point your browser to your Matomo URL and follow the instructions of the Installer.
```
nichtmax: Please write what the URL would be in that example. I think it should be https://isabell.uber.space/matomo

source/guide_matomo.rst
```diff
@@ -0,0 +1,107 @@
...
+
+You will need to enter the following information:
+  * your MySQL hostname, username and password: the hostname is ``localhost`` and you should know your MySQL credentials_ by now. If you don't, start reading again at the top.
+  * your Matomo database name: we suggest you use an additional_ database. For example: isabell_matomo
+  * Administrator (*Super User*) username and password: choose a username (maybe not *admin*) and a strong password for the super user
+  * Name and URL of the first website you want to track with Matomo (more can be added after installation)
+
```
nichtmax: I'm missing information about how to integrate Matomo so it can track. I think it would be helpful to mention the need to implement JavaScript Tracking Code before the closing </head> tag.

source/guide_matomo.rst
```diff
@@ -0,0 +1,107 @@
...
+You will need to enter the following information:
+  * your MySQL hostname, username and password: the hostname is ``localhost`` and you should know your MySQL credentials_ by now. If you don't, start reading again at the top.
+  * your Matomo database name: we suggest you use an additional_ database. For example: isabell_matomo
+  * Administrator (*Super User*) username and password: choose a username (maybe not *admin*) and a strong password for the super user
+  * Name and URL of the first website you want to track with Matomo (more can be added after installation)
+  
+Privacy
+=======
+Matomo can be configured to ensure that users' privacy is respected. This is required in some countries. A detailed guide for configuring privacy settings in Matomo can be found `here <https://matomo.org/docs/privacy/>`_.
```
nichtmax: Matomo respects "Do not track" settings by default and IP addresses are shortened by us anyway. I don't think it's necessary to mention privacy settings because there are very sane defaults.

GodMod: I rewrote this section. I tell the users that there are no additional settings needed, so they hopefully won't ask the uberspace support. I also added hints to the privacy policy and user opt-out. Therefore, I would recommend to keep this section in the new form.

source/guide_matomo.rst
GodMod: Added.

source/guide_matomo.rst
GodMod: Removed.

source/guide_matomo.rst
GodMod: Added.

source/guide_matomo.rst
GodMod: Added section for the tracking.

GodMod:
Added requested changes

PR #137, opened 2018-05-31
Add guide to install Tiny Tiny RSS

tobidope:


nichtmax:
We're not so happy about the ttrss authors opinion about [gas chambers](https://discourse.tt-rss.org/t/a-category-named-gas-chamber/649)... That's the reason why we removed and links from [the wiki](https://wiki.uberspace.de/cool). What do you think? We've got a responsibility here since we moderate the lab...

tobidope:
I think the creator is some kind of a troll, but the software is quite useful and doesn’t have any offensive parts in my opinion. It’s your business, so you decide.

source/guide_ttrss.rst
```diff
@@ -0,0 +1,200 @@
...
+Clone the Tiny Tiny RSS repository
+----------------------------------
+
+As Tiny Tiny RSS uses a rolling release model with the git master branch as current stable version, we
+wil use git to download the software. So go to the webroot and clone Tiny Tiny RSS into the subfolder tt-rss
+
+::
+
+ [isabell@stardust ~]$ cd ~/html
+ [isabell@stardust ~]$ git clone https://tt-rss.org/git/tt-rss.git tt-rss
```
nichtmax: Please don't use subfolders. Always use the standard document root ~/html.

source/guide_ttrss.rst
```diff
@@ -0,0 +1,200 @@
...
+=============
+
+Setup your database
+-------------------
+
+It's a good idea to create the tables under a separate database.
+
+::
+
+ [isabell@stardust ~]$ mysql -e "CREATE DATABASE isabell_tt_rss"
```
nichtmax: `mysql -e "CREATE DATABASE <username>_tt_rss"` would be better imho

source/guide_ttrss.rst
```diff
@@ -0,0 +1,200 @@
...
+
+Replace <isabell> with your own user. 
+
+Configure database and domain
+-----------------------------
+
+Now you should reach your installation under the URL ``https://<username>.uber.space/tt-rss`` (would be ``https://isabell.uber.space/tt-rss`` in our example).
+Open the URL and set the Database settings to
+
+ * Database Type: MySQL
```
nichtmax: In other guides that's: "your MySQL hostname, username and password: the hostname is localhost and you should know your MySQL credentials by now. If you don’t, start reading again at the top." Can you change that for consistency?

source/guide_ttrss.rst
```diff
@@ -0,0 +1,200 @@
...
+
+  * User: admin
+  * Password: password
+
+You will be greeted by a dialog to change the default password. Do that and optionally add one or more users with restricted rights
+for capturing and reading RSS feeds.
+
+Setup daemon to fetch RSS feeds
+-------------------------------
+Create ``~/etc/services.d/ttrss.ini`` with the following content:
```
nichtmax: I think it would be better to give a general example and then specify it with your example like [here](file:///Users/mo/Downloads/lab/build/html/guide_ghost.html#setup-daemon)

source/guide_ttrss.rst
```diff
@@ -0,0 +1,202 @@
...
+::
+
+ [isabell@stardust ~]$ mysql -e "CREATE DATABASE <username>_tt_rss"
+
+Replace <username> with your own user. 
+
+Configure database and domain
+-----------------------------
+
+Now you should reach your installation under the URL ``https://<username>.uber.space/tt-rss`` (would be ``https://isabell.uber.space/tt-rss`` in our example).
```
nichtmax: Not quite 😁 Please change to https://<username>.uber.space

nichtmax:
Hey @tobidope! Just letting you know that we removed the tt-rss guide because we have alternatives now. You'll get a Uberspace cup anyway as soon as we have them.

PR #143, opened 2018-06-05
Rename sectionauthor variable to author in Kirby guide

danielkratz:
Sorry, I also missed that you renamed the sectionauthor variable to author. Hadn't tested it because of my outdated local build environment. Sorry about that, updated everything. 🙈


luto:
;)

PR #148, opened 2018-06-08
We just tested it wordpress 4.9.6

JMoVS:
I don't know if this is worth updating for minor point releases but for what it's worth, we just tested it with 4.9.6.


luto:
yay! thanks :)

PR #158, opened 2018-07-01
Gitea: Bump Version and add SSH key management through gitea

damadmai:
`supervisorctl stop gitea` is needed to be able to overwrite the executable.


luto:
thank you!

PR #161, opened 2018-07-05
Add information for syncing with macOS

simp42:
At least starting with mac OS El Capitan the syncing with CardDAV and CalDAV
seem to require specific URLs.
See also: https://github.com/sabre-io/Baikal/issues/401

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -85,7 +85,7 @@ Since Gitea uses its own webserver, you need to find a free port and bind your a
 Change the configuration
 ------------------------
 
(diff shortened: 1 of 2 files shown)


SalocinHB:
I'm pretty sure these redirects are already present in the default .htaccess, is that why you closed the PR?

simp42:
Correct. The redirects are only necessary when you don't run NextCloud in the root directory. Unfortunately I only saw the fact that you have to write the documentation as if the application was in the root directory a little too late. If you install the application in the docroot the PR is unnecessary and thus I closed it.

PR #166, opened 2018-08-06
Added fix for update failure under subdomains

Reflic:
Yesterday the nextcloud update from 13.0.5 to 13.0.6 failed due to wrong ownerships and rights on all files.
I think this was caused by the installation into a subdomain folder.

Thanks to Matthias for knowing how to solve the error and in order to reduce support effort i thougt it would be nice to document it.

If anything is wrong or should be enhanced, just tell me!

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -132,6 +132,15 @@ Updates
 
 The easiest way to update Nextcloud is to use the web updater provided in the admin section of the Web Interface.
 
+If you have installed Nextcloud on a subdomain it can happen that the update fails: Access to the UI is not possible and HTTP 403 errors are thrown.
+In most cases this happens due to wrong owner rights which can be fixed with finishing the update via console and fixing the rights.
+::
+ 
+ [isaball@stardust ~]$ cd /var/www/virtual/$USER/html
+ [isaball@stardust ~]$ php occ upgrade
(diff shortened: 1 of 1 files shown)


tocsinDE:
[Uberspace 7.1.10 introduced SELinux](https://manual.uberspace.de/en/changelog.html#v7-1-10). This could be a reason as well. I haven't had issues with older updates in the past weeks.

luto:
yep, that's precisely the reason and this PR introduces a call to `restorecon`, which fixes the SELinux labels. Sadly, more tools than we are expected do not handle this correctly.

source/guide_nextcloud.rst
```diff
@@ -132,6 +132,15 @@ Updates
 
 The easiest way to update Nextcloud is to use the web updater provided in the admin section of the Web Interface.
 
+If you have installed Nextcloud on a subdomain it can happen that the update fails: Access to the UI is not possible and HTTP 403 errors are thrown.
+In most cases this happens due to wrong owner rights which can be fixed with finishing the update via console and fixing the rights.
```
luto: The wording is a little off here. The ownership rights are perfectly fine, bug the [SELinux security labels](https://wiki.gentoo.org/wiki/SELinux/Labels#Introduction) are wrong. The `restorecon` makes sure the labels are set as they are defined in the loaded selinux policy.

Issue #174, opened 2018-08-21
EditorConfig: "yes" is not a valid value for trim_trailing_whitespace


tstahn:
**Describe the bug**

When editing files in PhpStorm 2018.2.1 with plugin EditorConfig v182.2949.6 enabled, the following error is logged in the event log.

`EditorConfig: "yes" is not a valid value for trim_trailing_whitespace for file <file>`

This is caused by following settings in the `.editorconfig` file which d o not support a "yes" value, but require a boolean `true/false` instead:
```
insert_final_newline = yes
trim_trailing_whitespace = yes
```
See https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties.

**To Reproduce**

1. Install recent stable PhpStorm 2018.2.1
2. Install EditorConfig plugin from Settings -> Plugins
3. View event log

**Expected behavior**

No entries in the PhpStorm event log.

PR #176, opened 2018-08-21
[BUGFIX] Correct settings in .editorconfig

tstahn:
According to EditorConfig specifications (see https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties)
the value yes is not supported and must be true instead.

Resolves: #174


luto:
thanks :)

tstahn:
You're welcome.

PR #194, opened 2018-09-24
Fix wrong shape-outside for atom image

ManuelHu:
The atom image on the front page is rendered in front of the text because the file specified as shape-outside does not exist. Adding `/en/` as url prefix resolves this issue.

Before:
lab

After:
grafik

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -81,7 +81,7 @@ nav li {
     float: right;
     width: 30%;
     height: 30%;
-    shape-outside: url(/_images/atom.svg);
+    shape-outside: url(/en/_images/atom.svg);
     shape-margin: 1em;
 }
 


nichtmax:
Tested in Safari and Chrome, mobile and desktop. Rendered correct in every case...

luto:
I am pretty sure @ManuelHu is testing locally.

ManuelHu:
I was testing the live site, Firefox 62 both on desktop and mobile shows this bug. Actually, in the [manual](https://manual.uberspace.de/en/) the shape-outside url is prefixed with `/en/` and everything works as expected:

grafik
(Firefox 62 desktop)

brutus:
I too get no display errors, but @ManuelHu is right, the _url_ in the CSS seems wrong. And also failed to resolve. No idea right now, why it works in some cases :man_shrugging: 

Imo it should be either absolute — `shape-outside: url(/en/_images/atom.svg);` — or relative `shape-outside: url(../images/atom.svg);`. Since we deploy directly to `en/`, I think this PR is good. :pray:

PR #198, opened 2018-09-30
Guide for Puma

Theome:


source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+Update the permissions so that the application file is executable:
+
+::
+
+  [isabell@stardust ~]$ chmod +x ~/rubyapp/rubyapp.ru 
+  [isabell@stardust ~]$ 
+  
+This file is the Rackup file for the application. Add the following content to ``~/rubyapp/rubyapp.ru``, a simple Hello World application:
+
+.. code-block:: none
```
nichtmax: use `code-block:: ruby` instead.

source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+Next, create a Puma configuration file, update the permissions as before,
+
+::
+
+  [isabell@stardust ~]$ touch ~/rubyapp/config.rb
+  [isabell@stardust ~]$ chmod +x ~/rubyapp/config.rb
+  [isabell@stardust ~]$
+
+and add the following content:
+
```
nichtmax: There is an example code block in [the guide](https://github.com/Uberspace/lab/blob/master/STYLE.md#rules-for-guides). Can you please update this accordingly? First the general snippet, then the values according to your example. Users will just copy paste your code otherwise without changing values.

```
Create ``~/etc/services.d/ghost.ini`` with the following content:

.. warning:: Replace ``<username>`` with your username!

.. code-block:: ini

 [program:ghost]
 directory=/home/<username>/ghost
 command=env NODE_ENV=production /bin/node current/index.js

In our example this would be:

.. code-block:: ini

 [program:ghost]
 directory=/home/isabell/ghost
 command=env NODE_ENV=production /bin/node current/index.js
```

source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+  rackup '/home/isabell/rubyapp/rubyapp.ru'
+
+  # Bind Puma to the port
+  bind 'tcp://127.0.0.1:9000'
+
+
+Setup supervisord
+-----------------
+
+Create ``~/etc/services.d/rubyapp.ini`` with the following content:
```
nichtmax: same here.

source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+Setup .htaccess
+---------------
+
+Create the file ``~/html/.htaccess`` with the following content to forward requests from the outside to Puma:
+
+.. code-block:: none
+  DirectoryIndex disabled
+
+  RewriteEngine On
+  RewriteRule ^rubyapp/(.*) http://localhost:9000/$1 [P]
```
nichtmax: Don't mention additional document roots. Keep it simple. Don't use subfolders. Always use the standard document root ~/html. Always assume the document root is empty.

source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+  [isabell@stardust puma]$ supervisorctl restart rubyapp
+  rubyapp: stopped
+  rubyapp: started
+  [isabell@stardust puma]$ 
+
+
+Test
+----
+
+To validate the installation, open a browser and visit ``isabell.uber.space/rubyapp/``. If everything is set up correctly, you will see "Hello World".
```
nichtmax: Don't mention additional document roots. Keep it simple. Don't use subfolders. Always use the standard document root ~/html. Always assume the document root is empty.

source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+
+::
+
+  [isabell@stardust ~]$ mkdir /home/isabell/logs/puma
+  [isabell@stardust ~]$
+
+Edit the configuration file at ``~/rubyapp/config.ru`` and append the following line to redirect Puma's output and errors to log files:
+
+.. code-block:: none
+  stdout_redirect '/home/isabell/logs/puma/out.log', '/home/isabell/logs/puma/err.log', true
```
nichtmax: That's cool!

source/guide_puma.rst
```diff
@@ -0,0 +1,197 @@
...
+  [isabell@stardust ~]$ supervisorctl reread
+  rubyapp: available
+  [isabell@stardust ~]$ supervisorctl update
+  rubyapp: added process group
+  [isabell@stardust ~]$ 
+
+Setup .htaccess
+---------------
+
+Create the file ``~/html/.htaccess`` with the following content to forward requests from the outside to Puma:
```
nichtmax: Users will just copy and paste your code. Mention that the port needs to be changed and work with an example as above.

source/guide_puma.rst
nichtmax: Explicitly mention what needs to be replaced (the port)

nichtmax:
Can you please rebase on our master branch? The CI will then generate a preview from your branch which makes it much easier for us to test your guide.

Theome:
I will incorporate the changes you suggested, but want to follow up on one of the issues first:

You suggest that some snippets are shown first with generic placeholders like `<username>` and the again with an example like `isabell`. I don't think that this makes the tutorials easier to read or understand. On the contrary, having the code snippets twice inflates the length of the tutorial, making it more difficult to skim. 

In general, I agree that code snippets should be copy/pastable if possible, but if I have two snippets, one with `isabell` and the other with `<username>`, neither is copy/pastable, you have to edit both to make them work, so I don't see any advantage. If we follow this structure, what we actually do is explain the users how placeholders work over and over again. Given that "Uberspace is a hosting platform targeted at people who (...) generally prefer working with a text-based console", I would assume that the audience knows how placeholders work and prefers terse, easy to skim texts, i.e. tutorials where each snippet is shown only once. 

More concretely, I want to propose prepending the tutorial with a preamble instead, like:

> Throughout this tutorial, we use placeholders in example code, for example `isabell` as a placeholder for the username, and `9000` as a placeholder for a port. Make sure that you replace these placeholders with the correct values.

What do you think?

nichtmax:
Thanks for the feedback! You’re probably right. I need to think about it and I’ll be in touch.

nichtmax:
Hey @Theome! I'm really sorry for the delay, I hope you're not angry this took forever... :/ You suggest a change in the Guidelines and therefore in the long run a change in *every guide* since we want to stay consistent. We talked about this in the team and I should have kept you in the loop, my sincere apologies for not responding sooner.

Here are our thoughts: You're right, your change would make the guides simpler and we'll happily accept your guide the way it is.

One small detail though that has nothing to do with the username: [the `/rubyapp` path](https://github.com/Uberspace/lab/pull/198#discussion_r222583437). Is it okay for you if we change your guide accordingly or do you want to commit this change?

Theome:
Thanks for following up. I'm glad I could contribute to the lab. Sure, feel free to make changes to the guide as you see fit.

nichtmax:
@Theome I simplified a few things and merged the guide in https://github.com/Uberspace/lab/pull/279

Thank you for your participation!

PR #219, opened 2018-10-12
fix(guide-gitlab-runner): update binary paths

pascaliske:
unified usage of paths to binary files

btw: thanks for the guide, @bytekeks! 🙂


SalocinHB:
By default `$HOME/bin` is included in `$PATH`, so you can just use `gitlab-runner` on the shell

PR #225, opened 2018-10-17
[matomo] add auto-archiving option

Stunkymonkey:
i am not sure if using `/usr/bin/php71`hardcoded is a good choice.


SalocinHB:
> i am not sure if using `/usr/bin/php71`hardcoded is a good choice

It isn‘t. Please use `/usr/bin/php`.

PR #233, opened 2018-10-25
Add guide for .NET Core

jlnostr:
Based on my [tweet](https://twitter.com/jlnostr/status/1054852386659934214) and the [documentation of Microsoft](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-apache?view=aspnetcore-2.1&tabs=aspnetcore2x#configure-a-proxy-server), I've added some additional configuration steps for ASP.NET Core apps.

In general, users won't need these changes, as everything should work in most cases. However, there are certain things like the authentication, where setting the `X-Forwarded-...` headers are necessary.


luto:
Thanks for your pull request :)

@nichtmax there is a manual issue with similar changes. Should this go into the lab or the manual?

nichtmax:
@luto you’re right, that’s a candidate for [caveats](https://manual.uberspace.de/en/lang-dotnet.html#caveats).

@jlnostr would it be okay for you if we put that into the manual? You’ll miss out the hall of fame but of course get a cup as soon as they’re ready.

jlnostr:
@nichtmax I'm fine with that 😄

PR #241, opened 2018-11-02
Add guide to install PostgreSQL

maitrenageur:
I created a guide to install PostgreSQL on an Uberspace 7 instance.


luto:
@maitrenageur thank you very much :) the guide looks very well structured, contains just the right amount of additional explanations and is generally  awesome! I am looking forward to reading more from you, if you'd like.

fahrplaner:
"the port number with the one you wrote down earlier and " doesn't quite fit, does it?

maitrenageur:
I have seen, that the complete text block, which described to find a new own port number for the service is not more available. Is this a mistake, or is an own port number for my Uberspace not more necessary?

SalocinHB:
It’s no longer necessary, every Uberspace now gets its own network namespace

SalocinHB:
See https://manual.uberspace.de/background-network.html for details

PR #242, opened 2018-11-02
Re-sort port find to Prerequisites; Add redirect for tmp_cache

mrmk85:
Re-sort port find to Prerequisites
Add redirect for tmp_cache; Avoid conflict with other users on server

### DIFF
--- a/source/guide_seafile.rst
+++ b/source/guide_seafile.rst
@@ -49,6 +49,21 @@ Your URL needs to be setup:
  isabell.uber.space
  [isabell@stardust ~]$
 
+Configure port
+--------------
+
+Since seafile uses its own webserver and fileserver, you need to find 2 free ports. Run this command 2 times and write down the 2 ports.
+
+.. code-block:: console
+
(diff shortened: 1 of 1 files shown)


source/guide_seafile.rst
```diff
@@ -192,15 +192,29 @@ Change seahub config; Edit ``~/seafile/conf/seahub_settings.py`` and  add the fo
...
-
+  
   SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
-
+   
   FILE_SERVER_ROOT = SITE_BASE + '/seafhttp'
   CSRF_TRUSTED_ORIGINS = [SITE_NAME]
+  
+  #redirect to $USER tmp, avoid conflict with other users
+  CACHE_DIR = "/home/isabell/seafile/tmp/logs"
```
luto:
```suggestion
  CACHE_DIR = os.path.expanduser("~/seafile/tmp")
```

1. Why `.../logs`? It seems like only caching data will be stored there, no logs.
2. It is common to use `os.path.expanduser()` in django configs to make them more portable. Be sure to import it, though.

PR #248, opened 2018-11-07
pretix - credit to earlier guide author

ezzra:
forgot to reference the guide that helped me here


luto:
since you're fixing authors: I noticed that the [hall of fame](https://lab.uberspace.de/en/hall_of_fame.html) shows you twice now: Ezra and ezzra. Do you mind merging those two? :)

ezzra:
ah right, I am gonna fix this for all my guides in a new pull request, because it should actually be "ezra" :) But Iam gonna do it based on this branch here, so you could merge it first

PR #255, opened 2018-11-12
add guide for up1

ezzra:


source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+.. include:: includes/web-domain-list.rst
+
+Installation
+============
+
+First get the Up1 source code from Github_:
+
+::
+
+  [isabell@stardust ~]$ git clone https://github.com/Upload/Up1 up1 --depth 1
```
SalocinHB: are you sure about `--depth 1`?

source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+  Unpacking objects: 100% (37/37), done.
+  Note: checking out '059f199a6f94fc6c90cde1e4dab0b5b85b49533c'.
+  [isabell@stardust ~]$
+
+You have now downloaded the master development branch, because this might behave like a beta version of the software, you should check out the latest release. Please be sure to replace the pseudo version number ``66.6`` here with the latest version number from the release feed_:
+
+::
+
+  [isabell@stardust ~]$ cd ~/up1
+  [isabell@stardust up1]$ git checkout v66.6 --depth1
```
SalocinHB: You probably meant `git checkout tags/v1.0`, but that won't work either if repo was cloned with `--depth 1` (see above)

source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+  [isabell@stardust server]$ 
+
+You probably got the error that there are vulnerabilities to fix, so follow the instruction and run ``npm audit fix`` to install the latest package versions:
+
+::
+
+  [isabell@stardust ~]$ cd ~/up1/server
+  [isabell@stardust server]$ npm audit fix
+  [...]
+  fixed 17 of 17 vulnerabilities in 219 scanned packages
```
SalocinHB:
```
[isabell@stardust server]$ npm audit fix

Usage: npm <command>

where <command> is one of:
    access, adduser, bin, bugs, c, cache, completion, config,
    ddp, dedupe, deprecate, dist-tag, docs, doctor, edit,
    explore, get, help, help-search, i, init, install,
    install-test, it, link, list, ln, login, logout, ls,
    outdated, owner, pack, ping, prefix, profile, prune,
    publish, rb, rebuild, repo, restart, root, run, run-script,
    s, se, search, set, shrinkwrap, star, stars, start, stop, t,
    team, test, token, tst, un, uninstall, unpublish, unstar,
    up, update, v, version, view, whoami

npm <command> -h     quick help on <command>
npm -l           display full usage info
npm help <term>  search for help on <term>
npm help npm     involved overview

Specify configs in the ini-formatted file:
    /home/isabell/.npmrc
or on the command line via: npm <command> --key value
Config info can be viewed via: npm help config

npm@5.6.0 /opt/nodejs8/lib/node_modules/npm

Did you mean this?
    edit
[isabell@stardust server]$
```

source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+  [isabell@stardust up1]$ cp client/config.js.example server/config.js
+  [isabell@stardust up1]$
+
+Now edit ``~/up1/server/server.conf`` and replace ``<yourport>`` with the fresh portnumber you got earlier. Also set a random string for ``api_key`` and ``delete_key``:
+
+.. code-block:: none
+ :emphasize-lines: 2,3,13,18
+
+  {
+  "api_key": "<any_random_string>",
```
SalocinHB: Maybe use something like `< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c${1:-32};echo;` to generate the random strings?

source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+  },
+
+  "http": {
+    "enabled": true,
+    "listen": "localhost:<yourport>"
+  },
+
+  "https": {
+    "enabled": false,
+    "listen": "localhost:<yourport>",
```
SalocinHB: Not necessary. Have a closer look at the line directly above 😉

source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+Change the configuration
+------------------------
+
+Copy the example settings files to create your own configuration:
+
+::
+
+  [isabell@stardust ~]$ cd ~/up1
+  [isabell@stardust up1]$ cp server/server.conf.example server/server.conf
+  [isabell@stardust up1]$ cp client/config.js.example server/config.js
```
SalocinHB: Shouldn't this be `cp client/config.js.example client/config.js`?

source/guide_up1.rst
```diff
@@ -0,0 +1,217 @@
...
+Setup daemon
+------------
+
+Create ``~/etc/services.d/up1.ini`` with the following content:
+
+.. warning:: Replace ``<username>`` with your username!
+
+.. code-block:: ini
+
+ [program:up1]
```
SalocinHB: Doesn't work, working directory needs to be the one `server.js` is in. Add
```
directory=/home/<username>/up1/server/
```

source/guide_up1.rst
ezzra: ah yes correct, thats the result of a dirty change in the clone/checkout process :P 
Earlier I did the checkout already together with the clone command, but thought its easier to understand when I  split these two actions. But then it wont work with `clone --depth 1`. But `checkout v66.6` actually works, git can handle this. But of course its better to be verbose in the guide, so I will adapt this.

source/guide_up1.rst
ezzra: Oh, first I thought its a mistake because you did not follow the guide and updatet to `npm 6.4.0`, but it looks like that the update does not refresh the current npm version. You have to start a new ssh session to use the updated npm version. 

Do you have any idea how to make the npm update effective without restarting the terminal?

source/guide_up1.rst
ezzra: I was not sure if the guide should tell the people where they get there random strings from? If you think its ok, I can add it this way.

SalocinHB: Hm. I think it's better to give them a hint, otherwise many people probably leave this 'as is'.

source/guide_up1.rst
ezzra: oops

source/guide_up1.rst
SalocinHB: `hash -r` will force an update.

ezzra: ok nice

PR #264, opened 2018-11-17
[etherpad] use tag instead of branch

Stunkymonkey:
for example the current 1.7.0 is not heaving an branch called release/1.7.0
but there is a tag for it.


luto:
Thanks :)

PR #272, opened 2018-11-30
Correct pseudo version number

errotu:
Since "66.6.6" is used as pseudo version number in the rest of this tutorial, it should be used in the update section as well. This will prevent irritations.


luto:
thanks!

PR #273, opened 2018-11-30
Information about customization added

errotu:
Currently the personalization part is quite short. I added information regarding the use of css-files for customization.

### DIFF
--- a/source/guide_etherpad.rst
+++ b/source/guide_etherpad.rst
@@ -197,6 +197,10 @@ Personalization
 
 Take a deeper look into the ``~/etherpad/settings.json``, you still might want to adjust the title or the welcoming text of new created pads. If you want to use plugins, you will also need to set up a admin account there.
 
+You can also personalize the look of your installation by including custom css- oder js-files. See `Custom static files`_ in the documentation for further information.
+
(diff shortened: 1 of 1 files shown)


source/guide_etherpad.rst
```diff
@@ -197,6 +197,10 @@ Personalization
 
 Take a deeper look into the ``~/etherpad/settings.json``, you still might want to adjust the title or the welcoming text of new created pads. If you want to use plugins, you will also need to set up a admin account there.
 
+You can also personalize the look of your installation by including custom css- oder js-files. See `Custom static files`_ in the documentation for further information.
+
+**Important**: You need to place your custom-files in ``~/etherpad/node_modules/ep_etherpad-lite/static/custom`` **not** ``~/etherpad/static/custom`` for being recognized by your installation.
```
luto: isn't that directory overwritten on updates?

errotu: I didn't test it yet since no new update was published since I modified my installation. Nevertheless, as this is a "custom"-folder, I think it wouldn't get overwritten.

To be safe, we can wait until 1.7.1 is published and I'll check it.

SalocinHB: or maybe just try to install an older version, customize it and then update the installation.

errotu: I tested it (uprade from v1.6.6 to v1.7.0) and it worked without any problems.

Issue #271, opened 2018-11-28
CodiMD MySQL does not work with unicode emoji


kindfulkirby:
How to reproduce: Stop the CodiMD service, start it again, Unicode emoji (and possibly other unicode glyphs, haven't checked) will be broken.

There is a reason I used sqlite when I wrote that guide :wink: 

(Sorry about the broken link, btw, I could no longer stand by the h8 domain after being informed about what some people use those two characters for... -.-)

EDIT: I have made the original post this was based on available again on a new, much nicer domain: https://n.d9.lv/s/how_to_install_hackmd_on_uberspace

PR #277, opened 2018-12-02
Changed CodiMD guide to SQLite

kindfulkirby:
CodiMD is not compatible with Uberspace's MySQL installation, running it with MySQL can lead to corruption of Unicode data like emoji.
SQLite does not have this problem, sadly I was not able to run CodiMD with a database in any other location than the folder it is run from, neither absolute nor relative paths work with it.

### DIFF
--- a/source/guide_codimd.rst
+++ b/source/guide_codimd.rst
@@ -1,5 +1,7 @@
 .. highlight:: console
 .. author:: stunkymonkey <stunkymonkey.de>
+.. author:: Kirby <kindful.work>
+
 
 .. sidebar:: Logo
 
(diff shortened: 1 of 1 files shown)


SalocinHB:
What exactly is the problem with MySQL?

kindfulkirby:
Oh, right I forgot to reference the issue. This fixes #271.

SalocinHB:
The problem is most likely that the database is created with the `utf8` charset instead of `utf8mb4`.

[isabell@stardust ~]$ mysql -e "CREATE DATABASE <username>_codimd CHARACTER SET utf8 COLLATE utf8_general_ci;"`

kindfulkirby:
That was the first thing I've tried. It results in one of the columns being too large for it's type, because now every character uses 4 bytes and Uberspace's MySQL isn't compiled with the right flags for that.

Maybe don't close this with a "most likely"?

SalocinHB:
Are you sure you've tried that with MariaDB 10.3?

SalocinHB:
Just tried it, works with 10.3.

PR #280, opened 2018-12-04
Simple supervisord path

nichtmax:


source/includes/install-uwsgi.rst
```diff
@@ -15,7 +15,7 @@ Create  ``~/etc/services.d/uwsgi.ini`` with the following content:
   :emphasize-lines: 2
 
   [program:uwsgi]
-  command=/home/<username>/.local/bin/uwsgi --master --emperor /home/<username>/uwsgi/apps-enabled
+  command=%(ENV_HOME)s/.local/bin/uwsgi --master --emperor /home/<username>/uwsgi/apps-enabled
```
luto: the 2nd `<username>` can be replaced with `$(ENV_HOME)s` as well. Thus, the whole block can go :)

PR #283, opened 2018-12-09
PostgreSQL guide update

maitrenageur:
I solved small wording mistakes and placed the focus on an example (synapse) as data base name (line 351 and 358) instead to use the user name during the data base creation process to avoid misunderstandings.

Best regards

FM


nichtmax:
Thank you very much!

PR #284, opened 2018-12-11
auto(re)start supervisord services

SalocinHB:


luto:
`autostart` defaults to `true` and `autorestart` to `unexpected`, which means restart when the exit code is not 0 or 2, so an error occurred. So everything should be fine out of the box? Which issue does this resolve?

SalocinHB:
> Which issue does this resolve?

A few users reported that their Etherpads quit and weren't restarted automatically.

luto:
Then etherpad (or node?) is probably sloppy with its exit codes. Instead of adding the lines to every guide, what do you think about giving just etherpad a `autorestart=yes` line, so it restarts regardless of the exit code. `autostart` is true by default and not needed.

PR #286, opened 2018-12-16
[akaunting] add paragraph regarding htaccess

komaspieler:
.. to prevent internal server errors. Otherwise, the current release version ends up redirecting in loops until the internal redirect limit is reached.

### DIFF
--- a/source/guide_akaunting.rst
+++ b/source/guide_akaunting.rst
@@ -52,6 +52,9 @@ Your domain needs to be set up:
 Installation
 ============
 
+Step 1: Download latest release
+-------------------------------
+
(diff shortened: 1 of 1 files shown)


SalocinHB:
> Otherwise, the current release version ends up redirecting in loops until the internal redirect limit is reached.

Only if you install it in an additional domain document root, as described in https://manual.uberspace.de/web-documentroot.html#additional-documentroots. All lab guides assume standard folders, i.e. `/var/www/virtual/$USER/html`.

Also, `/var/www/virtual/$USER/.htaccess` isn't a very good location for your `.htaccess` if you're using additional docroots. Rather use one per domain.

Issue #175, opened 2018-08-21
Add guide for TYPO3 CMS


tstahn:
Since we already have a nice set of Content Management Systems on board, why not add TYPO3 CMS to the mix.

I have already been working on it and could provide the guide if you guys are interested,
https://github.com/tstahn/lab/blob/feature/add-guide-typo3-cms/source/guide_typo3-cms.rst

PR #309, opened 2019-02-04
[FEATURE] Add guide for TYPO3 CMS 9.5 LTS

tstahn:
Resolves: #175


source/guide_typo3-cms.rst
```diff
@@ -0,0 +1,164 @@
...
+   - Installing typo3/class-alias-loader (1.0.1)
+ [...]
+ [isabell@stardust isabell]$
+
+Now remove the `document root`_ and create a symlink to the ``public`` directory.
+
+::
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/
+ [isabell@stardust isabell]$ rm -rf html
```
luto: Even though all guides assume an empty, newly created uberspace, some newer users might not adhere to this principle. What do you think about adding a `.. warning::` above this command block, explaining that everything will be deleted?

tstahn: Better safe than sorry, so fine with me.

source/guide_typo3-cms.rst
```diff
@@ -0,0 +1,165 @@
...
+created earlier.
+
+If you wish to make changes to your installation at a later stage, use the `Install Tool`_.
+
+For further details, please have a look at the official `installation guide`_.
+
+Updates
+=======
+
+Subscribe to the `TYPO3 Announce`_ mailing list to get regular updates regarding new TYPO3 releases and security bulletins.
```
luto: Please document how to actually update a typo3 instance installed like in this guide. The built-in update seems to be disable by default. Depending on the complexity a link to the relevant typo3 documentation article might be sufficient.

luto:
Thank you for this guide :) it works great :grin: I've changed a few minor things and added notes on two sections that need a little expansion, in my opinion.

tstahn:
Thanks for the feedback. I added the suggested changes to the two sections you mentioned. Hope that works for you.

luto:
thank you!

PR #310, opened 2019-02-11
Disable maintance mode after running nextcloud upgrade

hanzei:


SalocinHB:
This shouldn't be necessary, 'occ upgrade' will disable maintenance mode after a successful upgrade. 

See https://docs.nextcloud.com/server/14/admin_manual/configuration_server/occ_command.html#command-line-upgrade-label

hanzei:
Well, on my two Nextcloud 15 instances it doesn't. I did download and install the release and run the upgrade command as described in https://lab.uberspace.de/guide_nextcloud.html#updates. The log ended with:
```
Finished code integrity check
Update successful
Maintenance mode is kept active
Reset log level
```

nichtmax:
Just a quick heads up: Sorry for the long waiting time, we'll get in touch with you soon. We're a small team and sometimes things take longer than we anticipated...

luto:
According to [the source(TM)](https://github.com/nextcloud/server/blob/master/lib/private/Updater.php#L134-L139), nextcloud keeps the maintenance on, if it was enabled already. If it's off when the update is started, it will be turned off again.

PR #311, opened 2019-02-11
Add guide for Mailman 3

### LINKED ISSUE #141: add a guide for Mailman3
GNU Mailman is a computer software application from the GNU Project for managing electronic mailing lists. Mailman is coded primarily in Python and currently maintained by Abhilash Raj. Mailman is free software, licensed under the GNU General Public License.

http://docs.mailman3.org/en/latest/

rikroe:
Fixes #141.

A guide for Mailman 3 on U7. Has been tested on a clean U7 but requires quite a lot of configuration and some helper scripts. 

Looking forward to your feedback!

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -1,0 +1,464 @@
+.. highlight:: console
+
+.. author:: Richard Kroegel <info@kroegel.org>
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/mailman.jpg
+      :align: center
+
+#########
+Mailman 3
+#########
+
+`Mailman`_ is free software for managing electronic mail discussion and e-newsletter lists. Mailman is integrated with the web, making it easy for users to manage their accounts and for list owners to administer their lists. Mailman supports built-in archiving, automatic bounce processing, content filtering, digest delivery, spam filters, and more.
+
+`Mailman 3`_ is a complete re-write of previous versions, consisting of a suite of programs that work together:
+
+  * **Mailman Core**; the core delivery engine. This is where you are right now.
+  * **Postorius**; the web user interface for list members and administrators.
+  * **HyperKitty**; the web-based archiver
+  * **Mailman client**; the official Python bindings for talking to the Core’s REST administrative API.
+
+----
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * Python_
+  * supervisord_
+  * Folder/File Permissions
+
+License
+=======
+
+Mailman is released under the GNU General Public License
+
+  * https://www.gnu.org/copyleft/gpl.html
+
+Prerequisites
+=============
+
+Your URL needs to be setup for web and mail:
+
+::
+
+ [isabell@stardust ~]$ uberspace web domain list
+ isabell.uber.space
+ [isabell@stardust ~]$ uberspace mail domain list
+ isabell.uber.space
+ [isabell@stardust ~]$
+
+Additionally, create a mailbox_ for Mailman to use to send e-mails. In this example, we are going to use ``forwarder@isabell.uber.space``.
+
+Installation
+============
+
+Get Mailman 3
+-------------
+Install Mailman 3 and its dependencies via pip.
+
+::
+
+ [isabell@stardust ~]$ pip3.6 install --user mailman postorius hyperkitty mailman-hyperkitty django whoosh
+ [...]
+ [isabell@stardust ~]$
+
+Get Dart Sass
+-------------
+Postorius and HyperKitty are build using SASS_ stylesheets which have to be compiled/decompiled before use. As U7 does not provide an SASS compiler, we need to download the *Dart Sass* standalone binary. Get the link of the latest Dart Sass binary for **linux-x64** from https://github.com/sass/dart-sass/releases, download and extract the binaries:
+
+::
+
+ [isabell@stardust ~]$ wget https://github.com/sass/dart-sass/releases/download/9.99.9/dart-sass-9.99.9-linux-x64.tar.gz
+ [isabell@stardust ~]$ tar xzvf dart-sass-9.99.9-linux-x64.tar.gz dart-sass
+ [isabell@stardust ~]$ mv dart-sass ./bin/ 
+ [isabell@stardust ~]$ rm dart-sass-9.99.9-linux-x64.tar.gz 
+ [isabell@stardust ~]$
+
+
+Get the mailman-suite example configuration
+-------------------------------------------
+To have a starting point for configuration, we use the mailman-suite_ example provided by the Mailman team and move it to a path where we can access it easily.
+
+::
+
+ [isabell@stardust ~]$ git clone https://gitlab.com/mailman/mailman-suite.git
+ [isabell@stardust ~]$ mv mailman-suite/mailman-suite_project/ .
+ [isabell@stardust ~]$ rm -rf mailman-suite
+ [isabell@stardust ~]$ mv mailman-suite_project/ mailman-suite
+ [isabell@stardust ~]$ 
+
+
+Get and enable uwsgi
+---------------------
+.. include:: includes/install-uwsgi.rst
+
+Get .qmail helper script
+-------------------------
+Mailman 3 uses LMTP to transfer emails locally. As qmail_ is not able to use this directly, we need to download a helper script from the mailman source and adjust it to our needs:
+
+::
+
+ [isabell@stardust ~]$ cd bin
+ [isabell@stardust bin]$ wget https://gitlab.com/mailman/mailman/raw/master/contrib/qmail-lmtp
+ [isabell@stardust bin]$ chmod +x qmail-lmtp
+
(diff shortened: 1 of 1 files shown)


source/guide_mailman-3.rst
```diff
@@ -0,0 +1,464 @@
...
+Installation
+============
+
+Get Mailman 3
+-------------
+Install Mailman 3 and its dependencies via pip.
+
+::
+
+ [isabell@stardust ~]$ pip3.6 install --user mailman postorius hyperkitty mailman-hyperkitty django whoosh
```
luto: `django` is already pulled in via a dependency, please remove it form the explicit list.

source/guide_mailman-3.rst
```diff
@@ -0,0 +1,464 @@
...
+ [isabell@stardust bin]$ wget https://gitlab.com/mailman/mailman/raw/master/contrib/qmail-lmtp
+ [isabell@stardust bin]$ chmod +x qmail-lmtp
+
+Now edit ``qmail-lmtp`` and change the two following lines:
+
+.. code :: python
+
+ 1: #!/usr/bin/env python3.6
+ [...]
+ 18:            os.environ['EXT' + (sys.argv[2] if len(sys.argv) >= 3 else '')] + "@" + os.environ['HOST'],
```
luto: Can you expand a bit on why this is needed? Is this only relevant to our specific setup or to all qmail installations? If all of them are affected, I think contributing this to the mailman repo would be nice.

luto:
Thanks for you PR. It's really comprehensive and nicely written, but I am not quite sure about the qmail integration. Adding new and editing existing scripts, which are executed and updated manually doesn't seem like a tidy solution to me. I realize that putting this together was a lot of work, but I fear that we will not be able to support this guide in the long term.

Mailman enables MTAs to react to events like ["create a new list"](https://gitlab.com/mailman/mailman/blob/master/src/mailman/app/lifecycle.py#L100), via the `config.mta.incoming` setting. The given value is "just" the name of a python class providing a couple of methods to manage lists: [create/delete/regenerate](https://gitlab.com/mailman/mailman/blob/master/src/mailman/interfaces/mta.py#L67-83). Do you think it might be reasonable to contribute an MTA implementation for qmail to the mailman core? This would allow the guide to be much slimmer, as all the .qmail files would be managed by mailman itself.
If not, we'll have to discuss how to proceed with this PR.

Thank you very much again!

source/guide_mailman-3.rst
Pokerkoffer: It is relevant for all qmail installations that want to use mailman3, see official docs: https://mailman.readthedocs.io/en/latest/src/mailman/docs/mta.html#qmail

rikroe: As mentioned, I was only following the official docs on this.
Unfortunately I was not able to use the "official" helper script in the original form as it expects an environment variable `EXT1 `(being called by ``echo '|qmail-lmtp 8042 1' > .qmail`` where the last ``1`` is used to stich the variable name of `EXT1 `together. 
When I invoked the original script via qmail, the environment variable ``EXT1`` was not found. On U7, we only have ``EXT``, ``EXT2``, ``EXT3`` and ``EXT4``. Not sure where this comes from though.

rikroe:
Thanks for your feedback!

I can understand your reservations about supporting this guide. After it was finished I a little bit surprised on the amount of text I produced.

Regarding the .qmail issue: please see above why I've done it this way. I can try to create a pull request in mailman-core to incorporate the change to this one line as it shouldn't break any of their stuff. 

Even if I would able to include the second qmail-related script (the creation of qmail files in `~` like in the mailman 2 guide) I still think that we would need an (updated) helper script. Mailman 3 only accepts incoming mails via LMTP and from what I've found in the net this helper script is the only solution except running another exim or postfix server (although piping the mail to text and send sending this via a Python SMTP implementation is arguably not the nicest design).

source/guide_mailman-3.rst
luto: According to [the qmail docs](http://www.qmail.org/man/man8/qmail-command.html), there is not `EXT1` :sweat_smile: did you ever encounter one in the wild and, if yes, where?

rikroe: No, actually I didn't. As said before, just followed the official docs. 
Checked the qmail docs as well yesterday and will create a pull request for the `qmail-lmtp` script in the official repository so we can fully comply with their documentation.

luto:
I'll close this for now. Feel free to re-open or create a new PR once the changes have been integrated into mailman itself. Thank you!

rikroe:
Hey, for now just trying to make clear what needs to for the guide to be accepted by you guys.
From your comments, there are two issues:

1. Implicitely installing django when it's a dependcy. 
2. Streamline the qmail/mailman integration.

Part 1 should not be a problem. I have to check this, I somehow remember that I had to install django implicitely as otherwise some django dependencies were not loaded correctly. Will update accordingly.

Part 2 is where the fun begins 😅  

- The [qmail-lmtp](https://gitlab.com/mailman/mailman/blob/master/contrib/qmail-lmtp) helper script has been updated in the official repository, so we don't need to edit the Python code anymore. 
However, the script is still required. I have found no other solution for qmail to use the LMTP protocol mandated by Mailman 3.
- I probably won't be able to create a Mailman 3 interface for creating/recreating .qmail files as you suggested anytime soon (time constraints and I'm not really good at Python). Would an additional helper script (that has to be triggered manually after creating a mailing list) be sufficient for you? Both Mailman 2 guides for U6 and U7 use the same approach. 

Would those changes be enough to reconsider including this guide? Otherwise I'll probably won't continue on this path much further. I can however fully understand you why you might not feel comfortable in officially adding this, no offense whatsoever taken 👍

luto:
> * The [qmail-lmtp](https://gitlab.com/mailman/mailman/blob/master/contrib/qmail-lmtp) helper script has been updated in the official repository, so we don't need to edit the Python code anymore.

nice! thank you.

> * I probably won't be able to create a Mailman 3 interface for creating/recreating .qmail files as you suggested anytime soon (time constraints and I'm not really good at Python). Would an additional helper script (that has to be triggered manually after creating a mailing list) be sufficient for you? Both Mailman 2 guides for U6 and U7 use the same approach.

I'll have to pass this onto @nichtmax. We'll discuss it next week and come back to you with an answer :) At the very least a 2nd script allowing the removal of lists is needed. But that shouldn't be a huge problem.

source/guide_mailman-3.rst
```diff
@@ -0,0 +1,464 @@
...
+ pid_file: /home/isabell/var/master.pid
+ lock_file: /home/isabell/var/locks/master.lck
+
+ [mailman]
+ layout: custom
+
+Daemonizing Mailman Core
+------------------------
+
+As we want to make sure that Mailman is started automatically, we need to set it up as a service. Due to mailman not having having the option to always run in foreground, we need some other means of controlling it.
```
luto: `mailman start` forks and then executes `.local/bin/master` (yes, it's just called "master") with `--config`. That one doesn't fork and stays in the foreground. Since that executable is publicly available and has a friendly `--help', I guess using it directly would be fine.

The start helper also does the following:

```
os.environ['MAILMAN_VAR_DIR'] = config.VAR_DIR
os.chdir(config.VAR_DIR)
```

so we could adapt the supervisord config to read

```
environment=MAILMAN_VAR_DIR="%(HOME)s/var"
directory=%(HOME)s/var
```

Although looking at [config.py](https://gitlab.com/mailman/mailman/blob/master/src/mailman/config/config.py#L158), where this variable is ultimately used, I am pretty sure we can skip that if all paths in the config are absolute (which they are).

luto:
After some research I was able to get rid of the `mailman start` wrapper script :tada: If you have the time that last change would be great. Otherwise I'll take care of it during the merge.

> I probably won't be able to create a Mailman 3 interface for creating/recreating .qmail files as you suggested anytime soon 

Thanks for looking into it! We'll merge this as-is, then. I really appreciate the work you put into this. Also, thanks for staying with me through this rather harsh process :sweat_smile: 

Github doesn't let me reopen this because the branch was force-pushed (who knew?). Once I've got an answer I'll merge this manually :)

rikroe:
Cool, many thanks for not just leaving it where it was!
Really appreciate the effort, just another reason for uberspace!

Will implement your suggestions, looks great! 🎉

Already changed some things and will create a new pull request soon! 

Thanks for your help 😊 And no offense for the process taken, I totally agree with your points! Therefore really appreciative!

source/guide_mailman-3.rst
rikroe: Not knowing that much about Python coding best practices, this seems somehow weird 😕 
But great that you found this out, will adapt the supervisord config.

If we set the environment variable and the working directory from supervisor, we probably don't even need to hard-code absolute paths into the config 🎉

source/guide_mailman-3.rst
rikroe: Remove explicit django pull

PR #312, opened 2019-02-11
Add make.bat for windows machines

rikroe:
This adds a make.bat to be able to build the Sphinx HTML on Windows. The original file comes from Sphinx directly (see [here](http://howto.nspx.ca/how-to-sphinx.html#using-sphinx-autobuild)) and has been adjusted for Uberlab. 

In a virtual environment, run ``make.bat html`` just like you would do on Linux:
```
(venv) PS R:\uberlab> .\make.bat html
[...]
The HTML pages are in build\html.
(venv) PS R:\uberlab> 
```

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -1,0 +1,36 @@
+@ECHO OFF
+
+pushd %~dp0
+
+REM Command file for Sphinx documentation
+
+if "%SPHINXBUILD%" == "" (
+	set SPHINXBUILD=sphinx-build
+)
+set SOURCEDIR=source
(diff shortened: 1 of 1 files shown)


luto:
Thank you! Since nobody in our team uses windows _and_ is working on the lab, I had to add a little disclaimer. I hope you don't mind :)

rikroe:
Hehe totally fine with me 👍 
Actually excpected something like that. Thanks for including anyway :)

PR #313, opened 2019-02-12
Updated prometheus version

Felix-Franz:


source/guide_prometheus.rst
```diff
@@ -178,6 +178,6 @@ If this is something you do not want to do, you could hide it behind a basic aut
 
 ----
 
-Tested with Prometheus_ 2.6.0, Uberspace 7.1.19.0
+Tested with Prometheus_ 2.6.0, Uberspace 7.1.19.0 and Prometheus_ 2.7.1, Uberspace 7.2.2.2
```
luto: Please just specify the newest versions here.

luto:
Thanks!
This took a while since we're not really set on the software version that should be provided in guides (`latest` vs. a pseudo-version vs. the lates one vs. none at all vs. ...). For now we'll stick to the latest one. Sorry for the missing status update on our part.

Issue #262, opened 2018-11-14
make it easier to link to manual


luto:
Almost all guides reference manual pages (e.g. MySQL) using the full URL. This is not easy to maintain in the future and adds noise.

```shell
$ grep -r manual.uberspace.de *.rst | wc -l
166
```

- [x] write a plugin, shortcode, something to reference the manual, e.g. `:manual:mysql`
- [x] replace all links by that shortcode
- [x] add a note about this to the style guide

FriedrichFroebel:
There already seems to be an official Sphinx extension to achieve this result: [`extlinks`](http://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html).

Activate the extension:
```python
# Enable the extension.
extensions = [
  'authorship',
  'sphinx.ext.extlinks',
]

# Configure the extension to kick in when the `manual` directive is used.
extlinks = {"manual": ("https://manual.uberspace.de/%s.html", "")}
```

Now linking to the manual is possible using
```rst
* :manual:`Python <lang-python>`
* :manual:`display text <target>`
```
in the RST files.

This may be simplified by using a custom version of the [extension](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/extlinks.py) to avoid having to write the display text besides the target itself (using a hard-coded dictionary or similar approaches).

luto:
considering the current situation, having to repeat the display text is fair, I guess :sweat_smile:

luto:
The `extlinks` extension is now active thanks to @FriedrichFroebel! This means that all  guides can now be refactored into the new link format. Take a look at the Akaunting guide (3e156181af7894818a68e2d7b4989c6c32a76fdf) as an example on how to do this.

PRs refactoring all or even single guides are very welcome!

FriedrichFroebel:
I will have a look at all the guides, replace the links to the manual and create a corresponding PR in the next days.

Additionally the style guide needs to be changed as well - the new PR to add a guide for *bludit* seems to still use the old way of linking for example, as only this issue (and my old PR) are mentioning this at the moment.

luto:
wow, that would be awesome! :tada: 

> the new PR to add a guide for _bludit_

ha, well, the styleguide (and apparently the reviewer, me) isn't updated yet :man_facepalming: :grin:

luto:
All done thanks to awesome @FriedrichFroebel :tada:

PR #315, opened 2019-02-18
Enable `extlinks` extension

FriedrichFroebel:
This enables the official [`extlinks`](http://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html) to improve links to the manual (see #262).

With this change referencing the manual is possible using one of the following directives:
```rst
  * :manual:`Python <lang-python>`
  * :manual_anchor:`Change Python version <lang-python.html#change-version>`
```
There are two different directives at the moment to support anchors while avoiding having to write the `.html` part in the "default" case. This may be changed if needed - as well as the naming of the second directive.

An additional change is to unify the indentation of the code inside the configuration file according to the values inside the `.editorconfig` file.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -87,7 +87,7 @@ extlinks = {
         'https://manual.uberspace.de/%s.html',
         ''
     ),
-    'manualanchor': (
+    'manual_anchor': (
         'https://manual.uberspace.de/%s',
         ''
     ),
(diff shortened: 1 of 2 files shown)


luto:
Nice! can you please apply the refactor to a single guide of your choice, so we've got an example to link in the issue?

FriedrichFroebel:
Done for the Akaunting guide which now uses the directives.

luto:
great! :tada: thank you!

PR #316, opened 2019-02-19
fix proxy rewrite

ezzra:
put some more logical coherence into it

### DIFF
--- a/source/includes/proxy-rewrite.rst
+++ b/source/includes/proxy-rewrite.rst
@@ -8,7 +8,9 @@ Create a ``~/html/.htaccess`` file with the following content:
  DirectoryIndex disabled
 
  RewriteEngine On
- RewriteRule ^(.*) http://localhost:<yourport>/$1 [P]
+ RewriteCond %{REQUEST_FILENAME} !-f
+ RewriteBase /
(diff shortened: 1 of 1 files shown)


luto:
This adds static file handling, which might not be needed by all applications. While we do not have a "policy" on that, I prefer only providing the htaccess rules needed for each app.

ezzra:
@luto that sounds like you are referencing https://github.com/Uberspace/lab/commit/f5fbad4b294f422cefcca08ebb4e86aa7cbe743a#diff-78baa3073cfcac7c8d6ec23835173ff5 where the original change to static file handling has been made. I am with you, that this might have been the wrong place, but it was already accepted. My PR here is only fixing it up, because at the moment its like this for all guides:

screenshot_2019-02-19_16-31-56

which is more magical than logical :)

luto:
Oh, I see, sorry :sweat_smile: yes, that doesn't make a lot of sense!

I've moved the rewrite plus static into a dedicated include (used in django only) and removed static handling from the default one.

luto:
The new rewrite-static also includes your fix. Thank you!

luto:
https://github.com/Uberspace/lab/commit/f3b49f7a85c9eb0cd0f9e9d79c8cdc51590a1e28

PR #320, opened 2019-02-28
add guide for bludit

netbuk:
create a guide for bludit cms


source/guide_bludit.rst
```diff
@@ -0,0 +1,126 @@
...
+  * admin password: set up your admin password.
+  
+
+Tuning
+======
+
+For plugins, themes and other stuff go to Bludit_.
+
+
+Updates
```
luto: Can you please add the concrete commands used to do this? Especially the step preserving `/bl-content/database/` would be very helpful to users!

source/guide_bludit.rst
```diff
@@ -0,0 +1,126 @@
...
+.. note:: Check the update feed_ regularly to stay informed about the newest version.
+
+
+1. First make a full backup from all files and folders.
+2. Remember which version of Bludit_ you are using for a possible roll-back.
+3. Download the latest version from Bludit_ or Github_ to your client.
+4. Extract the zip file.
+5. Replace existing files with the new files with your terminal or (s)ftp-client.
+
+.. warning:: The files in the directory ``/bl-content/database/`` should not be deleted. There the user accounts are stored.
```
luto: Are there no other files in `/bl-content` worth saving? Where do you store images and other assets in bludit?

luto:
Thanks for another submission :tada: I've fixed a nitpick and left two comments for you :) if you have any questions (or objections!) feel free to raise them :grimacing:

source/guide_bludit.rst
netbuk: Yes, there are the complete content files in this folder. I will set a folder tree and mark which files are important for an update.

source/guide_bludit.rst
netbuk: You mean the update process?
Mhm, since the update from the client to the server (see list) is performed, the way is actually already shown. For example, I use Transmit for such things. Other use e.g. Filezilla or the terminal.
I will try to write some textlines for using the terminal.

luto: Just like installation, the update should also be possible via the command line, to keep things consistent. Additionally, we don't require users to have "special" software like an sftp client on their machines.

netbuk: Done 
- add update process with terminal
- add folder structure with focus to important area
- add password recovery for admin account

luto:
merged manually, thanks! :tada:

PR #326, opened 2019-03-03
add guide for flatboard

netbuk:
create a guide for flatboard


source/guide_flatboard.rst
```diff
@@ -0,0 +1,190 @@
...
+.. sidebar:: Logo
+
+  .. image:: _static/images/flatboard.png
+      :align: center
+
+#########
+Flatboard
+#########
+
+Flatboard_ is a simple, lightweight, modern and fast flat-file forum, using Json and Markdown or BBcode.
```
luto: The control to select markdown instead of the default BBcode is actually disabled. Do you know why? :sweat_smile:   
image

netbuk: For the moment Markdown is not available, but should be fixed with one of the next updates.
bildschirmfoto 2019-03-04 um 15 27 19

luto:
Thanks!

Issue #331, opened 2019-03-07
:lab: always links to lab.uberspace.de


luto:
`:lab:` always links to lab.uberspace.de, not to the current domain. This is a problem in development setups and may be a problem in branch-domains, although I have not tested that.

FriedrichFroebel:
This is something I already thought about when introducing the lab directives, but settled with the current approach (using lab.uberspace.de as explicit target) to keep it as simple as possible.

One way to circumvent this could be to do this like the Pelican static site generator which uses one configuration file for development and another one for publishing - which in fact just imports everything from the development one and overwrites some of the configuration values afterwards. As the `extlinks` plugin is configured with a dictionary, replacing these values should not be the problem. This approach would need changes in the Makefiles (additional tasks) and probably is feasible with a known number of static setups only (for example for production and development purposes with a specific localhost port, but not for Netlify deployment as this includes the number of the PR inside the URL).

luto:
What about using domain-less links like `<a href="/guide_mailman-3.html"`?

FriedrichFroebel:
You are right, I did not think about this. Seems like the following should be used instead (did a quick test only):

```python
extlinks = {
    'manual': (
        'https://manual.uberspace.de/%s.html',
        ''
    ),
    'manual_anchor': (
        'https://manual.uberspace.de/%s',
        ''
    ),
    'lab': (
        '%s.html',
        ''
    ),
    'lab_anchor': (
        '%s',
        ''
    ),
}
```

As I just stumbled over the `requirements.txt` using a Sphinx version from Git, but mentioning version 1.7.3: This version has been released nearly a year ago, the current one is 1.8.4 (see [Release history](https://pypi.org/project/Sphinx/#history)).

FriedrichFroebel:
Another note: I just noticed that the only usage of the `:lab_anchor:` directive inside the Miniflux guide is missing the HTML extension and should read

```rst
:lab_anchor:`Guide <guide_postgresql.html#database-and-user-management>`
```

ebroda:
@luto isn't this closed by the merged PR?

luto:
Because the PR description did not say "closes #331".

PR #338, opened 2019-03-14
Do not use hardcoded lab URL for directive

FriedrichFroebel:
This changes the `:lab:` and `:lab_anchor:` directives to create domain-less links instead of the old ones enforcing links to `lab.uberspace.de` as discussed in #331. Additionally a broken link in the Miniflux guide gets fixed and the new guides are ported to use the directives if needed.

Besides this I updated the `requirements.txt` file to use a newer `sphinx-rtd-theme` version and to get rid of the Git commit for `sphinx` as the corresponding version mentioned inside the comment has been published nearly a year ago.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-# replace with 1.7.3 once released
 sphinx>=1.7.3,<2.0.0
 sphinx-rtd-theme==0.4.*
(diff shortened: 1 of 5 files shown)


luto:
thank you! :tada:

Issue #344, opened 2019-03-26
Mailman3 guide needs update due to new web backends


jonmz:
In the current guide, `localhost` is being used for LMTP. Unfortunately, at the time of writing, qmail doesn't enter the according network namespace which means that the `qmail-lmtp` script tries to connect to the _global_ `localhost` while Mailman3 is offering its LMTP port on the _user-specific_ `localhost` within the network namespace, so this is doomed to fail.

Either qmail needs to be fixed to enter the user's network namespace before calling `qmail-local` (which would be a more general solution), or the Lab guide needs to be updated to use `isabell.local.uberspace.de` instead of `localhost` in `qmail-lmtp` and in `mailman.cfg`.

luto:
> or the Lab guide needs to be updated to use isabell.local.uberspace.de instead of localhost in qmail-lmtp

https://gitlab.com/mailman/mailman/merge_requests/485

ezzra:
I am just trying to setting this up following the lab guide. What would I have to do to resolve this issue here for my installation? I included lutos fix from gitlab:

```
- lmtp = smtplib.LMTP("localhost", int(sys.argv[1]))
+ lmtp_host = sys.argv[3] if len(sys.argv) > 3 else 'localhost'
+ lmtp = smtplib.LMTP(lmtp_host, int(sys.argv[2]))
```
and also changed `mailman3-manage-qmail.sh` like

```
  i=$2
  p=62362 # Change this - first free port
  s=example.com # change this

  if [ $1 = "add" ]
  then
    echo Making links to $i in home directory...
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i-admin
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i-bounces
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i-confirm
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i-join
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i-leave
    echo "|/home/`whoami`/bin/qmail-lmtp $p 1 $s" > ~/.qmail-$i-owner
[...]

PR #357, opened 2019-04-08
some fixes to mailma3 guide

ezzra:
guide needs to be updated anyway because of web backends, but these fixes need to be applied apart from that


rikroe:
@ezzra many thanks for the fixes! Wanted to go over the guide by myself as well to update for web backends. Have you done any work on that already?

ezzra:
@rikroe no, as I pointed out in https://github.com/Uberspace/lab/issues/344 I dont really know how to fix that, but I would be VERY happy to have a fast solution here because I need to set this up for two projects!

PR #359, opened 2019-04-09
Update Mailman 3 guide for web backends

### LINKED ISSUE #344: Mailman3 guide needs update due to new web backends

rikroe:
Fixes #344.

The workaround mentioned in the issue is not needed anymore as the mailman source repo has been updated (thanks @luto).

I've additionally streamlined some parts (remove unnecessary qmail files, change ports from 900x to default).
Following the updated guide, I was able to get it working on a completely new U7 🎉


source/guide_mailman-3.rst
```diff
@@ -346,26 +351,17 @@ for Mailman 2 and is based on the script provided in the official installation i
  if [ $# = 2 ]
  then
    i=$2
-   p=61902 # Change this - first free port
+   p=8024 
+   h=isabell.local.uberspace.de
```
luto:
```suggestion
   h=${USER}.local.uberspace.de
```

this should work fine and doesn't require any manual changes.

rikroe: Even better!

luto:
thanks!

merged manually, because this MR contains unrelated commits. Try resetting your master to the master branch of this repo and using a dedicated branch for your next one. Sadly the github UI isn't terribly helpful with this :/

Issue #363, opened 2019-04-11
mailman3 hyperkitty


ezzra:
sorry for bringing up more mailman3 stuff ;)

It looks like the guide does not provide a proper way to keep hyperkitty (the archiving module) running. There needs to be setup some more crontabs and a qcluster process:

http://docs.list.org/en/latest/config-web.html#scheduled-tasks-required
http://docs.list.org/en/latest/config-web.html#running-the-task-queue-required

Before this hyperkitty did not show any archived lists for my installation. After I called the commands it does, but still no threads/emails. Mh, maybe someone else is already more experienced with the configuration here. Otherwise I would need to work on this.

rikroe:
Your comments are very appreciated, at least now I know that somebody actually puts it to use and it wasn't for nothing :D

I actually have a local commit adding a note on top stating that archiving is not working. Will create a PR for it soon. As I don't need it for my usecase, I'll probably skip trying to get it to work, sorry :worried: .

richard-dg:
Hi @ezzra, hi @rikroe,

thanks for the guide and thanks for this issue. I also stumbled upon the missing cronjobs for hyperkitty. I'm not sure if I understand correctly, that's why I'm asking. 

I just need to add those 7 jobs to my crontab?
> http://docs.list.org/en/latest/config-web.html#scheduled-tasks-required

What about this? How to set this up on an uberspace? Is this also a cronjob?
> http://docs.list.org/en/latest/config-web.html#running-the-task-queue-required
Thanks for clarification,

Richie

rikroe:
I'd guess you just need to add the 7 jobs to your [crontab](https://manual.uberspace.de/daemons-cron.html). 
The task queue should already be setup in the guide (see steps `~/uwsgi/apps-enabled/mailman-suite.ini`), but not 100% about this.
You have to try :) If only the cronjobs are missing, please make a small PR.

noave:
hyperkitty works now with #772

luto:
https://lab.uberspace.de/guide_mailman-3.html

PR #364, opened 2019-04-11
[mailman3] Polish guide & fix issues

### LINKED ISSUE #344: Mailman3 guide needs update due to new web backends

### LINKED ISSUE #362: mailman 3 updating
Currently there is no update strategy available within the guide. Actually its a problem that the guide states to remove the mailman_suite folder from the git versioned environment. I understand that the folder naming of the original repo is quite quirky, but that way it cannot be updated using git pull.

```
[isabell@stardust ~]$ git clone https://gitlab.com/mailman/mailman-suite.git
[isabell@stardust ~]$ mv mailman-suite/mailman-suite_project/ .
[isabell@stardust ~]$ rm -rf mailman-suite
[isabell@stardust ~]$ mv mailman-suite_project/ mailman-suite
[isabell@stardust ~]$
```

rikroe:
Fixes #362 (update strategy), fixes #344 (.qmail handling and proxy headers), adds a "not working" note for #363 (archiving does not work).

While update strategy, proxy headers and archiving are quite clear, the changes to the .qmail handling might need discussion. 
When leveraging ``.qmail-default`` for forwarding all incoming mails to mailman, some emails that have been received before might be getting lost. Therefore I added a warning as well as a note about manually adding .qmail-listname files (i.e. the old way, without an additional script).

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -342,7 +342,7 @@ Additionally, serve static files using apache:
 Setting up .qmail
 -----------------------------
 
-Because Mailman_ doesn't handle our .qmail-configuration automatically, we need to adjust  ``~/.qmail-default`` to forward all incoming mails to our LMTP handler. Update it with the following content. Make sure that ``8024`` is the LMTP port your :lab_anchor:`Mailman Core <guide_mailman-3.html#configure-mailman-core>` is listening and change your username twice:
+Because Mailman_ doesn't handle our .qmail-configuration automatically, we need to adjust  ``~/.qmail-default`` to forward all incoming mails to our LMTP handler. Update it with the following content. Make sure that ``8024`` is the LMTP port your :lab_anchor:`Mailman Core <guide_mailman-3.html#configure-mailman-core>` is listening on and change your username twice:
 
 .. code :: bash
 
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -227,13 +227,6 @@ After the REST backend has been configured, we need to configure the Django fron
  
  [...]
 
- USE_X_FORWARDED_HOST = True # Uncomment
- 
- SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https') # Uncomment
- SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_SCHEME', 'https') # Uncomment
- 
- [...]
-
  STATIC_ROOT = '/home/isabell/html/static/'
 
  [...]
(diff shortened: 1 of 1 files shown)


source/guide_mailman-3.rst
```diff
@@ -225,13 +227,6 @@ After the REST backend has been configured, we need to configure the Django fron
  
  [...]
 
- USE_X_FORWARDED_HOST = True # Uncomment
- 
- SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https') # Uncomment
- SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_SCHEME', 'https') # Uncomment
```
luto: While the host header is correct when using web backends, the protocol is still HTTP, which is normal in most proxy setups. Are you sure removing this doesn't cause unwanted HTTP=>HTTPS redirect-loops from the web app?

rikroe: No, I'm not sure. I've tried to access the website after removing and had no issues. However I don't know how Django will react here exactly.

luto: I vote for leaving it in, but I'm fine with removing it, if it works without it.

rikroe: +1 for leaving it in as well as it is resembling the actual technical setup.

rikroe:
@luto before this PR is merged I would like to check on some things/get you opinion regarding `.qmail-default`.

E.g. what would the impact be on actual mail boxes? I wrote that we need individual qmail files but meant only forwarding. The same should apply to mailboxes right? Which have previously been relying on `.qmail-default`?

So before this goes live I now feel that it needs to be a bit more descriptive.

rikroe:
After I've tried to come up with a solution that does conditional handling in qmail and failed, I've updated the guide to be more verbose and descriptive about the options for qmail handling. As soon as we have a new qmail guide in the manual, a link should be included and the section amended.

The only other real option I found would have been creating a (shell/python) script that first tries to forward to mailman and only if that fails, forwards to vmailmgr/vdeliver. It could be done in the qmail-lmtp helper script, but would totally blow this up. 

So now I hope the guide makes it clear that there is an active decision that needs to be taken with care.

PR #365, opened 2019-04-13
Update guide_radicale.rst

neeljp:
Change host according to https://manual.uberspace.de/web-backends.html?highlight=localhost#web-backends


neeljp:
Applications on uberspace7 need to listen on interface :: or 0.0.0.0 using localhost does not work.

PR #367, opened 2019-04-27
Information on how to increase php memory limit

judober:
I added information to the nextcloud entry on how to increase the php memory limit to the recommended value of 512 MB.


luto:
thanks! :tada: I shuffled the instructions around a bit, so php only needs one reload.

PR #372, opened 2019-04-29
add description for commit messages to style-guide

### LINKED ISSUE #370: reference guide in commit message
I would enjoy having a "commit message policy" to see immediately which guide is affected by a new commit. Just a line in `STYLE.md` that requests authors to mark their commits like `[tool] add update info` or just `tool - add update info`

That might be my special problem because I subscribed to the commit feed, but maybe its also good for others too :)

ezzra:
fixes #370


nichtmax:
Thanks!

PR #376, opened 2019-05-02
[mailman3] add misssing .qmail configuration

### LINKED ISSUE #375: mailman3: ~/.qmail-listname-default missing
Setting up .qmail:
In case `~/.qmail-default` shall be kept unchanged, `~/.qmail-listname` is to be created, but `~/.qmail-listname-default` should be created as well to allow using email commands like listname-subscribe@isabell.uber.space, etc.

jensguballa:
fixes #375


luto:
Thanks for the fix! :tada:

PR #378, opened 2019-05-05
fixed wrong step unter Updates

Fryboyter:
Step 3 describes how to create a new page. The reference to step 2 would be correct.


luto:
Oh, whoops. Close enough :grin: thanks!

Issue #379, opened 2019-05-06
Guide for Golang


ces92:
A guide for installing Go https://golang.org/

Snapstromegon:
If I'm correct this guide should be easy:
Download current version from [website](https://golang.org/dl/)
```
$ wget https://dl.google.com/go/go1.12.5.linux-amd64.tar.gz
```

Extract it to a basedir e.g. ~/go (first extracted level is go, so go would be created)
```
$ tar  -xzf go$VERSION.$OS-$ARCH.tar.gz # (tab complete would help)
```

And add ~/go/bin to path (needs to persist in .bashrc)
```
export PATH=$PATH:/home/$USER/go/bin
```

PR #382, opened 2019-05-11
[go] add go guide

ces92:
Installation guide for Go #379


SalocinHB:
Thank you for your guide. Unfortunately, we have decided not to accept it. Linuxbrew can become quite problematic and should only be used by users who know what they're doing.

PR #389, opened 2019-05-27
fixed a small typo in guide_postgresql.rst

fbskp:
temporary password file 'pgpass.tmp' had inconsistent file names across the guide (with/without dot in front)


SalocinHB:
fixed in #402

PR #391, opened 2019-05-30
add syncthing guide

Stunkymonkey:


luto:
Thanks :) Syncthing is a nice addition to the lab!

The installation works just fine, but after that I'm not able to sync any files from my laptop to my uberspace. Even though adding the folder works, the devices show up as "disconnected" on both ends. It seems that a direct TCP connection is necessary? If yes, that's [now possible](https://manual.uberspace.de/basics-ports.html). How did you manage to connect them?

luto:
Oh, I goofed up with the address handling. It actually works just fine! By default syncthing now uses a relay to transfer the data. @Stunkymonkey Would you like to expand the guide by adding the [firewall config](https://docs.syncthing.net/users/firewall.html), a new port as well as a short introduction how to enter the custom port into clients?

Stunkymonkey:
you mean the "LocalAnnouncePort"? (21027/UDP)

if someone would use it he also had to change it on their other device. Not sure if anyone would like to use it.

yes i can add it, but will not test it. Are you able to test this?

luto:
No, it should be `configuration=>options=>listenAddress`. A value like `tcp://0.0.0.0:22000` with whatever port `uberspace port add` outputs. If you're not interested, feel free to say no. I'll just add an issue for the future :)

Stunkymonkey:
should this be placed in "best practice" or would you recommend it generally?

Stunkymonkey:
sorry i messed a bit up with git, hope your notifications are not super annoying. everything should be right. I also tested it and it worked.

luto:
Thank you very much! :tada:

PR #413, opened 2019-06-18
initial guide for synapse

927589452:
works


927589452:
this is a working installation except starting and daemonizing the server

fahrplaner:
what exactly has to be done for the forwarding: for ease of use we will also forward 443 to the unsecure port ``8008``?

927589452:
@fahrplaner the web backend should be pointed to 8008, so clients can use https://my.domain.name for resolution;
i should clarify this

PR #416, opened 2019-06-19
[django] Add support for static files

The-Compiler:
Assuming a pretty standard Django project setup, with a /static folder inside the application and /static URLs to access the resources.

See https://uwsgi-docs.readthedocs.io/en/latest/StaticFiles.html


luto:
thanks!

Issue #58, opened 2018-04-04
Guide synapse matrix server


927589452:
Currently  working on a guide for the synapse matrix server

927589452:
Waiting for #59

glU79z:
Does this change anything? https://blog.uberspace.de/web-backends-websockets/

mulles:
@927589452  any updates concerning your guide?

927589452:
i was going to port this to the U7 https://janw.xyz/blog/installing-a-matrix-org-synapse-server-on- uberspace/ (original by @janw )

 I will take at least a week before I can try writing the guide again

927589452:
> Does this change anything? https://blog.uberspace.de/web-backends-websockets/
I still need a feature thats been announced with U7.3 ( thanks @luto ) [ports], as the matrix server needs to listen for federation servers and I need to set it up with my real dns a  SRV  Record for  _matrix._tcp: to announce the port, unless I can use 8443 or 8008; 
see https://github.com/Uberspace/manual/commit/8cd0cfc3fc2a98766c8d621f689f6a0f6cf20c53 
(or would this be possible in U7.4 with IPv6 at least)

luto:
Web backends should enable you to run synapse behind nginx, as it [can run on 443 behind a load balancer](https://matrix.org/docs/guides/faq#what-ports-do-i-have-to-open-up-to-join-the-global-matrix-federation%3F):

> You can also put Synapse entirely behind an existing TLS load balancer and not expose port 8448 at all. In this situation, Synapse will need to be configured to share the same public TLS certificate as the load balancer (as Synapse uses the public certificate for identity in other areas too, and it has to match the certificate that other servers see when they connect).

The ports feature in a future release will enable you to open ports in the 40.000-61.000 TCP/UDP range. They can be used in a `SRV _matrix._tcp` record, which seems to be the nicer solution. I am not sure how synapse normally handles certificate handling, though. Keep in mind, that you can currently not access the automatically issued let's encrypt certificates.

927589452:
> The ports feature in a future release will enable you to open ports in the 40.000-61.000 TCP/UDP range. They can be used in a `SRV _matrix._tcp` record, which seems to be the nicer solution. I am not sure how synapse normally handles certificate handling, though. Keep in mind, that you can currently not access the automatically issued let's encrypt certificates.

the last point hinders usage of the load balancer;
[...]

PR #418, opened 2019-06-24
Synapse guide new

927589452:
had some rewrites


927589452:
solves #58

luto:
nice! Since we know what's wrong with federation, do you mind delaying this guide until we fix the underlying bug in U7?

927589452:
I don't care (though I am looking forward for a mug ^^) 

https://github.com/Uberspace/lab/issues/58#issuecomment-503078471 mentions it beeing usefull without federation; this would be a nice family or office groupchat already

luto:
> though I am looking forward for a mug ^^

you got an emal ;)

source/guide_synapse.rst
```diff
@@ -0,0 +1,397 @@
+.. author:: 927589452
+.. highlight:: console
+
+.. tag:: python
+.. tag:: ports
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/mumble.svg
```
jpbernius: You can use the logo from here: https://matrix.org/images/matrix-logo.svg

927589452:
thanks @jpbernius 
fixed it

coderkun:
You know about this, right?
https://github.com/matrix-org/synapse/blob/master/docs/federate.md

927589452:
> You know about this, right?
> https://github.com/matrix-org/synapse/blob/master/docs/federate.md

yes, and my test setup uses the tcp announcement, but portforwardings do not yet work for IPv6 
at this point it is a nice group chat instance

927589452:
@coderkun if you can give me hints on how to setup the well known delegation I would be very thankful

coderkun:
I have never done this for Synapse/Matrix but as far as I understand it is simply a text file (`~/html/.well-known/matrix/server`) with the content

    {
        "m.server": "<synapse.server.name>[:<yourport>]"
    }

Maybe a `.htaccess` file to enforce the JSON mimetype cannot hurt, too.

927589452:
@coderkun implemented both

927589452:
> I have never done this for Synapse/Matrix but as far as I understand it is simply a text file (`~/html/.well-known/matrix/server`) with the content
> 
> ```
> {
>     "m.server": "<synapse.server.name>[:<yourport>]"
> }
> ```
> 
> Maybe a `.htaccess` file to enforce the JSON mimetype cannot hurt, too.

i didn't realize it would only be a text file^^
and serve it via the web backend

coderkun:
Just checked how [librem.one](https://librem.one) is doing the federation and they use the DNS approach:

`$ dig -t srv _matrix._tcp.librem.one`:

    ;; ANSWER SECTION:
    _matrix._tcp.librem.one. 300	IN	SRV	10 0 8448 chat.librem.one.

927589452:
> Just checked how [librem.one](https://librem.one) is doing the federation and they use the DNS approach:
> 
> `$ dig -t srv _matrix._tcp.librem.one`:
> 
> ```
> ;; ANSWER SECTION:
> _matrix._tcp.librem.one. 300	IN	SRV	10 0 8448 chat.librem.one.
> ```

Both are described in the guide

927589452:
@luto even though the https://federationtester.matrix.org/ marks us as not working we do federate with https://matrix.to/#/#datenkanal:matrix.org without problems

i think this guide is a go

PS: did the mail bounce?

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+.. tag_list::
+
+Synapse_ is the reference implementation of a matrix server.
+Matrix_ is federated chat protocol aiming to replace xmpp.
+
+----
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * PostgreSQL_
```
FriedrichFroebel: Both of the links here seem to be broken.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+  [isabell@stardust ~]$
+
+Installation
+============
+
+We will install synapse using pip, which makes it quite easy,
+but we start with creating a virtual environment for synapse to run in
+
+.. code-block:: console
+  :emphasize-lines: 1,2,8,18.27,37
```
FriedrichFroebel: This code block does not show up. You should replace the period with a comma.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+    --owner="synapse" \
+    --template=template0 \
+    synapse
+  [isabell@stardust ~]$
+
+Installation
+============
+
+We will install synapse using pip, which makes it quite easy,
+but we start with creating a virtual environment for synapse to run in
```
FriedrichFroebel: You probably want to end the sentence here.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+  (env) [isabell@stardust ~]$
+
+
+Configuration
+=============
+
+Generate standard config
+------------------------
+
+generate a config file ``~/synapse/homeserver.yaml`` and replace my.domain.name with the real domain you want your matrix username to end in.
```
FriedrichFroebel: Starting all of your sentences with an uppercase letter should be better. (Also applies to some sentences below.)

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+        bind_addresses: ['::','0.0.0.0']
+        x_forwarded: true
+        resources:
+          - names: [client]
+
+for ease of use we will make use of ``uberspace web backend`` and point it to 8008.
+
+.. include:: includes/web-backend.rst
+
+To enable federation as described _Matrix_Federation we need to announce it either via DNS or via .well-known.
```
FriedrichFroebel: The link here is not working.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+  .. image:: _static/images/matrix-logo.svg
+      :align: center
+
+######
+Synapse
+######
+
+.. tag_list::
+
+Synapse_ is the reference implementation of a matrix server.
```
FriedrichFroebel: Both links seem to be broken.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+ [isabell@stardust ~]$ supervisorctl status
+ synapse                            RUNNING   pid 26020, uptime 0:03:14
+ [isabell@stardust ~]$
+
+If it's not in state RUNNING, check your configuration.
+
+Updates
+=======
+
+Watch _Matrix_RSS to be notified of upgrades and if there is a update, use pip to update the installation:
```
FriedrichFroebel: The link here does not seem to work as well.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+  [isabell@stardust ~]$
+
+
+Administration
+==============
+
+Adding users
+------------
+
+Users can be added from the CLI:
```
FriedrichFroebel: There seems to be some unwanted duplication here.

source/guide_synapse.rst
```diff
@@ -0,0 +1,424 @@
...
+  synapse=> UPDATE users SET password_hash='$2b$12$yK16TMDMnvj97GFBoxF9QeP2N.U8oadindcjB0Uo9TkSI3CsgwV02' WHERE name='@isabell:my.domain.name';
+  UPDATE 1
+  synapse=>
+  [isabell@stardust ~]$
+
+.. note:: ``SELECT name from users;`` returns a list of all usernames
+
+
+Tested on uberspace 7.3.1.1 via riot.im/app on synapse 1.0.0.
+.. _Matrix_Federation: https://github.com/matrix-org/synapse/blob/master/docs/federate.md
```
FriedrichFroebel: Adding an empty line before the links may fix some of your link problems.

927589452:
Thanks @FriedrichFroebel for the suggestions

927589452:
@luto I just realized we could technically just announce everything (client and federation) on port 443 
do you have a preference there ?
Bonus: less work configuring ports etc
Malus: no split between federation and client

luto:
> we could technically just announce everything (client and federation) on port 443

Doing everything over 443 would be nice, I think! That would also fix the temporary IPv6 problem and thus federation.

> Malus: no split between federation and client

What are the real-world implications of not splitting them?

927589452:
there is even a guide on it
https://matrix.org/blog/2016/02/10/advanced-synapse-setup-with-lets-encrypt

federation check already returns ok; 
next commit will follow in minutes

fahrplaner:
I tried the (old) guide, with additions of the new guide and my server is running, add user and login with riot works. federation isn't yet (Non-200 response 404 from remote serverNon-200 response 404 from remote server), but we'll see how it goes. I tried DNS and .well-known

here some suggestions for the guide:
1) after installing postgres, the PATH wasn't correct in the venv, so I had to extend the PATH in order to install psycopg2
2) please add the full command for the web backend: uberspace web backend set / --http --port 8008 (correct?)

luto:
> PS: did the mail bounce?

@927589452 drop us one at hallo@uberspace.de

927589452:
>     1. after installing postgres, the PATH wasn't correct in the venv, so I had to extend the PATH in order to install psycopg2

did you source the .bash_profile as described in https://lab.uberspace.de/guide_postgresql.html?#installation Step 3 before creating the venv?

927589452:
>     2\. please add the full command for the web backend: uberspace web backend set / --http --port 8008 (correct?)

uberspace web backend list
/.well-known/matrix apache
/ http:8008 => OK, listening:  ....


so your command should be correct

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+        bind_addresses: ['::','0.0.0.0']
+        x_forwarded: true
+        resources:
+          - names: [client, federation]
+
+And point the ``uberspace web backend`` to it.
+
+.. include:: includes/web-backend.rst
+
+To enable federation as described MatrixFederation_ we need to announcei, that we are listening on port 443 (the reverse proxy), either via DNS or via .well-known.
```
fahrplaner: "announcei" should be "announce"

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+============
+
+We will install synapse using pip, which makes it quite easy,
+but we start with creating a virtual environment for synapse to run in.
+
+.. code-block:: console
+  :emphasize-lines: 1,2,8,18,27,37
+
+  [isabell@stardust ~]$ mkdir -p ~/synapse
+  [isabell@stardust ~]$ virtualenv -p python3.6 ~/synapse/env
```
luto: What is your opinion on installing Synapse with `pip install --user` instead? That way the installation and service setup is easier to handle and less prone to mistakes.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+.. include:: includes/web-backend.rst
+
+To enable federation as described MatrixFederation_ we need to announce, that we are listening on port 443 (the reverse proxy), either via DNS or via .well-known.
+
+DNS announcement
+----------------
+
+The port can can be announced by setting a DNS record in the following format:
+
+.. code-block::
```
luto: This (and the following code block) [are not displayed](https://deploy-preview-418--uberlab.netlify.com/guide_synapse.html#dns-announcement) because the code language is missing:

```
.../guide_synapse.rst:239: WARNING: Error in "code-block" directive:
1 argument(s) required, 0 supplied.

.. code-block::
```

```suggestion
.. code-block:: none
```

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+
+   _matrix._tcp.<yourdomain.com> <ttl> IN SRV <priority> <weight> <port> <synapse.server.name>
+
+For example like this:
+
+.. code-block::
```
luto: see above 

```suggestion
.. code-block:: none
```

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+
+.. note:: this can be checked by running ``dig -t srv _matrix._tcp.my.domain.name``
+
+.well-known announcement
+------------------------
+
+The federation port can also be announced via a file ``~/html/.well-known/matrix/server``
+
+.. code-block:: json
+  {
```
luto: missing empy line between `.. code-block` and the actual, causing this [not to render](https://deploy-preview-418--uberlab.netlify.com/guide_synapse.html#well-known-announcement):

```
.../guide_synapse.rst:256: WARNING: Error in "code-block" directive:
maximum 1 argument(s) allowed, 5 supplied.

.. code-block:: json
  {
    "m.server": "my.domain.name:443"
  }
```

```suggestion

  {
```

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+
+  [isabell@stardust ~]$ mkdir -p ~/synapse
+  [isabell@stardust ~]$ virtualenv -p python3.6 ~/synapse/env
+    Running virtualenv with interpreter /usr/bin/python3.6
+    Using base prefix '/usr'
+    New python executable in /home/matrites/synapse/env/bin/python3.6
+    Also creating executable in /home/matrites/synapse/env/bin/python
+    Installing setuptools, pip, wheel...done.
+  [isabell@stardust ~]$ source ~/synapse/env/bin/activate
+  (env) [isabell@stardust ~]$ pip install --upgrade pip
```
luto: The installation works fine using both the default pip and setuptools. Please skip these steps to make the guide a bit shorter.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+Generate standard config
+------------------------
+
+Generate a config file ``~/synapse/homeserver.yaml`` and replace my.domain.name with the real domain you want your matrix username to end in.
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ source ~/synapse/env/bin/activate
+  (env) [isabell@stardust ~]$ cd ~/synapse
+  (env) [isabell@stardust synapse]$  python -m synapse.app.homeserver \
```
luto: This needs `jinja2`, which is not listed as a dependency. Please install it manually during the installation step, for now. The [bugfix for this](https://github.com/matrix-org/synapse/pull/5514/files) is not yet released :/

```
[matrixx@helin synapse]$ python3.6 -m synapse.app.homeserver \
>   --server-name my.domain.name \
>   --config-path homeserver.yaml \
>   --generate-config \
>   --report-stats=yes
Traceback (most recent call last):
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/matrixx/.local/lib/python3.6/site-packages/synapse/app/homeserver.py", line 64, in <module>
    from synapse.rest import ClientRestResource
  File "/home/matrixx/.local/lib/python3.6/site-packages/synapse/rest/__init__.py", line 32, in <module>
    from synapse.rest.client.v2_alpha import (
  File "/home/matrixx/.local/lib/python3.6/site-packages/synapse/rest/client/v2_alpha/account.py", line 21, in <module>
    import jinja2
ModuleNotFoundError: No module named 'jinja2'
```

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+    --server-name my.domain.name \
+    --config-path homeserver.yaml \
+    --generate-config \
+    --report-stats=[yes|no]
```
luto: As there is a manual change required at this step, please use the [`emphasize-lines` option](https://www.sphinx-doc.org/en/1.5/markup/code.html#line-numbers) to point users to it.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+        resources:
+          - names: [client, federation]
+
+And point the ``uberspace web backend`` on ``/`` to the listener on port 8008.
+
+.. include:: includes/web-backend.rst
+
+To enable federation as described MatrixFederation_ we need to announce, that we are listening on port 443 (the reverse proxy), either via DNS or via .well-known.
+
+DNS announcement
```
luto: Please mark DNS and .well-known as "Option A:" and "Option B:" to make clear that only one of them is needed.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+    --config-path homeserver.yaml \
+    --generate-config \
+    --report-stats=[yes|no]
+  (env) [isabell@stardust ~]$
+
+Set the Synapse_ to listen for federation and clients on port 8008 without encryption in the config file ``~/synapse/homeserver.yaml``.
+
+.. code-block:: yaml
+
+      - port: 8008
```
luto: Please add a bit more context to this step. Something like this would be nice:

```

Set the Synapse_ to listen for federation and clients on port 8008 without encryption in the config file ``~/synapse/homeserver.yaml``. To do this, locate the `listeners:` section and modify the entry with `port: 8008`:

.. code-block:: yaml
  :emphasize-lines: 9

  # Unsecure HTTP listener: for when matrix traffic passes through a reverse proxy
  # that unwraps TLS.
  #
  # If you plan to use a reverse proxy, please see
  # https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.rst.
  #
  - port: 8008
    tls: false
    bind_addresses: ['::', '0.0.0.0']
    type: http
    x_forwarded: true

File: source/guide_synapse.rst [hunk 21]
```diff
@@ -0,0 +1,412 @@
...
+        type: http
+        tls: false
+        bind_addresses: ['::','0.0.0.0']
+        x_forwarded: true
+        resources:
+          - names: [client, federation]
```
luto: The default resource comes with `compress: false`, which is not present here. Since the webserver in front of synapse will handle compression, leaving it in seems reasonable to me.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+.. code-block::
+
+  _matrix._tcp.my.domain.name 3600 IN SRV 10 5 443 my.domain.name.
+
+.. note:: this can be checked by running ``dig -t srv _matrix._tcp.my.domain.name``
+
+.well-known announcement
+------------------------
+
+The federation port can also be announced via a file ``~/html/.well-known/matrix/server``
```
luto: The directory `~/html/.well-known/matrix` does not exist yet. Please add a short `mkdir` step to ensure it's there.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+
+.. code-block:: json
+  {
+    "m.server": "my.domain.name:443"
+  }
+
+This has to be made available under ``/.well-known/matrix`` via the web backend:
+
+.. include:: includes/web-backend.rst
```
luto: The `web-backend.rst` snippet is mainly used to create web backends listening on a port. Since we're working with a static file, this does not apply. Can you add the concrete command, please? Probably `uberspace web backend set /.well-known/matrix/server --apache`.

luto: Maybe we should rename that snippet to `web-backend-port.rst` :sweat_smile:.

source/guide_synapse.rst
```diff
@@ -0,0 +1,412 @@
...
+
+Configure Certificates
```
luto: `server_name` needs to be updated to match the hostname and certificate. Otherwise the federation checker fails with error `MatchingServerName`. Can you please add it? :)

luto: Oh, my bad. It does!

luto:
@927589452 This works very well! Thank you for all the work you put into this guide! I've left comments to improve the formatting and "ease of use" in a few places. Are up for a few more modifications/fixes? :grin:

source/guide_synapse.rst
927589452: thought I already added it

source/guide_synapse.rst
luto: @927589452 what do you thinkk?

luto: If you're not interested in changing the guide, but do generally not oppose the `--user` way of doing things, please say so. I'm happy to do it :grin:

927589452: I can do it; the original source i went from used the venv and the official guide does so too https://github.com/matrix-org/synapse/blob/master/INSTALL.md

927589452:
the original password hash function was referenced by a direkt link; not sure where this will be now

luto:
Commands installed by `pip install --user` end up in `.local/bin`, which is in the standard `$PATH`. So `register_new_matrix_user` and `hash_password` should work just fine.

luto:
Thanks so much for the update! I'll test the guide on a fresh uberspace and the probably merge it. It's a really nice one :)

luto:
since Github won't let me merge this "due to conflicts" and due the high number of commits, I squashed this and set you as the commit-author for the whole thing. Thanks! :tada: it's a very nice guide :)

Issue #414, opened 2019-06-18
supervisord include in all guides


927589452:
the most part for supervisord could be move into a snippet, except the supervisord unit

luto:
Can you name an example? The current supervisord sections (e.g. [guide_etherpad.html#setup-daemon](https://lab.uberspace.de/guide_etherpad.html#setup-daemon)) contain tool-specific paths and the tool name in console output. Generalizing the config part makes it less copy-paste-able. Removing the tool name from console output might confuse reads, but sounds reasonable to reduce clutter.

927589452:
exactly the latter would have been my suggestion
the service name could be replaced with SERVICE leading to a more uniform snippet, as some tools split the parts and put more information into it (e.g. https://lab.uberspace.de/guide_postgresql.html?highlight=postgresql#setup-daemon)
the first part is quite important, as a tool may need to be started on a special way like using the virtualenv's python

927589452:
also the snippet (or maybe the manual https://manual.uberspace.de/daemons-supervisord.html#supervisord ) should hint at supervisord not sourcing any environment variable like the ones from .bash_rc and .bash_profile

luto:
- [ ] look for all guides using supervisord
- [ ] replace the `supervisorctl reread/update/status` part with `.. include:: includes/supervisord.rst`

927589452:
grep supervisord source/* |  awk -F ':' '{print $1 }' | uniq

- [ ] source/guide_autossh.rst
- [ ] source/guide_babybuddy.rst
- [ ] source/guide_buildbot.rst
- [ ] source/guide_codimd.rst
- [ ] source/guide_cryptpad.rst
- [ ] source/guide_django.rst
- [ ] source/guide_embetty.rst
- [ ] source/guide_etherpad.rst
- [ ] source/guide_ghost.rst
- [ ] source/guide_gitea.rst
- [ ] source/guide_gitlab-runner.rst
- [ ] source/guide_hugo.rst
- [ ] source/guide_invoiceninja.rst
- [ ] source/guide_jenkins.rst
- [ ] source/guide_jingo.rst
- [ ] source/guide_mailman-3.rst
- [ ] source/guide_mailman.rst
- [ ] source/guide_miniflux.rst
- [ ] source/guide_mongodb.rst
- [ ] source/guide_mumble.rst
- [ ] source/guide_postgresql.rst
- [ ] source/guide_pretix.rst
- [ ] source/guide_prometheus.rst
[...]

PR #420, opened 2019-06-24
initial try for an example

927589452:
an idea for a snippet like #414


source/includes/supervisord.rst
```diff
@@ -0,0 +1,26 @@
+To make sure the service is restarted after failure and after server reboots,
+configure a ``~/etc/services.d/SERVICE.ini`` for :manual:`supervisord <daemons-supervisord>`:
+
+.. code-block:: ini
+
+  [program:SERVICE]
+  command=%(ENV_HOME)s/bin/SERVICE
```
luto: as noted in the issue, we'd like to keep this part copypasteable.

luto:
thank :)

PR #428, opened 2019-06-26
changed to supervisord snippet

927589452:
Fixing #414

### DIFF
--- a/source/guide_autossh.rst
+++ b/source/guide_autossh.rst
@@ -111,7 +111,7 @@ This will make sure that autossh_ is automatically started if the host reboots b
 
 .. include:: includes/supervisord.rst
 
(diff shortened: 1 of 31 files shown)


927589452:
@luto this fixes #414

luto:
WOHA! Thank you so much!

FriedrichFroebel:
Just out of curiosity: Why does this PR remove `supervisord` from the basic concepts? Shouldn't one still be familiar with it for following the guide?

luto:
It removes it from a couple of guides that claimed to use it, but didn't. Apparently a few ones like django use it, but a bit differently, so they went unnoticed. I'll re-add the note later. Thanks for the heads-up!

927589452:
basically i just searched for supervisord @FriedrichFroebel and if I didn't find any mention in the code i remove the reference
if I found a mention I search for the reread part and replaced it with the include

PR #432, opened 2019-06-28
Added External URL Support section

nowag:
Changing the internal uber.space URL to an external URL with ghost command line as stated in the Ghost manual does not completely work as the users do not have root privileges to run sudo, the config files have to manually be updated as there is a change in the host IP address and the process has to be restarted with ``supervisorctl`` instead of the ghost-cli. Therefore I added this section to clarify such a change of the URLs.

### DIFF
--- a/source/guide_ghost.rst
+++ b/source/guide_ghost.rst
@@ -191,23 +191,21 @@ Update URL with ghost-cli as followed. Change the URL to your external URL in th
  [isabell@stardust ghost]$ ghost config url https://my-domain.com
  [isabell@stardust ghost]$
 
-You now need to readjust your ``~/ghost/config.production.json`` to use the right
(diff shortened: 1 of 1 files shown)


source/guide_ghost.rst
```diff
@@ -174,6 +175,49 @@ Finishing installation
...
+network interface and change the URL for Ghost. Find the following block and change host ``0.0.0.0`` to your Uberspace IPv4 address	e.g. 185.26.156.0 (You can find your IPv4 address in the your Uberspace admin panel under "Datasheet" > "IPv4 address"). Additionally change the URL to your external URL. Change the respective highlighted lines:
+
+.. code-block:: none
+ :emphasize-lines: 2,5
+
+ {
+   "url": "https://my-domain.com",
+   "server": {
+     "port": 2368,
+     "host": "185.26.156.0"
```
nowag: Wrong. Still has to be `0.0.0.0`

nowag: Changed in commit 2 (42849f40c8ab02e1c0b767416d50c43f39e7e07b).

luto:
Thank you! :)

Issue #107, opened 2018-05-03
add a guide for prosody xmpp-server


fapsi:
Hey, I would love to see a guide for [prosody](https://prosody.im/). Since I've currently not enough time, I create this as a remainder (for me), but maybe someone else will already have a look at it.
One can learn from/adapt this [tutorial](https://github.com/alexh-name/plaintext.blog/blob/master/src/hosting/Uberspace/prosody.txt), but I think one should ask for permission before doing that and also credit it.

Requirements regarding the Guide:
- Prosody 0.10 (or trunk?) ([sources](https://prosody.im/nightly/)/[config](https://prosody.im/doc/configure))
- TLS-Only (including file transfer)
- [SQL-Integration](https://prosody.im/doc/modules/mod_storage_sql) (Messages/Users/etc.) 
- Group-Chat (config)
- File-Transfer (config)
- Using [Community-Modules](https://modules.prosody.im/index.html)
- (Auto-)Update-Mechanism (including lua-deps, community-modules)
- ... (TBC)

SalocinHB:
Requires the ability to open ports in the firewall. Not (yet) possible on U7.

fapsi:
Thanks for the ports-feature! As far as I see unfortunately also requires #334 since relevant clients(e.g. conversations) don't support web-sockets (https://github.com/siacs/Conversations/issues/1170) . Anyway I will share my version of the guide (temporary without tls) soon!

taddydevil:
I see the guide was closed because its too complex to support. I had some problems too (connect to other servers failed because invalid ssl connection or so).
Im writing down a simple german [Installation Guide](https://blog.bastlerkarre.de/installation-jabber-xmpp-prosody-auf-uberspace-7/) on my blog with the following [client-to-server](https://xmpp.net/result.php?id=1469588) and [server-to-server](https://xmpp.net/result.php?id=1469590) report. [This plugins](https://compliance.conversations.im/server/bastlerkarre.de/) are active right now. Maybe we have a second chance to get the guide into the lab?

SalocinHB:
@taddydevil Sure, if you open another PR, we'll have a look at it

taddydevil:
Is it possible to use the web backends instead of the port opening? Especially for SRV-Records? My first attempt with the http upload for files failed. I think if its possible the guide should be much more easier.

taddydevil:
can be closed

luto:
https://lab.uberspace.de/guide_prosody.html

PR #212, opened 2018-10-07
Add seafile guide

Finn0811:
Closes issue #133

PR #435, opened 2019-06-29
Guide for prosody xmpp server.

fapsi:
Added guide requested in #107 .
Unfortunately I didn't get my .rockspec as official one on luarocks and had to use some ugly workaround for setting the C99-flag (because luarocks has some limitations in their .rockspec format; see request/progress in https://github.com/luarocks/luarocks/issues/129#issuecomment-494843403 ). But maybe you could provide some input on this.

### DIFF
--- a/source/guide_prosody.rst
+++ b/source/guide_prosody.rst
@@ -127,18 +127,6 @@ Further optional ones (``luadbi-mysql``, ``lua-zlib`` and ``luaevent``) can be i
 
 To list the installed packages with their versions use the command ``luarocks list``.
 
-Set c99 compilation flag
-------------------------
-In order to built prosody we have to activate C99-mode as global setting for all packages compiled via luarocks ``~/.luarocks/config-5.1.lua``.
-
-::
-
- variables = {
- CC= "gcc -std=c99"
- }
-
-.. note:: This is a temporary workaround since luarocks not allowing setting this flag via ``.rockspec``-file.
-
 MySQL-database
 --------------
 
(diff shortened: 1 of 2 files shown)


source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+The following domains should be set up (to set up prosody for the domain ``stardust.space``):
+
+::
+ 
+ [isabell@stardust ~]$ uberspace web domain list
+ stardust.space
+ groupchat.stardust.space
+ [isabell@stardust ~]$
+
+Additionally to the dns_ A-records we need the following SRV-records:
```
luto: Please add the `A` and `AAAA` records to the table below.

source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+| _xmpp-server._tcp.groupchat.stardust.space  | 18000 | IN    | SRV  | 0        | 5      | SERVERPORT   | stardust.uber.space   |
++---------------------------------------------+-------+-------+------+----------+--------+--------------+-----------------------+
+
+.. note:: Adapt the port-numbers and change ``stardust`` as well as ``stardust.space`` accordingly in your dns settings of your prefered provider.
+
+TLS
+---
+For simplicity we use openssl and the certificates created via letsencrypt_ for above domains which are already provided on the host after adding the domains and visiting them.
+
+Configure luarocks
```
luto: Hm, these environment variables should really be set automatically by us for all uberspaces. I'll create an issue internally, so we can skip this section here.

source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+ [isabell@stardust ~]$
+
+Install lua-dependencies
+------------------------
+
+The following dependencies_ (``luasocket``, ``luaexpat`` and ``luafilesystem``) are required:
+
+::
+
+ [isabell@stardust ~]$ luarocks install luasocket --local && \
```
luto: Please split these commands into four lines with their own prompt.

source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+------------------------
+In order to built prosody we have to activate C99-mode as global setting for all packages compiled via luarocks ``~/.luarocks/config-5.1.lua``.
+
+::
+
+ variables = {
+ CC= "gcc -std=c99"
+ }
+
+.. note:: This is a temporary workaround since luarocks not allowing setting this flag via ``.rockspec``-file.
```
luto: Is there any upstream bug for this? if yes, please link it.

source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+
+MySQL-database
+--------------
+
+Create a separate db for prosody data:
+
+::
+
+ [isabell@stardust ~]$ mysql -e "CREATE DATABASE ${USER}_prosody CHARACTER SET utf8"
```
luto: By default we use `utf8mb4`, which supports emojis and a lot of other characters. The old `utf8` charset can only store 3 bytes, which causes problems. Is there a reason why we need `utf8` here?

source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+ [isabell@stardust ~]$
+
+Installation
+============
+
+Install prosody with an (currently unofficial) ``.rockspec``-file:
+
+::
+
+ [isabell@stardust ~]$ luarocks install --server=http://luarocks.org/manifests/fapsi prosody --local
```
luto: This is your personal luarocks manifest, right? Guides should use the official sources, if at all possible. Is there a way to install prosody using the official sources only? :/

source/guide_prosody.rst
```diff
@@ -0,0 +1,309 @@
...
+
+::
+
+ [isabell@stardust ~]$ supervisorctl reread
+ prosody: available
+ [isabell@stardust ~]$
+
+And then update and start your new service:
+
+::
```
luto: Replace this with the new `.. include:: includes/supervisord.rst`, please :)

source/guide_prosody.rst
fapsi: Added!

source/guide_prosody.rst
fapsi: Let me know what to change here.

source/guide_prosody.rst
fapsi: Added!

source/guide_prosody.rst
fapsi: Not necessary anymore since we are building from source now.

source/guide_prosody.rst
fapsi: No specific reason, I was just unaware of this problems.. I removed the definition completely, since other guides don't mention it at all and trusting your default values, right? Note that the seafile guide also uses the utf8 definiton..

source/guide_prosody.rst
fapsi: I thought that it would be prettier to use luarocks for the prosody installation step too, because it's easier to maintain and update, but you're right with using the official sources. Since luarocks has the mentioned problems the guide looks clearer now by building from the sources.
Another not quite related question: Since prosody 0.12 will only support Lua 5.2, are there any plans to provide it on CentOS 7?

source/guide_prosody.rst
fapsi: Done!

source/guide_prosody.rst
luto: Seafile doesn't (or didn't) support utf8mb4, see https://github.com/Uberspace/lab/pull/212#discussion_r224017740.

source/guide_prosody.rst
luto: you can skip the entire "Configure luarocks" block :) we'll take care of that globally in the next release.

nichtmax:
I'm really sorry but this is too complex to support, there hasn't been _one_ of our team who got it working. [Shoot us a mail](mailto:hallo@uberspace.de), you'll get your reward anyway of course ☺️

Issue #433, opened 2019-06-29
GitLab Runner: wrong path gitlab-runner-11.1.0


klml:
[Configure the supervisord service](https://github.com/Uberspace/lab/blob/master/source/guide_gitlab-runner.rst#configure-the-supervisord-service) creates a shell script with `/home/$USER/bin/gitlab-runner-11.1.0` but there is only the file `/home/$USER/bin/gitlab-runner`.
Works with `gitlab-runner` in the script for me.

@bytekeks do you want a PR with the versioned path `gitlab-runner-11.1.0` or the plain path `gitlab-runner`?

bytekeks:
Yes, you are right.

Please see the section `Download latest version`

```
[isabell@stardust ~]$ wget -O /home/$USER/bin/gitlab-runner https://gitlab-runner-downloads.s3.amazonaws.com/latest/binaries/gitlab-runner-linux-amd64
```

wget's output file does not contain a version.

@klml Yes, you can remove the version from the line in `gitlab-runner.sh`, if you prefer.

Thanks for submitting this issue :)

klml:
thx for merging.

Can we close this issue?

PR #436, opened 2019-06-30
remove the version from the line in gitlab-runner.sh #433

klml:
remove the version from the line in gitlab-runner.sh , as discussed in #433


luto:
Thanks!

Issue #437, opened 2019-06-30
GitLab Runner: gitlab-runner.ini does not accept $USER


klml:
I got the error, supervisord did not found `/home/$USER/bin/gitlab-runner.sh`, yet it exists.

```
[klmlrun@hartley ~]$ supervisorctl status
gitlab-runner                    FATAL     can't find command '/home/$USER/bin/gitlab-runner.sh'
```

I fixed it for me, by expanding $USER inside the .ini
```
[klmlrun@hartley ~]$ sed -i s/\$USER/$USER/g etc/services.d/gitlab-runner.ini 
[klmlrun@hartley ~]$ cat etc/services.d/gitlab-runner.ini 
[program:gitlab-runner]
command=/home/klmlrun/bin/gitlab-runner.sh
[klmlrun@hartley ~]$ supervisorctl update
gitlab-runner: stopped
gitlab-runner: updated process group
[klmlrun@hartley ~]$ supervisorctl status
gitlab-runner                    RUNNING   pid 15165, uptime 0:00:05
```
I am not sure, but does supervisord using bash variables? 
@bytekeks  Should we expand $USER?

Does anybody has a better way than sed?

luto:
there is `$(ENV_HOME)s` :)

klml:
@luto thanks ;) 

I found this
>;  - Shell expansion ("~" or "$HOME") is not supported.  Environment
> ; variables can be expanded using this syntax: "%(ENV_HOME)s".

from [supervisor](https://github.com/Supervisor/supervisor/blob/master/supervisor/skel/sample.conf#L7)

PR #438, opened 2019-07-01
GitLab Runner: gitlab-runner.ini does not accept $USER

klml:
GitLab Runner: gitlab-runner.ini does not accept $USER #437


luto:
Thanks!

PR #449, opened 2019-07-13
[mosca] Add guide for Mosca (Node.js MQTT broker)

### LINKED ISSUE #448: Guide for Mosca
[Mosca](https://github.com/mcollina/mosca) is a Node.js MQTT broker.

I’m working on a guide since it was a bit of a struggle to set this up properly.

mischah:
Hej,

here is the guide for installing and running [Mosca](https://github.com/mcollina/mosca) an Uberspace.

Cheers, Michael

---

Closes #448


luto:
Node.js 6.x has been EOL for more than two months and is no longer receiving any security updates. Since we do not want to provide users with resources to run insecure software, I'm afraid we cannot include this guide. Once Mosca has been updated to support a non-EOL node.js version, I'm more than happy to take a closer look, though!

Issue #361, opened 2019-04-11
CodiMD is not able to build anymore


luksal:
I'm trying to install the latest CodiMD version 1.3.2, but the installation always breaks when web pack tries to use the terser plugin. In the lab manual it's stated that you should remove the ParallelUglifyPlugin but it's not used anymore.
After executing `node run build` I get the following error message:
```
> CodiMD@1.3.2 build /home/lsv7/codimd
> webpack --config webpack.prod.js --progress --colors --bail

 96% [0] chunk asset optimization TerserPluginGetötet
```

luksal:
Quick workaround: If you copy the `webpack.dev.js`to `webpack.prod.js`, then the build is running fine

luto:
This guide has seen a lot of changes since this issue was opened. If it is still broken, please open a new issue.

PR #451, opened 2019-07-18
[CodiMD] Make installation work for the latest release v1.4.0

makomi:
Known issue: One should try to improve the part about removing `UglifyJS`. This is related to #361.

### DIFF
--- a/source/guide_codimd.rst
+++ b/source/guide_codimd.rst
@@ -372,7 +372,7 @@ if you are having problems remove the ``~/codimd/`` and move ``~/codimd-old/`` b
 
 
 .. _HackMD: https://hackmd.io/
-.. _CodiMD: https://github.com/codimd
+.. _CodiMD: https://github.com/codimd/server
 .. _release: https://github.com/hackmdio/codimd/releases
 
 ----
--- a/source/guide_codimd.rst
+++ b/source/guide_codimd.rst
@@ -200,7 +200,7 @@ Also, comment out each line of ``public/vendor/ot/compress.sh`` with a ``#``:
   #./ajax-adapter.js \
   #./editor-client.js
 
-.. note:: This is how it worked at the time of writing and it might be different for the version you downloaded. If it is different, try commenting out everything related to ``UglifyJS``. If the compilation fails the first time due to RAM limitations, just start it again. For me, it always worked the second time.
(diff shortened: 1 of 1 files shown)


source/guide_codimd.rst
```diff
@@ -28,182 +30,266 @@ Inspired by Hackpad, with more focus on speed and flexibility, and built from Ha
...
-Your blog URL needs to be setup:
+URL setup
+---------
+
+Your URL needs to be set up:
 
 .. include:: includes/web-domain-list.rst
 
-Install the package manager ``yarn`` using the node.js package manager (``npm``):
+See :manual:`domains <web-domains>` for setting up a custom URL like ``write.example.com``:
```
luto: All guides should use `isabella.uber.space` as an example domain, that does not need to be added manually.

makomi: I know, that is why I included the instructions for `isabell.uber.space` as required. In addition I want users to be able to configure the service for a custom domain.

luto: Do the steps for `isabell.uber.space` differ from `write.example.com`, except for the domain name? Adding domains is explained [in the manual](https://manual.uberspace.de/web-domains.html) and shouldn't be repeated in guides. Is there a reason for a custom domain I seem to be missing?

makomi: Yes, they differ in two cases and including this additional information makes the setup easier:

**(1)** While `isabell.uber.space` is set up automatically when creating a Uberspace account, a custom domain needs to be set up manually. Of course, this information can be looked up in the Uberspace manual, but since this section of the guide is called "URL setup", I want to provide the user with the necessary command for setting up their URL - whatever it might be.

**(2)** In the guide's section called "[Configure web backend](https://deploy-preview-451--uberlab.netlify.com/guide_codimd.html#configure-web-backend)", at least for me, it was not obvious how to configure the web backend of a custom domain. The same argument as for (1) applies here as well.

I am not familiar with the other guides you host as part of UberLab 7. If adding this additional information goes against the style you use, I will remove the information. Yet, I would much rather keep it in there and make the guide easier to follow.

makomi: Please have a look at commit 15502cc. It should resolve this issue.

luto:
Thank you! :tada:

PR #452, opened 2019-07-20
corrected wording to avoid misunderstandings

DJCrashdummy:
although both are stored in the admin database, [root](https://docs.mongodb.com/manual/reference/built-in-roles/#root) and [admin](https://docs.mongodb.com/manual/reference/built-in-roles/#dbAdminAnyDatabase) users are 2 different things.


luto:
Thanks!

Issue #380, opened 2019-05-06
Guide for ezmlm


behrisch:
I would like to write a guide for ezmlm installation and usage.

luto:
https://lab.uberspace.de/guide_ezmlm.html

PR #457, opened 2019-07-28
adding ezmlm guide, refs #380

behrisch:


source/guide_ezmlm.rst
```diff
@@ -0,0 +1,159 @@
...
+ ezmlm-send:           OK
+ ezmlm-send (from):    OK
+ ezmlm-send trailer:   OK
+ omitbottom (-manage): OK
+ omitbottom (-get):    OK
+ Cleaning up...
+ [isabell@stardust ezmlm-idx-7.2.2]
+
+
+The test call should show that everything is OK. Also ``man ezmlm-make`` should work.
```
luto: This fails for me, because `~/man` is empty (so is `~/bin`). I guess this is missing a `make install`, which fails due to missing directories:

```
./installer "`head -n 1 conf-bin`" < BIN
./installer "`head -n 1 conf-etc`" < ETC
./installer "`head -n 1 conf-man`" < MAN
./installer "`head -n 1 conf-lib`" < LIB
installer: fatal: unable to mkdir /home/isabell/lib/ezmlm: file does not exist
make: *** [install] Error 111
```

After a quick `mkdir ~/lib` + `make install`, it works just fine. Can you please add those missing steps? :)

luto:
woop! thank you!

PR #461, opened 2019-08-20
Update the Hugo guide

jlnostr:
Changes with this PR:

- Updated version of Hugo
- Set the `HUGO_CACHEDIR`, so that the build will succeed

### DIFF
--- a/source/guide_hugo.rst
+++ b/source/guide_hugo.rst
@@ -2,6 +2,7 @@
 
 .. author:: Christian Kantelberg <uberlab@mailbox.org>
 .. author:: luto <http://luto.at>
+.. author:: Julian Oster <https://jlnostr.de>
 
 .. tag:: lang-go
 .. tag:: blog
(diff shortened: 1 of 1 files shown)


SalocinHB:
What makes you think you can't access `/tmp`?

jlnostr:
As stated in the second commit message:

> If not provided, Hugo fails with the following error message:
>
> Error: add site dependencies: create deps: failed to create file caches
> from configuration: mkdir /tmp/hugo_cache/modules: permission denied
> 
> Therefore the environment variable is set before the execution.

Setting the environment variable `HUGO_CACHEDIR` helps.

SalocinHB:
Ah, I see what the problem is. I tested this just now on an otherwise empty host and it just worked. But of course you're on a host with other users, so if one of them already used this guide, they would have created /tmp/hugo_cache and now own that directory, so you can't create it yourself.

PR #463, opened 2019-08-26
Guide WsgiDAV (webdav server)

SeBaBu:


source/guide_wsgidav.rst
```diff
@@ -0,0 +1,200 @@
...
+
+ ~/.wsgidav/wsgidav.yaml
+ 
+
+
+Config file:
+
+::
+
+ server: "cheroot"
```
luto: Full configuration dumps don't tend to age well with updates. Can you please start with the default config and then outline what needs to be changed? That makes sure that your guide will work well even in a year or two :)

source/guide_wsgidav.rst
```diff
@@ -0,0 +1,200 @@
...
+
+Configure ``supervisord``
+=========================
+
+Create ``~/etc/services.d/wsgidav.ini`` with the following content:
+
+.. code-block:: ini
+
+ [program:wsgidav]
+ command=wsgidav -c /home/isabell/.wsgidav/wsgidav.yaml
```
luto: You can use `%(ENV_HOME)s` to get rid of the static "isabell" here and on the line below. That way users can just copy/paste your config :)

source/guide_wsgidav.rst
```diff
@@ -0,0 +1,200 @@
...
+The following template is the minimum to get wsgidav up and running. You have to change provider_mapping, user_mapping and the port (if another service is already running on 8080).
+If you miss some functionality or want more information, take a look at the sample file and the explanation of the configuration options:
+https://wsgidav.readthedocs.io/en/latest/user_guide_configure.html#sample-wsgidav-yaml
+
+
+Config file path:
+
+::
+
+ ~/.wsgidav/wsgidav.yaml
```
luto: Since `~/.wsgidav` isn't in the config search path of wsgidav anyway, what do you think about putting the config file in the already present `~/etc`?

source/guide_wsgidav.rst
```diff
@@ -0,0 +1,200 @@
...
+     accept_basic: true
+     accept_digest: true
+     default_to_digest: true
+     trusted_auth_header: null
+     domain_controller: null
+ 
+ nt_dc:
+     preset_domain: null
+     preset_server: null
+ pam_dc:
```
luto: Enabling PAM will allow any unrelated user on the system to login using their SSH username/password. Please disable it.

luto:
Thanks for the new guide :) I've left a few comments on what I think needs tweaking. Please let me know what you think!

PR #465, opened 2019-08-31
Add Contao Manager as alternative install method

hendrikmaus:
On a fresh Uberspace U7 account, I could not allocate enough memory to run the `composer` call locally. The obvious `php -d memory_limit=-1` does not do the trick, as well as a custom php.ini config. Therefore I experimented with the Contao Manager beta product and contribute the successful outcome.


luto:
ooh, very nice. We're kind of bummed out by the memory requirements of composer :/ after some debugging, it seems to be "normal" for it to use multiple gigs of RAM just to install a few packages -- or rather: resolve the dependencies for a few packages. well..

Thanks for the addition!

PR #468, opened 2019-09-05
Update guide_codimd.rst

kimdiallo:
without `NODE_OPTIONS=--max_old_space_size=1024`npm run build is killed by OOM

### DIFF
--- a/source/guide_codimd.rst
+++ b/source/guide_codimd.rst
@@ -209,7 +209,7 @@ Now, let's build the front-end bundle. This will take some time.
 
 .. code-block:: console
 
-  [isabell@stardust codimd]$ npm run build
+  [isabell@stardust codimd]$ NODE_OPTIONS=--max_old_space_size=1024 npm run build
   [...]
   [isabell@stardust codimd]$
 


SalocinHB:
Can't get this to build with the option either 
```
[testnico@stardust codimd]$ NODE_OPTIONS=--max_old_space_size=1024 npm run build
npm WARN lifecycle The node binary used for scripts is /bin/node but npm is using /opt/nodejs8/bin/node itself. Use the `--scripts-prepend-node-path` option to i
nclude the path for the node binary npm was executed with.

> CodiMD@1.5.0 build /home/testnico/codimd
> webpack --config webpack.prod.js --progress --colors --bail

 96% [0] chunk asset optimization TerserPluginKilled
[testnico@stardust codimd]$
```

nichtmax:
Does not work

makomi:
Did you try it a second time right after? See the note in line 203 stating "[...] If the compilation fails the first time due to RAM limitations, just start it again. It tends to work the second time."

For me, it always failed the first time and always worked the second time. Would this be acceptable for you?

Personally, I wouldn't invest time to fix this and keep the guide online, because it always works the second time and I am not familiar with the tech stack, i.e. I would likely have to spend a lot of time to fix it and not significantly improve the result. I acknowledge that it is a dirty solution - and yet - a reliably working solution.

If it no longer works by just running the build command a second time, it is likely that CodiMD's code changed and additional modifications have to be made, to make the guide work again. This would be the second reason I can think of for why it does not build for you. I would like to understand which of the two it is.

luksal:
Works for me with option `NODE_OPTIONS=--max_old_space_size=1024` and CodiMD version 1.5.0

PR #469, opened 2019-09-06
Add Gmail hint

bhase:
Inform the user about issue when testing the new created mailing list with gmail addresses


SalocinHB:
Oh, great tip!

PR #477, opened 2019-09-11
[HumHub] add new guide humhub

creissig:
finished guide for humhub


nichtmax:
Thanks for the guide! I fixed up a few things here and there. Check your mails for your reward ☺️

Issue #488, opened 2019-10-07
Guide for Shopware 5/6


Jaglag:
Could someone add a guide to proper Install Shopware 5/6?

https://www.shopware.com/de/download/

Thank you

Jaglag:
Thank you for adding the guide.
I waiting for the shopware 6 fix.

PR #500, opened 2019-10-14
[shopware-6] Add guide for Shopware 6

ihmels:
This PR adds a guide for Shopware 6 (#488).

The reason for creating a separate guide is that Shopware 6 has been completely rewritten and the installation is different.

~~Currently there seems to be a bug in the installation script with PHP 7.2. It's also in the early access stage, so I'm waiting for the stable release.~~


luto:
Shopware 6 seems to be still in early access, but I don't think that's much of a problem. If you'd like to try with PHP 7.3 to circumvent the installer bug, I'd be happy to include your guide!

ihmels:
I've tried the installation again on a fresh Uberspace with Shopware 6.0.0-ea2 and it worked.

luto:
nice, thank you!

PR #502, opened 2019-10-22
[resilio-sync] Add step to open a firewall port for Resilio Sync

nielsmde:
Opening a custom port allows Resilio Sync to establish direct connections to peers,
instead of relying on a relay server.

Setting the "listening_port" has to be done through a configuration
file. For consistency, all command line arguments  are included in this file.


nichtmax:
Thanks!

PR #509, opened 2019-10-30
install redmine with mysql using puma and web backends

fahrplaner:


luto:
@SalocinHB any status on this? does it work? :grin:

PR #514, opened 2019-11-11
Add <username> as remote user

zvyn:
In case local and remote usernames differ the remote username has to be specified in the mosh address.


luto:
thank you!

PR #521, opened 2019-11-28
Update db pw to match the default one

splitt3r:
Just a very very small change. This way it matches the default settings from https://github.com/Uberspace/lab/blob/master/source/includes/my-print-defaults.rst.


luto:
Thanks!

PR #526, opened 2019-12-12
[minio] add guide for minio

mariusbertram:
The guide is finished.


source/guide_minio.rst
```diff
@@ -0,0 +1,152 @@
...
+
+
+Prerequisites
+=============
+Download the latest MinIO binary into your bin folder and make it executable:
+
+.. code-block:: console
+
+    [isabell@stardust ~]$ wget https://dl.min.io/server/minio/release/linux-amd64/minio -O /home/isabell/bin/minio
+    [isabell@stardust ~]$ chmod +x /home/isabell/bin/minio
```
luto: Please `~/bin` here and in the previous line, so the commands can be copy-pasted more easily.

source/guide_minio.rst
```diff
@@ -0,0 +1,152 @@
...
+    [isabell@stardust ~]$
+
+Create a folder for your data:
+
+.. code-block:: console
+
+    [isabell@stardust ~]$ mkdir /home/isabell/minio
```
luto: Just use `mkdir minio` here or `mkdir ~/minio` here. All guides assume that the sessions starts out in the home dir.

source/guide_minio.rst
```diff
@@ -0,0 +1,152 @@
...
+.. note:: MinIO default port is 9000
+
+Setup config
+============
+
+Run Minio once to get the master AccessKey and SecretKey:
+
+.. code-block:: console
+
+    [isabell@stardust ~]$ /home/isabell/bin/minio server /home/isabell/minio
```
luto: Since `minio` is in `~/bin`, the much shorter `minio server ~/minio` (or `$HOME/minio` in case the substitution doesn't work as excepted) should work here.

source/guide_minio.rst
```diff
@@ -0,0 +1,152 @@
...
+    [isabell@stardust ~]$
+
+Setup Supervisord
+=================
+Create the configuration ``~/etc/services.d/minio.ini``:
+
+.. code-block:: ini
+
+    [program:minio]
+    command=~/bin/minio server ~/minio
```
luto: This should work with `command=minio (...)`, because it's in the `$PATH`.

source/guide_minio.rst
luto: Oh, minio auto-creates the directory. So we can skip this step.

source/guide_minio.rst
luto: `~` doesn't work here, please use `%(ENV_HOME)s/minio` as the parameter.

mariusbertram:
I've commited your suggestions.

source/guide_minio.rst
mariusbertram: > `~` doesn't work here, please use `%(ENV_HOME)s/minio` as the parameter.

`~` does work just checked it on biela

luto: I just checked it on machholz :sweat_smile: all other guides use `$(ENV_HOME)s`, so it seems nicer for consistency's sake in any case.

luto:
Perfect! Thanks for you changes!

PR #538, opened 2020-01-10
Add guide for aptly

### LINKED ISSUE #368: Guide for aptly
https://www.aptly.info/

> aptly is a swiss army knife for Debian repository management: it allows you to mirror remote repositories, manage local package repositories, take snapshots, pull new versions of packages along with dependencies, publish as Debian repository.

I installed this on an uberspace yesterday. I'm going to write a guide soon.

Thomas--S:
Closes #368.

I didn't add a logo for now because I'm not sure if I can just copy it for this guide (copyright etc.).

### DIFF
--- a/source/guide_aptly.rst
+++ b/source/guide_aptly.rst
@@ -1,0 +1,339 @@
+.. highlight:: console
+
+.. author:: Thomas S. <https://github.com/Thomas--S/>
+
+.. tag:: package-management
+
+#####
+aptly
+#####
+
+.. tag_list::
+
+aptly_ is described on its website like this:
+    *aptly is a swiss army knife for Debian repository management: it allows you to mirror remote repositories, manage local package repositories, take snapshots, pull new versions of packages along with dependencies, publish as Debian repository.*
+
+----
+
+
+License
+=======
+
+License information can be found here
+
+  * https://github.com/aptly-dev/aptly/blob/master/LICENSE
+
+Prerequisites
+=============
+
+First of all, you'll need to install Homebrew_. This tool will be used to install aptly.
+
+.. code-block:: console
+ :emphasize-lines: 28
+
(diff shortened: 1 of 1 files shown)


source/guide_aptly.rst
```diff
@@ -0,0 +1,339 @@
...
+=======
+
+License information can be found here
+
+  * https://github.com/aptly-dev/aptly/blob/master/LICENSE
+
+Prerequisites
+=============
+
+First of all, you'll need to install Homebrew_. This tool will be used to install aptly.
```
luto: aptly provides [static builds](https://www.aptly.info/download/), so I don't think there is a need install brew here. What do you think about just downloading the `.tar.gz` and unpacking the binary to `~/bin`?

Thomas--S: I've chosen brew because it seemed to be the more convenient approach, especially regarding updates.

luto: On U6 brew caused a lot of headaches in support, because it tends to rebuild all kinds of dependencies. So, if we can, we like to avoid it. Considering that there are static builds, which do not need any kind of compilation, we'd rather use those. That way the guide has fewer moving parts.

If you don't want to put in the extra work, I understand. In that case I'll just change things around. Let me know, please!

Thomas--S: I can change the guide according to your suggestion, but most likely I won't be able to do so before the 10th of February. I hope this is okay for you.

Thomas--S:
As recommended, I changed the guide to use the binary distribution instead of brew.
Please tell me, if further changes are needed.

Thanks for you patience.

Spacelord09:
https://github.com/Uberspace/lab/blob/master/CONTRIBUTING.md

Spacelord09:
https://github.com/Uberspace/lab/blob/master/STYLE.md

luto:
@Thomas--S please shoot us an email to hallo@uberspace.de for your reward :)

Issue #458, opened 2019-07-28
Add guide for Flask


cmacht:
I've set up Flask with the help of the existing Django guide, but would be happy to share my tweaks for uwsgi and some of the decisions I made to get Flask running that differ from it. Would writing a Flask guide be helpful?

SparksFlyx3:
Please share! I spent 8 hours & didn't get my Flask App deployed on Uberspace..

cmacht:
Hi Martina, it's still on my list. It's very similar to setting up Django, maybe this might help in the meanwhile: https://lab.uberspace.de/guide_django.html

SparksFlyx3:
Didn't work for me. I switched to Heroku - easy deployment within 30 minutes.

tomaculum:
It did not work for me either, since I got strange module not found errors even though the modules were installed.

What I did to get it running:

1. Django guide step 1

2. setup project folder and virtual environment

   ```
   mkdir ~/MyProject
   cd ~/MyProject
   python3.6 -m venv myvenv
   source myvenv/bin/activate
   pip install uwsgi flask
   ```

3. setup basic flask app

   > setup app.py

   ```
   from flask import Flask, request

   app = Flask(__name__)

   @app.route('/')
   def hello_world():
       return 'Hello World!'

   if __name__ == '__main__':
       app.run()
   else:
       application = app
   ```

4. Django "Configure web server" step:

   > replace content of setup daemon .ini file:

   ```
   [uwsgi]
   base = /home/<username>/MyProject
   chdir = /home/<username>/MyProject

   http = :8000
   master = true

   home = %(base)/myvenv
   pythonpath = %(base)/myvenv

   module = app
   callable = app

   uid = <username>
   gid = <username>

   ```

5. edit line in `uwsgi.ini` from step 1
   ```
   command=/home/<username>/MyProject/myvenv/bin/uwsgi --master --emperor %(ENV_HOME)s/uwsgi/apps-enabled
   ```


Especially the 5th step doesn't seem quite right since the uswgi.ini file shouldn't be linked to a specific project. 

I'm looking forward to your guide and maybe in the meantime this saves someone some time.

luto:
https://lab.uberspace.de/guide_flask.html

PR #552, opened 2020-02-03
[flask] add guide for flask

bwiessneth:
Added guide for setting up flask with virtualenv and uwsgi. #458


luto:
Thank you! :tada:

PR #561, opened 2020-02-14
Switch postorius and hyperkitty to prevent Django3

pheraph:
See https://github.com/Uberspace/lab/issues/535#issuecomment-573645274


luto:
pheww.. that's more of a dirty work around, than a fix. I'll take a deeper look.

PR #563, opened 2020-02-21
use lc-collate and lc-ctype C during createdb

lub:
It's the only supported configuration in recent Synapse versions.

https://github.com/matrix-org/synapse/blob/master/docs/postgres.md#set-up-database

### DIFF
--- a/source/guide_synapse.rst
+++ b/source/guide_synapse.rst
@@ -45,6 +45,8 @@ You need a running :lab:`Postgresql <guide_postgresql>` database server, a dedic
   Enter it again:
   [isabell@stardust ~]$ createdb \
     --encoding=UTF8 \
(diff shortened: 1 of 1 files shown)


source/guide_synapse.rst
```diff
@@ -45,6 +45,8 @@ You need a running :lab:`Postgresql <guide_postgresql>` database server, a dedic
   Enter it again:
   [isabell@stardust ~]$ createdb \
     --encoding=UTF8 \
+    ----lc-collate=C \
+    ----lc-ctype=C \
```
luto:
```suggestion
    --lc-collate=C \
    --lc-ctype=C \
```

The `--help` shows these options with just two dashes (`--`), do we really need four of them?

PR #564, opened 2020-02-21
remove pg env vars from synapse.ini

lub:
These are not needed, as all PostgreSQL credentials are configured via homeserver.yaml


luto:
That's right! Can you please also update [the section](https://deploy-preview-564--uberlab.netlify.com/guide_synapse.html#configure-certificates) to mention the DB settings in the text above the config block? Or, better yet, spit it into "Configure Certificates" and "Configure Database Access".

luto:
Merged manually, thanks! :tada:

PR #573, opened 2020-03-07
Update guide_selfoss.rst

cblte:
added section if user wants to put selfoss into a subdomain.

### DIFF
--- a/source/guide_selfoss.rst
+++ b/source/guide_selfoss.rst
@@ -149,6 +149,11 @@ You can now log in to your installation and start adding and reading feeds.
 
 Check out the official `selfoss documentation`_ for explanation of further configuration parameters.
 
+Subdomain
+---------
+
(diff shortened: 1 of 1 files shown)


SalocinHB:
Setting up subdomains is explained in the manual.

cblte:
this is not about adding or setting up a subdomain. it describes a configuration setting you need to set when you use Selfoss with a subdomain and not in a subdirectory to html. Maybe this was a misunderstanding.

nichtmax:
All guides are for a fresh account and all tools have to be installed into the main DocumentRoot due to simplicity and uniformity.

cblte:
> All guides are for a fresh account and all tools have to be installed into the main DocumentRoot due to simplicity and uniformity.

I fully agree. I added this just as additional information. maybe the wording is not the best. will think about it and redo.

PR #583, opened 2020-03-19
Docs: update Matomo installation manual

gustavpursche:
While the old method still works the recommended official way looks a [little different now](https://matomo.org/docs/installation/).

I've also deleted some of the output `wget` creates, since it isn't helpful for the installation and clutters the manual.

### DIFF
--- a/source/guide_matomo.rst
+++ b/source/guide_matomo.rst
@@ -29,7 +29,7 @@ Matomo_ (formerly known as Piwik) is an open source website tracking tool (like 
 Prerequisites
 =============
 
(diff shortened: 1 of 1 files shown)


source/guide_matomo.rst
```diff
@@ -34,7 +34,7 @@ We're using :manual:`PHP <lang-php>` in the stable version 7.1:
 ::
 
  [isabell@stardust ~]$ uberspace tools version show php
- Using 'PHP' version: '7.1'
+ Using 'PHP' version: '7.2'
```
SalocinHB: Please change the version in the preceding paragraph as well

Issue #586, opened 2020-03-21
Guide for Wordpress with Bedrock


herzogmedia:
Hi, I'm writing a guide for using Wordpress with [Bedrock](https://roots.io/bedrock/).

PR #587, opened 2020-03-22
[wp-bedrock] Added Guide for Wordpress with Bedrock

herzogmedia:
As announced in #586 here is my guide for using bedrock on uberspace. It is partly based on the standard wordpress guide, which I hope is ok for you.


source/guide_wordpress-bedrock.rst
```diff
@@ -0,0 +1,248 @@
...
+.. note:: You can leave the ``WP_ENV`` Setting on ``development`` for now, but don't forget to set it on ``production`` when launching your site.
+  See the Bedrock-Documentation_ for more Info on Environment-Settings
+
+You now need to set your :manual:`document root <web-documentroot>` to the ``bedrock/web/`` directory. To do this, we delete the standard document root folder and create a symlink instead.
+
+.. warning:: Please make sure your DocumentRoot is empty before removing it. This step will delete all contained files.
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ rm -r /var/www/virtual/$USER/html
```
luto: All guides should assume a fresh, empty uberspace. Please change this to `rmdir ...`, so the commend fails when this is not the case. 

```suggestion
.. code-block:: console

 [isabell@stardust ~]$ rmdir /var/www/virtual/$USER/html

```

source/guide_wordpress-bedrock.rst
```diff
@@ -0,0 +1,248 @@
...
+ [isabell@stardust ~]$ ln -s /var/www/virtual/$USER/bedrock/web /var/www/virtual/$USER/html
+ [isabell@stardust ~]$
+
+
+Point your browser to your domain, ``https://isabell.uber.space/wp/wp-admin`` in this example, to start the wordpress installation process. Here you will
```
luto: This fails because wordpress doesn't want to create the DB itself. Please add a step in "Prerequisites" creating it like so:

```
$ mysql -e "CREATE DATABASE ${USER}_wp"
```

herzogmedia: ok, changed it and also the wording for the .env-Configuration to match.

herzogmedia:
Thanks for reviewing and glad you like it!

luto:
Thanks for the swift changes!

PR #590, opened 2020-03-24
Kimai: Bump versions, add update of app_secret

ebroda:
I just installed Kimai on my U7 and updated the lab guide regarding version and secret during the installation. 1.8 also works fine on U7.


luto:
Thank you! :tada:

PR #595, opened 2020-03-27
[roundcube] add guide for roundcube

pxlfrk:
Hi, i've added a guide for [Roundcube](https://roundcube.net/).

closes#593

### DIFF
--- a/source/guide_roundcube.rst
+++ b/source/guide_roundcube.rst
@@ -1,0 +1,222 @@
+.. highlight:: console
+
+.. author:: Achim | pxlfrk <hallo@pxlfrk.de>
+
+.. tag:: lang-php
+.. tag:: web
+.. tag:: mail
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/roundcube.svg
+      :align: center
+
+#################
+Roundcube Webmail
+#################
+
+.. tag_list::
+
+Roundcube webmail is a browser-based IMAP client with an easy-to-use user interface. It provides full functionality you expect from an email client, including MIME support, address book, folder manipulation, message searching and spell checking.
+
+The global Uberspace Webmail-Instance is also based on roundcube. Deploying your own roundcube webmail is ideal if you want to add functionality by plugins or customize the theme etc.
+
+----
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * :manual:`MySQL <database-mysql>`
+  * :manual:`PHP <lang-php>`
+  * :manual:`domains <web-domains>`
+  * :manual:`Mail <mail-access>`
+
+License
+=======
+
+Roundcube Webmail is released under the GNU General Public License_ version 3 or any later version with exceptions for skins and plugins.
+
+
+Prerequisites
+=============
+
+We're using :manual:`PHP <lang-php>` in the stable version 7.3:
+
+::
+
+ [isabell@stardust ~]$ uberspace tools version use php 7.3
+ Selected PHP version 7.3
+ The new configuration is adapted immediately. Patch updates will be applied automatically.
+ [isabell@stardust ~]$ uberspace tools version show php
+ Using 'PHP' version: '7.3'
+ [isabell@stardust ~]$
+
+If you want to use your Roundcube Webmail with your own domain you need to setup your domain first:
+
+::
+
+ [isabell@stardust ~]$ uberspace web domain list
+ isabell.uber.space
+ [isabell@stardust ~]$
+ 
+You'll need your MySQL :manual_anchor:`credentials <database-mysql.html#login-credentials>`. Get them with ``my_print_defaults``:
+
+::
+
+ [isabell@stardust ~]$ my_print_defaults client
+ --default-character-set=utf8mb4
+ --user=isabell
+ --password=MySuperSecretPassword
+ [isabell@stardust ~]$
+ 
+
+
+
+Installation
+============
+
+``cd`` to your :manual:`document root <web-documentroot>`, download the latest release of Roundcube and extract it. Move all containing files up one level and remove the emtpty directory as well as the archive afterwards.
+
+.. note:: Check the Roundcube_ website or `Github Repository`_ for the latest stable release and copy the download link to the Complete.tar.gz file. Then use ``wget`` to download it. Replace the URL with the one you just got from GitHub/Website.
+
+.. code-block:: console
+ :emphasize-lines: 2,4,5,6,7,8
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
+ [isabell@stardust html]$ wget https://github.com/roundcube/roundcubemail/releases/download/1.4.3/roundcubemail-1.4.3-complete.tar.gz
+ [isabell@stardust html]$ tar xfz *.tar.gz
+ [isabell@stardust html]$ cd roundcubemail-1.4.3
+ [isabell@stardust roundcubemail-1.4.3]$ mv * .[a-z]* ../ 
+ [isabell@stardust roundcubemail-1.4.3]$ cd ..
+ [isabell@stardust html]$ rm -r roundcubemail-1.4.3/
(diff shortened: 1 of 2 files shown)


source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+
+If you want to use your Roundcube Webmail with your own domain you need to setup your domain first:
+
+::
+
+ [isabell@stardust ~]$ uberspace web domain list
+ isabell.uber.space
+ [isabell@stardust ~]$
+ 
+You'll need your MySQL :manual_anchor:`credentials <database-mysql.html#login-credentials>`. Get them with ``my_print_defaults``:
```
luto: Please use the `.. include:: includes/my-print-defaults.rst` include instead.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+ The new configuration is adapted immediately. Patch updates will be applied automatically.
+ [isabell@stardust ~]$ uberspace tools version show php
+ Using 'PHP' version: '7.3'
+ [isabell@stardust ~]$
+
+If you want to use your Roundcube Webmail with your own domain you need to setup your domain first:
+
+::
+
+ [isabell@stardust ~]$ uberspace web domain list
```
luto: Please use the `.. include:: includes/web-domain-list.rst` include instead.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+
+#################
+Roundcube Webmail
+#################
+
+.. tag_list::
+
+Roundcube webmail is a browser-based IMAP client with an easy-to-use user interface. It provides full functionality you expect from an email client, including MIME support, address book, folder manipulation, message searching and spell checking.
+
+The global Uberspace Webmail-Instance is also based on roundcube. Deploying your own roundcube webmail is ideal if you want to add functionality by plugins or customize the theme etc.
```
luto: It's not! :grin: we're using rainloop for uberspace 7.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+The final configuration can easily be done in the browser.  Point your Browser to your installation URL ``https://isabell.uber.space/installer/``.
+
+All prerequisite checks should be *OK* - click *NEXT*.
+
+.. note:: You can safely ignore the *NOT AVAILABLE*-Warnings at the SQLite, SQL Server etc Databases, as we'll use the MySQL Database.
+
+Enter the following details at the Tab *CREATE CONFIG*:
+
+General configuration
+---------------------
```
luto: Please use `^^^^^^^^^^^^^` headings for this and the other configuration inputs. That way they are correctly grouped as part of "Roundcube Installer" instead of being at the same level.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+
+On the last tab *Test config* click on **initialize database** to initialize the database.
+
+.. note:: Optionally you can check your SMTP & IMAP config by entering valid credentials down here, but it's not necessary.
+
+Your done. Point your Browser to your installation URL ``https://isabell.uber.space`` and
+admire your shiny new webmail client!
+
+
+Best practices
```
luto: While I haven't actually tried the guide yet, it looks like this creates an open webmail client for anyone it use. Is that the case? If yes, please add a section here on how to restrict to it a domain or mail address. Otherwise anyone on the planet can use it and possibly exploit security issues, which are only relevant to logged-in users.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+
+
+
+Updates
+=======
+
+.. note:: Check the update feed_ regularly to stay informed about the newest version.
+
+If there is a new version available, **always make a backup first**.
+Refer to the Upgrade_ Manual of the official Roundcube Wiki for further instructions.
```
luto: This link is broken. You can create links with spaces in them like you did above:
```
`Roundcube Installer`_
```

luto: Please outline the easy `./bin/installto.sh` updating method here. It's fine to link to the more complicated manual way for more information.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+``cd`` to your :manual:`document root <web-documentroot>`, download the latest release of Roundcube and extract it. Move all containing files up one level and remove the emtpty directory as well as the archive afterwards.
+
+.. note:: Check the Roundcube_ website or `Github Repository`_ for the latest stable release and copy the download link to the Complete.tar.gz file. Then use ``wget`` to download it. Replace the URL with the one you just got from GitHub/Website.
+
+.. code-block:: console
+ :emphasize-lines: 2,4,5,6,7,8
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
+ [isabell@stardust html]$ wget https://github.com/roundcube/roundcubemail/releases/download/1.4.3/roundcubemail-1.4.3-complete.tar.gz
+ [isabell@stardust html]$ tar xfz *.tar.gz
```
luto: You can use `tar --strip-components=1` to get rid of the `cd; mv; cd; rm` below. That way all the contents of `roundcubemail-1.4.3` inside the tar will end up directly in the current directory.

source/guide_roundcube.rst
```diff
@@ -0,0 +1,222 @@
...
+.. tag:: web
+.. tag:: mail
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/roundcube.svg
+      :align: center
+
+#################
+Roundcube Webmail
```
luto: While roundcube itself isn't terribly consistent about that, the project seems to be named "Roundcube" not "Roundcube Webmail". Please adapt this, to keep the our guide list clean. Feel free to add `webmail` to the tag list to make it easier to discover, though!

source/guide_roundcube.rst
pxlfrk: well ... my fault. I removed the paragraph.

source/guide_roundcube.rst
pxlfrk: I changed the wording in the guide to `Roundcube`.

source/guide_roundcube.rst
pxlfrk: fixed the link.

source/guide_roundcube.rst
pxlfrk: changed the format as suggested.

source/guide_roundcube.rst
pxlfrk: I'm not sure if i get your point right. Partly yes, this guide is intended to help people create their own webmail client, **but it needs to be specific to one IMAP-Server**, in our example `stardust.uberspace.de`. The Authentication in both IMAP & SMTP is encrypted via SSL/TLS.
So the only accepted domains/mail adresses are the ones valid at the configured Mailserver. The user could also configure the webmail-client to adapt to a foreign Mailserver, yes.

There are ways to restrict the access to specific domains, but in my opinion that goes far beyond the aim of a installation guide, which this seems to be. In order to avoid unnecessarily length i linked to further instructions in the regarding sections instead.

source/guide_roundcube.rst
pxlfrk: i added the update-section to the guide.

source/guide_roundcube.rst
pxlfrk: I changed it to your suggestion, but couln't get rid of the final `rm -r *.tar.gz`. Initially i wanted to use curl, but i didn't get it to work properly. Thanks for any further suggestion to optimize.

luto: the last rm is fine :) thanks!

source/guide_roundcube.rst
luto: Ah, I see. The `default_host` doesn't just set a _default_, it restricts selection to the given server(s). Sorry for the noise then!

luto:
Thank you! :tada:

PR #444, opened 2019-07-12
Guide for ejabberd

fm0de:
Guide for ejabberd including Erlang installation. Could possibly split in two guides.

PR #596, opened 2020-03-29
Guide for ejabberd

jo-mei:
Hi,
since I just set this up myself, I took the time to rework the guide from #444


luto:
Thanks for your guide :)

nichtmax:
heads up: we'll provide libyaml with one of the upcoming Uberspace 7 updates. No need to compile it and no need to fiddle with `.bash_profile`

jo-mei:
ok, thanks for the heads up. I will update the guide once libyaml is rolled out.

source/guide_ejabberd.rst
```diff
@@ -0,0 +1,327 @@
...
+  * ``--prefix=$HOME/opt/ejabberd/``: Install to your personal uberspace
+  * ``--enable-user=$USER``: Allow execution of ejabberd as $USER
+  * ``--enable-mysql --enable-new-sql-schema``: optionally compile with mysql support
+  * ``--enable-sqlite``: optionally compile with sqlite support
+
+Run ``./configure --help`` to see all options.
+
+::
+
+ [isabell@stardust ~]$ cd $HOME/src/
```
luto, line 76: this directory doesn't exist.

source/guide_ejabberd.rst
```diff
@@ -0,0 +1,327 @@
...
+ [isabell@stardust ejabberd-20.03]$ ./autogen.sh
+ [isabell@stardust ejabberd-20.03]$ ./configure --enable-user=$USER --prefix=$HOME/opt/ejabberd --enable-mysql --enable-new-sql-schema
+ [isabell@stardust ejabberd-20.03]$ make install
+
+Make the control script available through ``.bash_profile``:
+
+.. code-block:: bash
+
+ # Add ejabberd binaries to PATH
+ export PATH=$HOME/opt/ejabberd/sbin:$PATH
```
luto, line 89: since there is just `ejabberdctl` in that directory, a `ln -s ~/opt/ejabberd/sbin/ejabberdctl ~/bin/` seems like an easier solution. What do you think?

source/guide_ejabberd.rst
jo-mei: fixed.

source/guide_ejabberd.rst
jo-mei: also fine with me.

Issue #342, opened 2019-03-21
Guide for Fathom


ebroda:
https://usefathom.com/ provides real-simple, privacy centered tracking of your website usage. "Fathom Analytics provides simple, useful websites stats without tracking or storing personal data of your users."

I'd just deployed it to an Uberspace and going to write a guide for the Uberlab.

sdoering:
Locking forward to this. I would love to try out fathom - did you (@ebroda) stumble upon bigger roadblocks? Or was it more or less straightforward? 

Thanks in advance.

youngpilot:
I can't get fathom to run on my uberspace. Fathom states ERROR: File not executable. 
Can somebody help?

PR #606, opened 2020-04-03
Add guide for Fathom Lite 

ebroda:
Finally, after over a year, add the proposed guide for Fathom Lite (previously named Fathom).
closes Uberspace/lab#342


luto:
Thanks for your guide :)

tobimori:
Worked flawlessy, thanks for the guide!

nichtmax:
Thank you! Works like a charm.

PR #607, opened 2020-04-03
Fix some more typos

ebroda:
After improving the dict for lab-specific words, these issues remained. With the better dict no only 10 spelling errors (German words in a tutorial, false negative) remain.
PR for spellchecking follows.


luto:
Thank youu! :tada:

PR #605, opened 2020-04-03
Fix typos in guides found by sphinxcontrib.spelling

ebroda:
Just out of curiosity I added a spellchecking-feature to the lab build. This found more than 1000 errors, which were mostly because of too specific terms not in the default dictionary. I fixed some of the obvious spelling mistakes in this PR.

The spellcheck integration is in this commit:
https://github.com/ebroda/lab/commit/03e8790eeb80bfb0df68f9f61012c2208cc2e1ee

Run with make spelling to create a report.

PR #608, opened 2020-04-03
Spell checking

ebroda:
As suggested by @luto in #605 here the PR for adding spell checking as additional command to make.

To run the spell check execute ``make spelling``. This returns a list of spelling errors of the guides. To reduce the amount of false-positives the dict.txt adds some Uberlab specific words.

The ``spelling-dict-tools.py`` provides two small helper functions to re-sort the dict and analyze the output of the spelling command regarding frequency of misspelled words and so the adding in the dict.txt.


nichtmax:
This is awesome! Thanks a lot!

nichtmax:
I'm sorry, I was too fast, we can't get it working. Sphinx has to be updated to version > 2

```
ERROR: sphinxcontrib-spelling 4.3.0 has requirement Sphinx>=2.0.0, but you’ll have sphinx 1.8.5 which is incompatible.
```

but even if we do that there are still errors:

```
(venv) ➜  lab git:(update_sphinx) make spelling
Running Sphinx v3.0.3
Extension error:
Could not import extension sphinxcontrib.spelling (exception: The 'enchant' C library was not found and maybe needs to be installed.
See  https://pyenchant.github.io/pyenchant/install.html
for details
)
make: *** [spelling] Error 2
(venv) ➜  lab git:(update_sphinx)
```

I don't think it's possible to install `enchant` in Netlify.. Any ideas?

ebroda:
Sorry, I wasn't aware of that.

I tried locally on WLS Ubuntu and even though the spelling lib says at least 2.0.0 it also works with 1.8.5. I also had teh problem with enchant which I could install via apt. Seems like in the other test I had it somehow pre-installed ...

A short google also didn't show an easy way to do it - but maybe someone else has an idea.

PR #610, opened 2020-04-04
[isso] add guide for isso

### LINKED ISSUE #569: Guide for "isso" commenting platform
I would like to write the guide for the "isso" commenting platform. 

It is powered by python and can be used to inject commenting functionalities to blogs or other static site generators. It works similar as commento.

ReyNow:
Hello, I finished the Guide for Isso.

Closes #569


luto:
Thanks for your guide :)

PR #611, opened 2020-04-04
Postgres guide: order install steps to be more intuitive

DoctypeRosenthal:
To avoid confusion in the install process I reordered the install steps just as we discussed in our email correspondence.


luto:
Thanks! :)

PR #613, opened 2020-04-06
Coturn

### LINKED ISSUE #581: Add guide for turn server (coturn)
It would be awesome to have a family video-chat with nextcloud talk (or via xmpp jitsi). Unfortunately p2p connections fail as often as not, so a turn server would be the only solution to make this family friendly. 
I think coturn requires libevent which I could not find on the manual page, so I guess it is not supported or has to be installed manually.

jo-mei:
fixes #581 

I could not find any logo for it.
And I have no setup for user/password authentication, just static-auth because it was sufficient for my use case.

### DIFF
--- a/source/guide_coturn.rst
+++ b/source/guide_coturn.rst
@@ -118,7 +118,7 @@ Create ``~/etc/services.d/coturn.ini`` with the following content:
 .. code-block:: ini
 
   [program:coturn]
-  command=%(ENV_HOME)s/opt/coturn/bin/turnserver -c %(ENV_HOME)s/etc/coturn/turnserver.conf
+  command=%(ENV_HOME)s/opt/turnserver/bin/turnserver -c %(ENV_HOME)s/etc/coturn/turnserver.conf
 
 .. include:: includes/supervisord.rst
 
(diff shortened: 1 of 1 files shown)


GerJuli:
Thanks a lot for this guide!
I tried the installation today but had some hassle with setting up the service. I think you got the wrong path in the coturn.ini

What you suggested is 
```
[program:coturn]
command=%(ENV_HOME)s/opt/coturn/bin/turnserver -c %(ENV_HOME)s/etc/coturn/turnserver.conf
```
which led me to the following
```
[moanos@reinmuth coturn]$ supervisorctl status
coturn                           FATAL     can't find command '/home/moanos/opt/coturn/bin/turnserver'
```
I think it should be 

```
[program:coturn]
command=%(ENV_HOME)s/opt/turnserver/bin/turnserver -c %(ENV_HOME)s/etc/turnserver/turnserver.conf
```

jo-mei:
you are right. should be fixed now.
Also there was a typo in the synapse setup.

thomas-kammerer:
Thanks for this great guide! Just found a typo in the configuration:
`pkey` instead of `key`
After changing this I was able to access the admin session via https :)

GerJuli:
I tried this as 
```
cert=/home/moanos/etc/certificates/moanos.uber.space.crt
pkey=/home/moanos/etc/certificates/moanos.uber.space.key
dh-file=/home/moanos/etc/coturn/dhparam-2066.pem
```
Is this what you mean?
I don't know if this is the expected result: https://moanos.uber.space/ 

@thomas-kammerer

thomas-kammerer:
Right, before that the https connection didn't work. If you create a user you're able to login to the web interface..
https://testtk.uber.space:49740/

jo-mei:
@thomas-kammerer should be corrected now.

@GerJuli I think you have to connect directly via your port to see the admin interface. If you connect via webbackend (as I suppose you did with the link above) nginx will handle https and the connection afterwards will be plain.

GerJuli:
Thanks a lot, that did the trick!

luto:
Thank you!

PR #614, opened 2020-04-08
Minor grammatical fix in guide_roundcube

keith24:
"Your" is possessive, "you're" is a contraction of "you are".  Should be "you're" here, as in "you are done".


luto:
Thanks!

PR #627, opened 2020-04-21
Gitea: Add hint for migrate before restart on update

ebroda:
I just updated my Gitea instance (from 1.10.0 to 1.11.4) and it was not running after the update. I was impatient and found out, that Gitea was applying some migrations without giving any notice. Only if you actively trigger gitea to do the migration it's showing some progress. I've added a hint on the Gitea manual and also updated the default installation to the current version.

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -55,18 +55,18 @@ make sure that the file can be executed.
 ::
 
   [isabell@stardust ~]$ mkdir ~/gitea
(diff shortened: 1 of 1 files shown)


ebroda:
As I have some different problem again, I took another look at the documentation and this does not seem to do what it promises by the name ... 

>How can I create users before starting Gitea
> Gitea provides a sub-command gitea migrate to initialize the database, after which you can use the admin CLI commands to add users like normal.
[Gitea Docs](https://docs.gitea.io/en-us/faq/)

I'll re-check and maybe re-open this PR if I'm sure about the behaviour and solution.

PR #634, opened 2020-04-23
Added directory where goaccess.conf lives

qbi:


luto:
saw #639 first, sorry :/ thanks, though!

PR #639, opened 2020-04-25
[goaccess] update location of the default configuration

dubbl:
The default configuration seems to have been moved to `etc/goaccess/...`.


luto:
Thanks!

PR #633, opened 2020-04-22
[cryptpad] Updated for CryptPad 3.16.0

pietsch:
Since CryptPad version 2, the structure of `config.js` changed substantially, so I updated this guide for the current release 3.16.0.

PR #643, opened 2020-04-29
Update cryptpad to 3.16.0

ebroda:
I just tried to install Cryptpad. As the guide was quite old, some things have changed, mostly regarding the configuration. 
Also I updated the order in the guide, reduced the git pull a little bit and updated the version.


ebroda:
Ah, I didn't saw #633 - I will have a look at the PR and propose an update for that PR if necessary.

PR #646, opened 2020-04-29
[cryptpad] Update for 3.16.0 update

ebroda:
Merge my changes of #643 with the changes in #633 
Restructuring the guide a little bit to be more like the other guides, use depth=1 to save a bit of space on checkout and some small language issues.


luto:
This now conflicts, because I merged #633 first, sorry :/ We usually go through these in chronological order, oldest first. If you build upon an older PR, please leave a note there saying that we need to look at yours first.

ebroda:
I thought that it would work without problems if you merge the changes as they are. However I created a new pr in #655

PR #652, opened 2020-05-02
[WKD] Add guide for Web Key Directory

### LINKED ISSUE #473: Guide for Web Key Directory
I'd like to share my notes how i set up a Web Key Directory on U6 (but I think there shouldn't be any differences for U7). I currently do not have time to translate this to ReST, but maybe someone else has.

https://spacekookie.de/blog/usable-gpg-with-wkd/
> [WKD] stands for "Web Key Directory" and is a standard for making a users public key available via their e-mail provider or server with the domain that corresponds to their e-mail address. There's several clients (such as Enigmail in Thunderbird) that will use this standard to automatically fetch a user's public key, when writing an e-mail to them.

See also https://wiki.gnupg.org/WKD .

# Create Directory Structure
```bash
mkdir -p ~/html/.well-known/openpgpkey/hu
touch ~/html/.well-known/openpgpkey/policy
```
~~Allow Directory Access:~~
DO NOT DO THIS:
~~echo "Options +Indexes" >> \~/html/.well-known/openpgpkey/hu/.htaccess~~
(can be a privacy issue, see Starkstromkonsuments comment below)

Instead use
```plaintext
## WEB KEY DIRECTORY ##
<IfModule mod_mime.c>
   ForceType application/octet-stream
   Header always set Access-Control-Allow-Origin "*"
</IfModule>
```
in `~/html/.well-known/openpgpkey/hu/.htaccess`.

# Calculate WKD-Hash of Mailuser
Take the prefix of your e-mail address (i.e. in `isabell@stardust.uberspace.de`, this would be `isabell`), hash it with SHA-1 and then encode the output with z-base-32.

You can use https://cryptii.com/pipes/z-base-32 for that (add a hash-block with SHA-1 before the z.base-32-block).
* E-Mail: isabell@startdust.uberspace.de
* Mailuser: isabell
* WKD-Hash: mmuhurigesr7z8hzf6sh5cmfsnmiiyyr

Alternatively you can invoke gpg with the option --with-wkd-hash on the machine where your keys reside:
```bash
gpg --with-wkd-hash --list-public-keys "isabell@stardust.uberspace.de"
[...]
      0123456789ABCDEF0123456789ABCDEF01234567
uid        [uneingeschränkt] Isabell <sabell@stardust.uberspace.de>
           mmuhurigesr7z8hzf6sh5cmfsnmiiyyr@stardust.uberspace.de
```

# Exporting and storing public GPG key

Export the **binary** Version of your public key into a file named as the WKD-hash calculated before:
```bash
gpg --export 0123456789ABCDEF0123456789ABCDEF01234567 > mmuhurigesr7z8hzf6sh5cmfsnmiiyyr
scp mmuhurigesr7z8hzf6sh5cmfsnmiiyyr isabell@stardust.uberspace.de:/home/isabell/html/.well-known/openpgpkey/hu
```

# Test
```bash
env GNUPGHOME=$(mktemp -d) gpg --locate-keys isabell@stardust.uberspace.de
```

Starkstromkonsument:
closes #473

### DIFF
--- a/source/guide_wkd.rst
+++ b/source/guide_wkd.rst
@@ -6,17 +6,15 @@
 .. tag:: mail
 .. tag:: self-hosting
 
-#######
-Web Key Directory (WKD) for GnuPG
-#######
+#######################
+Web Key Directory (WKD)
+#######################
 
 .. tag_list::
 
-[WKD] stands for "`Web Key Directory`_" and is a standard for making a users public key available via their e-mail provider or server with the domain that corresponds to their e-mail address. There's several clients (such as Enigmail in Thunderbird) that will use this standard to automatically fetch a user's public key, when writing an e-mail to them.
+[WKD] stands for "`Web Key Directory`_" and is a standard for making a users GnuPG / OpenPGP public key available via their e-mail provider or server with the domain that corresponds to their e-mail address. There's several clients (such as Enigmail in Thunderbird or OpenKeyChain on Android) that will use this standard to automatically fetch a user's public key, when writing an e-mail to them.
 
-Web Key Directories provide an easy way to discover public keys through HTTPS. They provide an important piece to the infrastructure to improve the user experience for exchanging secure emails and files.
-
-In contrast to the public keyservers a Web Key Directory does not publish mail addresses. And it is an authoritative pubkey source for its domain.
+Web Key Directories provide an easy way to discover public keys through HTTPS. They provide an important piece to the infrastructure to improve the user experience for exchanging secure emails and files. In contrast to the public keyservers a Web Key Directory does not publish mail addresses and it is an authoritative pubkey source for its domain.
 
 ----
 
@@ -28,7 +26,7 @@ In contrast to the public keyservers a Web Key Directory does not publish mail a
 Prerequisites
 =============
 
-You need the public keys for the e-mail identities you want to provide via WKD (e.g. isabell@stardust.uberspace.de).
+You need the public keys for the e-mail identities you want to provide via WKD (e.g. isabell@stardust.uberspace.de). This guide does not explain how to generate GnuPG / OpenPGP keys.
 
 The domain for the advanced method has to be set up at your DNS-Hoster and within your uberspace:
 
@@ -42,7 +40,7 @@ The domain for the advanced method has to be set up at your DNS-Hoster and withi
 Create Directory Structure
 ==========================
 
-advanced method
+Advanced method
 ---------------
 
 The "advanced" method seems to be queried in front of the "direct" method by clients. If the subdomain openpgpkey resolves to a valid IP but the host at this IP does not provide the WKD (e.g. because of a wildcard in the DNS-Records), the WKD-lookup may fail.
@@ -57,9 +55,9 @@ Create the directories and the empty files:
  [isabell@stardust isabell]$ touch openpgpkey.stardust.uberspace.de/.well-known/openpgpkey/stardust.uberspace.de/.htaccess
  [isabell@stardust ~]$
 
-Add the following lines to the .htaccess-file to set the correct headers:
+Add the following lines to the empty ``.htaccess`` file to set the correct headers:
 
-.. code-block:: apache2
+.. code-block:: apache
 
  ## WEB KEY DIRECTORY ##
  <IfModule mod_mime.c>
@@ -70,7 +68,7 @@ Add the following lines to the .htaccess-file to set the correct headers:
 .. warning:: Make sure that there is no automatic directory listing! It is not necessary and it reveals the number of emailadresses (and their hashes) in the WKD (this can be a privacy issue)!
 
 
-direct method
+Direct method
 -------------
 
 Create the directories and symlinks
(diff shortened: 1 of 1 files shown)


source/guide_wkd.rst
```diff
@@ -0,0 +1,197 @@
...
+Prerequisites
+=============
+
+You need the public keys for the e-mail identities you want to provide via WKD (e.g. isabell@stardust.uberspace.de). This guide does not explain how to generate GnuPG / OpenPGP keys.
+
+The domain for the advanced method has to be set up at your DNS-Hoster and within your uberspace:
+
+::
+
+ [isabell@stardust ~]$ uberspace web domain list
```
luto, line 35: Please use the snippet like so `.. include:: includes/web-domain-list.rst`.

source/guide_wkd.rst
```diff
@@ -0,0 +1,197 @@
...
+---------------
+
+The "advanced" method seems to be queried in front of the "direct" method by clients. If the subdomain openpgpkey resolves to a valid IP but the host at this IP does not provide the WKD (e.g. because of a wildcard in the DNS-Records), the WKD-lookup may fail.
+
+Create the directories and the empty files:
+
+::
+
+ [isabell@stardust ~]$ cd /var/www/virtual/isabell/
+ [isabell@stardust isabell]$ mkdir -p openpgpkey.stardust.uberspace.de/.well-known/openpgpkey/stardust.uberspace.de/hu/
```
luto, line 53: `$user.$host.uberspace.de` domains do not exist anymore. There is only `$user.uber.space`.

source/guide_wkd.rst
```diff
@@ -0,0 +1,197 @@
...
+Advanced method
+---------------
+
+The "advanced" method seems to be queried in front of the "direct" method by clients. If the subdomain openpgpkey resolves to a valid IP but the host at this IP does not provide the WKD (e.g. because of a wildcard in the DNS-Records), the WKD-lookup may fail.
+
+Create the directories and the empty files:
+
+::
+
+ [isabell@stardust ~]$ cd /var/www/virtual/isabell/
```
luto, line 52: please use `/var/www/virtual/$USER/` and similar variables in this snippet and the others to make the instructions copy-pasteable.

source/guide_wkd.rst
```diff
@@ -0,0 +1,197 @@
...
+ [someuser@somehost ~]$ scp mmuhurigesr7z8hzf6sh5cmfsnmiiyyr isabell@stardust.uberspace.de:hu/
+ [someuser@somehost ~]$
+
+Testing
+=======
+
+First of all, these two URLs should be available using your browser:
+
+ * Advanced method: https://openpgpkey.stardust.uberspace.de/.well-known/openpgpkey/stardust.uberspace.de/hu/mmuhurigesr7z8hzf6sh5cmfsnmiiyyr
+ * Direct method: https://stardust.uberspace.de/.well-known/openpgpkey/hu/mmuhurigesr7z8hzf6sh5cmfsnmiiyyr
```
luto, line 142: Users cannot host content on `$host.uberspace.de`, so I guess there is a typo in that URL?

source/guide_wkd.rst
```diff
@@ -0,0 +1,197 @@
...
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * e-mail encryption with GnuPG / OpenPGP 
+  * :manual:`domains <web-domains>`
+
+Prerequisites
+=============
+
+You need the public keys for the e-mail identities you want to provide via WKD (e.g. isabell@stardust.uberspace.de). This guide does not explain how to generate GnuPG / OpenPGP keys.
```
luto, line 29: If you'd `isabell@uber.space` would make for a nicer and shorter example. That's an optional chnage, though :sweat_smile:

source/guide_wkd.rst
Starkstromkonsument: I did not use the snippet because I wanted to point out that the subdomain openpgpkey.[...] has to exist. Do you want me to change it anyway?

source/guide_wkd.rst
Starkstromkonsument: I thought "stardust.uberspace.de" is the equivalent to "example.org" to use for the guides. I could use isabell@example.org and openpgpkey.example.org for the guide if you prefer it though.

source/guide_wkd.rst
Starkstromkonsument: Should I use example.org and openpgpkey.example.org?

source/guide_wkd.rst
luto: Judging by the official docs the `openpgpkey.` subdomains seem to be more of a workaround and are not recommended. Can we get by by using only `.well-known/`?

Starkstromkonsument: Hi luto, can you provide a link to the docs you refer to please? In case you refer to the stopgap-method described at https://wiki.gnupg.org/WKD, were are not doing that in the proposed guide. We are using our own server (or webspace ;-) and do not redirect to a third party.

IMHO the openpgpkey-subdomain (Advanced method) is queried first. Only if it does not resolve by DNS, the direct method is used as a fallback (as pointed out here: https://wiki.gnupg.org/WKDHosting#Publishing). This could be a problem for people who have Wildcard-DNS-Entries set up.

Another guide at https://keys.openpgp.org/about/usage#wkd-as-a-service only describes the advanced method and does not even mention the direct one.

I would like to leave it in the guide to be fully compliant.

I propose to rework my guide to use those examples:

- isabell@example.org
- https://openpgpkey.example.org/.well-known/openpgpkey/example.org/hu/
- https://example.org/.well-known/openpgpkey/hu/

Starkstromkonsument: @Marmo What do you think about this, as the one who initially opened the issue?

luto: I managed to miss-read the docs then, sorry! :/ the proposed trio of placeholders are fine. Thanks!

Starkstromkonsument: Thank you for your feedback. I will push a commit, changing the Domain to example.org. What about the snippet `includes/web-domain-list.rst`? Should I use it anyway or not?

luto: no snippet in that case. Maybe a hint that the entire guide isn't very useful, if you don't own a domain.

luto:
perfect :) thank you! :tada:

PR #653, opened 2020-05-03
Add shell command for disabling maintenance mode after update

ubuntudroid:
Sometimes an update leaves you stuck in maintenance mode.

I've added a line to the shell commands of the update section which disables maintenance mode. It should be safe to always call this, even if maintenance mode is already disabled.


luto:
Thanks!

Issue #641, opened 2020-04-28
[redis] redis is preinstalled


nichtmax:
remove the compilation and just use the preinstalled binaries. Milestone is probably U7.6.2

luto:
will be released in v7.7.0

luto:
https://github.com/Uberspace/lab/pull/727

PR #654, opened 2020-05-03
Use clang to build redis

jrast:
Seems like Redis >= 6.0.0 fail to build with the GCC version available on uberspace. Using clang works.

See https://github.com/antirez/redis/issues/6286

Note: This is up for discussion. I just tried to installing redis using this guide and it failed. Using clang worked. But maybe there is a better solution?


luto:
Thanks :) we'll pre-install redis in the near future #641, so the complication step will be removed entirely.

PR #657, opened 2020-05-07
[mosquitto] Add guide for mosquitto

Snapstromegon:
Exlipse Mosquitto is one of the most popular (MQTT) message brokers with support for MQTT 3.1, 3.1.1 and 5.

This guide provides informations on how to create a basic broker with password authentication and websocket support.


nichtmax:
Thanks for the guide! We'll look into installing the RPMs globally and then get back to you.

Snapstromegon:
If you're installing the libwebsocket RPM, it would make the install easier.
If you also install mosquitto I'll rewrite the guide to show how to start mosquitto as a user with the default config and how to provide a custom config.

Snapstromegon:
Closed by mistake

nichtmax:
You'll get libwebsocket too ☺️

abirke:
Hi everyone,

we've documented another way of installing Mosquitto with @transistortim (see https://github.com/abirke/lab/tree/guide/mosquitto). It uses MQTT over TLS instead of WebSockets and building from source instead of extracting from the RPM packages.

Maybe we can try to merge the guides. It might be an interesting alternative to some users?

Have a nice evening,
André

Snapstromegon:
@abirke You're correct, that my guide does not support TCP connections. This can be easaly fixed by requesting a port in the firewall and setting the config like you did.
(This would be even easier, if Uberspace would provide a "--tcp" option for web backends, so the user doesn't have to mind changing certificates)

I also took a look at you guide and have some thoughts on it:
- Installation: Be more specific to look for the latest "source" release
- I'd prefer to use prebuild binaries as this would avoid problems during build time (e.g. RAM usage) and reduces the load on the systems. Also in my opinion it's easier to follow and if uberspace at one point decides to natively support it, the switch will be easier as you're most likely already use the version uberspace would provide.

Have you checked, that Mosquitto reloads the cert and key files when they change?

Nevertheless thanks for your feedback!

abirke:
Hi @Snapstromegon,

thanks for your review. Reloading the certificates is a fair point. The Mosquitto broker doesn't do it, so we added a section on how to automate a regular restart of the service.

As for the installation process, building the from the latest source is probably a personal preference ;) It also might be solved by the Uberspace team soon.

Our main goal here was to add documentation on the TCP variant, especially for low-end clients like ESP8266. Feel free to copy this part over if you like.

Snapstromegon:
@abirke I'll wait for the uberspace team to resolve the blocker and then merge your work regarding the tcp connection into my guide.

Maybe someone from uberspace can give a say wether to prefer the source or the rpm version for the guide.

Snapstromegon:
Why is the Close and comment button so prominent to click 🤦‍♂️

Snapstromegon:
Uberspace just dropped full support for mosquitto. ❤ Thanks Guys!
It works just like expected. Will you release a guide how to create your own mosquitto config as part of the manual, or should I transform this PR (or a new one) into a guide for this?

luto:
We're not planning to add any config examples, no. So we'd be very happy to tap into your knowledge there ;) feel free to close this PR and open a new one on how to configure mosquitto, if you'd like.

Snapstromegon:
Closing this PR in favor of a future one for mosquitto configuration.

(This time for real)

PR #661, opened 2020-05-15
[sulu] add guide for sulu

### LINKED ISSUE #660: Guide for Sulu CMS
I'm setting up my new website and will write a guide for [sulu cms](https://sulu.io) during that.

ckrack:
Sulu is a CMS based on Symfony. It's very flexible, can host multiple webspaces in one installation and even run as a headless cms.

/closes #660


source/guide_sulu.rst
```diff
@@ -0,0 +1,209 @@
...
+.. note:: You can optionally configure Sulu to send emails in the same place.
+
+
+When you’re done with the configuration, populate the database with Sulu’s default data by following the build walkthrough.
+
+.. code-block:: console
+ :emphasize-lines: 31
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/sulucms
+ [isabell@stardust sulucms]$ ./bin/adminconsole sulu:build dev
```
luto, line 124: Since `APP_ENV` is `prod`, should we also use `prod` instead of `dev` here?

source/guide_sulu.rst
```diff
@@ -0,0 +1,209 @@
...
+ APP_ENV=prod
+ DATABASE_URL=mysql://isabell:MySuperSecretPassword@localhost/isabell_sulucms?serverVersion=mariadb-10.3.22
+
+.. note:: You can optionally configure Sulu to send emails in the same place.
+
+
+When you’re done with the configuration, populate the database with Sulu’s default data by following the build walkthrough.
+
+.. code-block:: console
+ :emphasize-lines: 31
```
luto, line 121: you're off by one line :grin:

source/guide_sulu.rst
luto: Doing that seems to skip user creation, so the additional commands `./bin/adminconsole sulu:security:role:create` and `./bin/adminconsole sulu:security:user:create` are needed.

luto: This also forces users to choose proper password instead of logging in using `admin:admin`, so it seems like the better way to do it. What do you think?

ckrack: Yeah, I agree. I know too many people that don't change their default passwords, too..

luto: This still says "dev". Is that on purpose?

ckrack: Nope! Fixed it

luto:
Thank you! :tada:

PR #662, opened 2020-05-16
[phantombot] add guide for phantombot

tobimori:
PhantomBot is an actively developed open source interactive Twitch bot written in Java 
that provides entertainment and moderation for your channel.


luto:
thank you! :tada:

PR #665, opened 2020-05-17
[Docusaurus] add guid for Docusaurus

thomas-kammerer:


luto:
Works great. Thanks! :)

PR #668, opened 2020-05-20
[gitea] add auto update script

ezzra:
I don't know if automated update scripts fit to the uberspace philosophy, but here you go :)


luto:
closing this in favor of #677, which is a de-facto rewrite. sorry :/

PR #670, opened 2020-05-24
Upgrade guide_gitea.rst

FrankZisko:
Some improvements:
* general: Using variables for an easy procedure update.
* domain: Easy choosing of an own domain and domain subfolder.
* security: The installation has not to be finished via web call. So there is no open platform after installation.
* security: Administrator will be created with random password from /dev/urandom. -> Printed to bash output and saved into a file.
* security: Registration via web or via OpenID is disabled.
* comfort: user has not to print, copy and paste MariaDB username and password.
* log: Some logging files will be written.
* uninstall: An uninstallation script will be written.
* update: Update steps are more specified.

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -1,4 +1,4 @@
-.. author:: luto <http://luto.at>
+.. author:: Frank ℤisko, https://frank.zisko.io
 
 .. tag:: lang-go
 .. tag:: audience-developers
@@ -11,15 +11,14 @@
   .. image:: _static/images/gitea.png
       :align: center
 
+
 #####
 Gitea
 #####
 
 .. tag_list::
 
-Gitea_ is a painless self-hosted Git service written in Go and distributed under the MIT License, designed to provide
-similar core functionality as GitLab or GitHub, but at a much smaller resource footprint. The project started as a fork
-of the then-popular Gogs_, but quickly became more active and bigger than the original.
+Gitea_ is a self-hosted Git service with a functionality similar to GitHub, GitLab and BitBucket. It's a fork of Gogs_ and uses the same MIT licence_. As most applications written in Go it's easy to install.
 
 ----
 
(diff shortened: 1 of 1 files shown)


luto:
I am very sorry, there quite a few big incompatibilities with your guide compared to the rest of the lab:

1. Guides are supposed to demonstrate you how to install  a tool interactively. This may not be the fastest way to get to your goal, but it ensures that you actually understand what you're doing. This guide is structured like a script, able to run from start to finish without intervention. I'm a dev myself (and I guess you are one) so I value that, but it's not the goal of the lab.
2. it replaces the entre default configuration instead of just changing the values that need changing. This makes the guide hard to use once the default config updates. Responsibly updating gitea to a new version would also require first diffing your config against the old default, making the changes again manually and then get a fresh config dump for the updated guide.
3. to make clear where files are going, guides usually install into `~/appname`, so `~/gittea` without a variable pointing to that place.
4. each console line is prefixed with `[isabell@stardust ~]$ ` to make sure the view in the lab matches what users see in their terminal. This may not be needed for more experienced people, but is a major help for inexperienced users, many of who are using the lab to get started.
5. Uninstallation is nice, but not necessary. All guides assume that they're installed into an empty uberspace. So, if you'd like to uninstall agian, delete the space.
6. the guide uses `cat >> file <<__EOF__` lines to write files which aren't used in the Lab at all to make the guides easier to read.

Most of this is outlined in our [STYLE.md](https://github.com/Uberspace/lab/blob/master/STYLE.md), which is linked in [CONTRIBUTING.md](https://github.com/Uberspace/lab/blob/master/CONTRIBUTING.md). The guides for [syncthing](https://lab.uberspace.de/guide_syncthing.html), [taskd](https://lab.uberspace.de/guide_taskd.html), [wallabag](https://lab.uberspace.de/guide_wallabag.html) are great examples.

If you'd like to rewrite the guide according to our style guide and follow the style of all the other guides more closely, I'd be more than happy to take another look. As is, we can't include it, sorry.

FrankZisko:
I see. Thank you for explaining what I had not considered. I missed the style guide. It's right there in BIG letters. Sry. 🙈🤦

PR #681, opened 2020-06-16
[nextcloud] remove HSTS info

ezzra:
HSTS header default has been changed by Uberspace and nextcloud will no
longer complain


luto:
:tada:

PR #686, opened 2020-06-18
Add SSL configuration to guide for icecast2

thisven:
To me icecast2 configuration is missing SSL option for secure connection to admin interface and streaming option via HTTPS. I'm wondering why this has not been considered although it's mentioned in official documentation at https://icecast.org/docs/icecast-2.4.1/config-file.html#path.


luto:
nice, thanks!

Issue #680, opened 2020-06-16
TLS certificates not needed for synapse


hanseartic:
There is no need to configure TLS certificates for synapse, as the web-backend is handling TLS and passing traffic unencrypted to synapse.

hanseartic:
fixed with https://github.com/Uberspace/lab/compare/d225f8fc331679fccd71ee6c1874f11f0611075c...2a30b8441aa5921ecea1fe1be0f74fa5384f5f26

PR #689, opened 2020-06-21
[synapse] remove TLS certificate section

hanseartic:
TLS certificates are not used in reverse-proxy setup, because nginx is already handling TLS. See also hint in the generated config file

 ```
listeners:
   # TLS-enabled listener: for when matrix traffic is sent directly to synapse.
   #
   # Disabled by default. To enable it, uncomment the following. (Note that you
   # will also need to give Synapse a TLS key and certificate: see the TLS section
   # below.)
   #
   #- port: 8448
   #  type: http
   #  tls: true
   #  resources:
   #    - names: [client, federation]
```

which states that TLS needs to be configured *if* this listener were enabled

### DIFF
--- a/source/guide_synapse.rst
+++ b/source/guide_synapse.rst
@@ -179,16 +179,6 @@ This has to be made available under ``/.well-known/matrix`` via the web backend:
   Set backend for /.well-known/matrix to apache.
   [isabell@stardust ~]$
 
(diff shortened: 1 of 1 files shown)


hanseartic:
addresses #680

luto:
Thanks!

PR #690, opened 2020-06-21
[offen] add initial guide for installing Offen web analytics

m90:
This adds a guide for installing [Offen](https://github.com/offen/offen) on an Uberspace (I just created one today and ran the following successfully).

~~One problem I ran into here: I tried using the default MariaDB installation first (as it should be compatible with MySQL), but the Golang SQL driver exploded in ways I could not understand when trying to do this. We have been running Offen against "plain" MySQL instances successfully already so it must be some subtle difference that breaks it. If you have any experience on why this happens, I'd be happy to change this guide to use MariaDB instead (as it makes a lot of sense to use it when you have the DB up and running in any case). This is the issue I am seeing: https://github.com/go-gorm/gorm/issues/3001~~

---

Disclaimer: I'm an author and maintainer of this project.


m90:
I figured out how to use MariaDB correctly, so this guide is now using the local DB instead of SQLite.

luto:
@m90 just FYI, no need to change this. To save some bytes and make the source less verbose can use

```
``-quiet``
```

instead of

```
:code:`-quiet`
```

source/guide_offen.rst
```diff
@@ -0,0 +1,326 @@
...
+
+Database setup
+------------
+
+Offen will store its data in the MariaDB provided by your Uberspace. First,
+create a dedicated database for Offen to use:
+
+::
+
+ [isabell@stardust ~]$ mysql -e "CREATE DATABASE isabell_offen"
```
luto, line 189: Please use `$USER`:

```suggestion
 [isabell@stardust ~]$ mysql -e "CREATE DATABASE ${USER}_offen"
```

source/guide_offen.rst
```diff
@@ -0,0 +1,326 @@
...
+
+Point the web backend to Offen
+------------
+
+As a last step you need to point your web backend to your Offen instance that
+is now running on the default port 3000:
+
+::
+
+ [isabell@stardust ~]$ uberspace web backend set / --http --port 3000
```
luto, line 250: Please use the provided snippet like so:

```
.. note::

    offen is running on port 3000.

.. include:: includes/web-backend.rst
```

source/guide_offen.rst
```diff
@@ -0,0 +1,326 @@
...
+.. tag:: analytics
+.. tag:: privacy
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/offen.png
+      :align: center
+
+##########
+Offen
```
luto, line 15: Please make sure the length of the `####` likes matches the length of `Offen` to get rid of build warnings.

source/guide_offen.rst
```diff
@@ -0,0 +1,326 @@
...
+
+::
+
+ [isabell@stardust ~]$ cp ~/tmp/offen-download/offen-linux-amd64 ~/bin/offen
+ [isabell@stardust ~]$ which offen
+ ~/bin/offen
+ [isabell@stardust ~]$
+
+Cleaning up
+-----
```
luto, line 133: Please make sure the length of the `----` likes matches the length of `Cleaning up` to get rid of build warnings. Same for all following headings.

m90:
@luto Thanks for your review. I fixed the build warnings and addressed the other comments too. Would you mind having another look?

PR #691, opened 2020-06-22
[prosody] add guide for prosody

taddydevil:
Prosody is a modern XMPP communication server. It aims to be easy to set up and configure, and efficient with system resources. Additionally, for developers it aims to be easy to extend and give a flexible system on which to rapidly develop added functionality, or prototype new protocols.
I tried to make a more easier guide in the hope to bring prosody into the lab requested in #107 

I used some parts 1:1 of the rejected guide by fapsi. I dont know how to handle it with the credits (maybe mark as co author?)

### DIFF
--- a/source/guide_prosody.rst
+++ b/source/guide_prosody.rst
@@ -162,24 +162,25 @@ Create a database for prosody:
 Installation
 ============
 
-.. note:: Check out the latest stable release on https://prosody.im/downloads/source/.
+.. note:: Check out the latest stable release on https://prosody.im/downloads/source/. We'll set a temporary environment variable for this session to handle it more easier with the version number in the files and directories.
 
-Download the latest stable release from source to ``~/prosody/prosody-0.11.5``:
+Download and extract the latest stable release from source into ``~/prosody/``:
 
 ::
 
- [isabell@stardust ~]$ mkdir ~/home/prosody
- [isabell@stardust ~]$ cd ~/home/prosody
- [isabell@stardust ~]$ wget https://prosody.im/downloads/source/prosody-0.11.5.tar.gz
+ [isabell@stardust ~]$ VERSION=0.11.5
+ [isabell@stardust ~]$ wget https://prosody.im/downloads/source/prosody-$VERSION.tar.gz --directory-prefix=$HOME/prosody/
  [...]
- [isabell@stardust ~]$ tar -xzf prosody-0.11.5.tar.gz
+ [isabell@stardust ~]$ tar --extract --gzip --file=$HOME/prosody/prosody-$VERSION.tar.gz --directory=$HOME/prosody/
+ [isabell@stardust ~]$ rm ~/prosody/prosody-$VERSION.tar.gz
+ [isabell@stardust ~]$
 
 Configure, build and install prosody:
 
 ::
 
- [isabell@stardust ~]$ cd ~/prosody/prosody-0.11.5
- [isabell@stardust ~]$ ./configure --ostype=linux --prefix=$HOME --libdir=$HOME/prosody --datadir=$HOME/prosody --with-lua-include=/usr/include
+ [isabell@stardust ~]$ cd ~/prosody/prosody-$VERSION
+ [isabell@stardust prosody-0.11.5]$ ./configure --ostype=linux --prefix=$HOME --libdir=$HOME/prosody --datadir=$HOME/prosody --with-lua-include=/usr/include
  Lua version detected: 5.1
  Lua interpreter found: /usr/bin/lua...
  Checking Lua includes... lua.h found in /usr/include/lua.h
@@ -192,11 +193,11 @@ Configure, build and install prosody:
 
  Done. You can now run 'make' to build.
 
- [isabell@stardust ~]$ make
+ [isabell@stardust prosody-0.11.5]$ make
  [...] 
- [isabell@stardust ~]$ make install
+ [isabell@stardust prosody-0.11.5]$ make install
  [...] 
- [isabell@stardust ~]$ 
+ [isabell@stardust prosody-0.11.5]$ 
 
 Configuration
 =============
@@ -267,9 +268,9 @@ Additionally I recommend the ssl ciphers and options to reach a high security sc
  s2s_timeout = 300
  ssl = {
    dhparam = "/home/isabell/etc/prosody/certs/dhparam-4096.pem";         
-	 cafile = "/etc/pki/tls/certs/ca-bundle.trust.crt";
-	 ciphers = "EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA+SHA384:EECDH+ECDSA+SHA256:EECDH+aRSA+SHA384:EECDH+aRSA+SHA256:EECDH:EDH+aRSA:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS:!RC4:!SEED:!AES128:!CAMELLIA128";
-	 options = { "no_sslv2", "no_sslv3", "no_tlsv1"; "no_ticket", "no_compression", "cipher_server_preference", "single_dh_use", "single_ecdh_use" };
+   cafile = "/etc/pki/tls/certs/ca-bundle.trust.crt";
+   ciphers = "EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA+SHA384:EECDH+ECDSA+SHA256:EECDH+aRSA+SHA384:EECDH+aRSA+SHA256:EECDH:EDH+aRSA:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS:!RC4:!SEED:!AES128:!CAMELLIA128";
+   options = { "no_sslv2", "no_sslv3", "no_tlsv1"; "no_ticket", "no_compression", "cipher_server_preference", "single_dh_use", "single_ecdh_use" };
  }
  authentication = "internal_hashed"
  pidfile = "/home/isabell/prosody/prosody.pid";
@@ -324,9 +325,9 @@ Finishing installation
 Create your first user:
 
 .. code-block:: console
- :emphasize-lines: 2,3 
+ :emphasize-lines: 1-3 
 
- [isabell@stardust ~]$ prosodyctl adduser name@isabell.uber.space
+ [isabell@stardust ~]$ prosodyctl adduser isabell@isabell.uber.space
  Enter new password:
  Retype new password:
  [isabell@stardust ~]$
@@ -340,12 +341,13 @@ For updates simply repeat the steps described in the Installation_ part:
 
 ::
 
- [isabell@stardust ~]$ cd ~/prosody
- [isabell@stardust ~]$ wget https://prosody.im/downloads/source/prosody-0.XX.X.tar.gz
+ [isabell@stardust ~]$ VERSION=X.XX.X
+ [isabell@stardust ~]$ wget https://prosody.im/downloads/source/prosody-$VERSION.tar.gz --directory-prefix=$HOME/prosody/
  [...]
- [isabell@stardust ~]$ tar -xzf prosody-0.XX.X.tar.gz
- [isabell@stardust ~]$ cd ~/prosody/prosody-0.XX.X
- [isabell@stardust ~]$ ./configure --ostype=linux --prefix=$HOME --with-lua-include=/usr/include
+ [isabell@stardust ~]$ tar --extract --gzip --file=$HOME/prosody/prosody-$VERSION.tar.gz --directory=$HOME/prosody/
+ [isabell@stardust ~]$ rm ~/prosody/prosody-$VERSION.tar.gz
+ [isabell@stardust ~]$ cd ~/prosody/prosody-$VERSION
+ [isabell@stardust prosody-X.XX.X]$ ./configure --ostype=linux --prefix=$HOME --libdir=$HOME/prosody --datadir=$HOME/prosody --with-lua-include=/usr/include
  Lua version detected: 5.1
  Lua interpreter found: /usr/bin/lua...
  Checking Lua includes... lua.h found in /usr/include/lua.h
@@ -358,12 +360,12 @@ For updates simply repeat the steps described in the Installation_ part:
 
  Done. You can now run 'make' to build.
 
- [isabell@stardust ~]$ make
+ [isabell@stardust prosody-X.XX.X]$ make
  [...] 
- [isabell@stardust ~]$ make install
+ [isabell@stardust prosody-X.XX.X]$ make install
  [...] 
- [isabell@stardust ~]$ supervisorctl restart prosody
- [isabell@stardust ~]$ 
+ [isabell@stardust prosody-X.XX.X]$ supervisorctl restart prosody
+ [isabell@stardust prosody-X.XX.X]$ 
 
 XMPP Clients
 ============
--- a/source/guide_prosody.rst
+++ b/source/guide_prosody.rst
@@ -371,7 +371,7 @@ XMPP Clients
 As a personal note I want to recommend following XMPP clients:
 
 * Gajim_ for Linux / Windows.
-* Converations_ for Android.
+* Conversations_ for Android.
 * ChatSecure_ for iOS.
 
 ----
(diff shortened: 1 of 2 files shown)


source/guide_prosody.rst
```diff
@@ -0,0 +1,393 @@
...
+ Using Lua from: /usr
+
+ Done. You can now run 'make' to build.
+
+ [isabell@stardust ~]$ make
+ [...] 
+ [isabell@stardust ~]$ make install
+ [...] 
+ [isabell@stardust ~]$ supervisorctl restart prosody
+ [isabell@stardust ~]$
```
EV21, line 366: Let's make this section Copy&Paste-friendly (in combination with the other suggestion).
```suggestion
 [isabell@stardust ~]$ VERSION=X.XX.X
 [isabell@stardust ~]$ cd ~/prosody
 [isabell@stardust prosody]$ wget https://prosody.im/downloads/source/prosody-$VERSION.tar.gz
 [...]
 [isabell@stardust prosody]$ tar -xzf prosody-$VERSION.tar.gz
 [isabell@stardust prosody]$ cd ~/prosody/prosody-$VERSION
 [isabell@stardust prosody-X.XX.X]$ ./configure --ostype=linux --prefix=$HOME --with-lua-include=/usr/include
 Lua version detected: 5.1
 Lua interpreter found: /usr/bin/lua...
 Checking Lua includes... lua.h found in /usr/include/lua.h
 Checking if Lua header version matches that of the interpreter... yes
 Writing configuration...

 Installation prefix: /home/isabell
 Prosody configuration directory: /home/isabell/etc/prosody
 Using Lua from: /usr

 Done. You can now run 'make' to build.

 [isabell@stardust prosody-X.XX.X]$ make
 [...] 
 [isabell@stardust prosody-X.XX.X]$ make install
 [...] 
 [isabell@stardust prosody-X.XX.X]$ supervisorctl restart prosody
 [isabell@stardust prosody-X.XX.X]$ 
```

taddydevil:
> These are just suggestions, deny them if you don't like them. 😉
> Always test 🧪 them before you apply them.

Hey, thank you for the review. I like the idea of the VERSION variable and I change the steps later. The only one I dont like is the part with the script. I comment it seperatly.

source/guide_prosody.rst
```diff
@@ -0,0 +1,393 @@
...
+ Prosody configuration directory: /home/isabell/etc/prosody
+ Using Lua from: /usr
+
+ Done. You can now run 'make' to build.
+
+ [isabell@stardust ~]$ make
+ [...] 
+ [isabell@stardust ~]$ make install
+ [...] 
+ [isabell@stardust ~]$
```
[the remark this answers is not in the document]
taddydevil: Its a nice aspect and I didnt now that you can run configure and make with the directory declaration. And thats maybe the problem. I think to configure, make and install the package IN the directory is the most known way to do this and for new users its a little bit confuse maybe. Other guides "make" it in the directory too ;) I want to keep this pattern.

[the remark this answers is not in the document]
EV21: That is why I actually just mentioned it and didn't put it into the code-suggestion. It's just a personal taste of doing stuff because I also write shell scripts where `cd` is a very confusing element.
It is just an idea so people could also put the elements of your guide into a script, hit `enter` - and it's done :wink:
I know most guides do it the "cd-way", so another way could also be confusing. So just keep it simple. I agree.

EV21:
I was wandering why the installation is creating a prosody folder inside the prosody folder. 
It looks like the makefile is adding another prosody to the path of the LIBDIR.

It's not a big deal, but I don't think it's nice.

Suggestion 1
Without setting the parameter `--libdir=$HOME/prosody` the default setting would be:
LIBDIR="$PREFIX/lib" + makefile `/prosody` -> `/home/isabell/lib/prosody/`.
So it's not in `~/prosody`

Suggestion 2
Or if you set `--libdir=$HOME` you will get this:
```
~]$ tree -d prosody/
~/prosody/
├── core
├── modules
...
├── net
...
├── prosody-0.11.5
...
```
Suggestion 3: `--libdir=$HOME/lib` -> `~/prosody/lib/prosody/` :question:  :thinking:
Suggestion 4: Don't change it, it's not a beauty contest :laughing:

taddydevil:
Without any changes you have 4 directories afaik:

```
~/
├── etc/prosody (configuration path)
├── lib/prosody (later installed files)
├── var/lib/prosody (directory for http_upload and community modules)
├── prosody/prosody-0.11.5 (source files)
```

The reason why I set the directories to the one in $HOME is to have all prosody files overall in one directory.

To suggestion 2: 
I know that prosody is duplicated because I dont want to install prosody in the directory where http_upload, community modules and the pidfile is located like:

```
~/prosody/
├── http_upload
├── community-modules
├── core
├── modules
...
```

To suggestion 3:
The path is identical like in suggestion 1 in `~/lib/prosody` (not `~/prosody/lib/prosody`)

In my opinion I dont like the `$HOME/var/lib/prosody` and `$HOME/lib/prosody` directories. I dont know, probably because I not use this directories on my uberspace and everytime I had no idea where the http_upload is located (`~/lib/` or `~/var/lib` 😄 )

I changed it to unterstand the paths better but I know its not beautiful. If there are any default rules for something like this please tell me. Personally I use this one directory with the duplicated prosody and that makes it easier work with that for me.

EV21:
> The path is identical like in suggestion 1

My brain was already in hibernation mode. :sleeping: 
The parameter would be `--libdir=$HOME/prosody/lib`.

> The reason why I set the directories to the one in $HOME is to have all prosody files overall in one directory.
...
Personally I use this one directory with the duplicated prosody and that makes it easier work with that for me.

Makes sense. I was thinking about the single-use-concept of uberspaces, so actually isabells meaning of life would be to run prosody. :rocket: If you are running multiple different services with your uberspace it could be confusing with tons of directories, so in that case you would wish for a central directory/namespace for each app. 
People would create an uberspace e.g. isabell-prosody to run prosody, so I would prefer the default structure plus prefix (home)

> ~/
> ├── etc/prosody (configuration path)
> ├── lib/prosody (later installed files)
> ├── var/lib/prosody (directory for http_upload and community modules)
> ├── prosody/prosody-0.11.5 (source files)

If I need to access some files very often I would also create a symbolic link like there is for html.

taddydevil:
So, now I have the following structure:

```
[taddydev@brewington ~]$ tree -d
├── etc
│   └── prosody //Config directory
│       └── certs
├── lib //Source directory
│   └── prosody
│       └── core
│       └── modules
│       └── [...]
├── var //Data directory
│   └── lib
│       └── prosody
│           └── http_upload
│               └── [...]
│           └── prosody-0.11.5
│               └── [...]
│           └── prosody-modules
│               └── [...]
```

And I think, this is nice. The directory  `~/home/prosody` is no longer needed.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+
+Additionally I recommend the ssl ciphers and options to reach a high security score. You can check it over the `IM Observatory`_.
+
+.. code-block:: lua
+ :emphasize-lines: 2,3, 26, 29, 32, 38, 43-45, 50, 51, 53, 56, 57, 59
+
+ ---------- Server-wide settings ----------
+ admins = { "isabell@uber.space" }
+ plugin_paths = { "/home/isabell/var/lib/prosody/prosody-modules" }
+ modules_enabled = {
```
luto, line 242: It seems to me like this just adds `mam` and `csi_simple` to the default list. What do you think about changing the instructions to just add/uncomment these two instead of replacing the whole list? That way the guide would be more compatible on updates and we can keep the comments in the config file.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+   "version";
+   "uptime";
+   "time";
+   "ping";
+   "register";
+   "mam";
+   "csi_simple";
+   "admin_adhoc";
+ allow_registration = false
+ c2s_require_encryption = true
```
luto, line 263: These are already in the default config, no need to set them again. Please remove them to keep the guide shorter.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+ c2s_ports = { C2S-PORT }
+ s2s_require_encryption = true
```
luto, line 265: This is already in the default config, no need to set them again. Please remove it to keep the guide shorter.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+ s2s_secure_auth = true
+ s2s_ports = { S2S-PORT }
+ s2s_timeout = 300
+ ssl = {
+   dhparam = "/home/isabell/etc/prosody/certs/dhparam-4096.pem";         
+   cafile = "/etc/pki/tls/certs/ca-bundle.trust.crt";
+   ciphers = "EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA+SHA384:EECDH+ECDSA+SHA256:EECDH+aRSA+SHA384:EECDH+aRSA+SHA256:EECDH:EDH+aRSA:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS:!RC4:!SEED:!AES128:!CAMELLIA128";
+   options = { "no_sslv2", "no_sslv3", "no_tlsv1"; "no_ticket", "no_compression", "cipher_server_preference", "single_dh_use", "single_ecdh_use" };
+ }
+ authentication = "internal_hashed"
```
luto, line 275: This is already in the default config, no need to set them again. Please remove it to keep the guide shorter.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+ daemonize= false;
+ storage = "sql"
+ sql = { 
+   driver = "MySQL", 
+   database = "isabell_prosody", 
+   username = "isabell", 
+   password = "MySuperSecretPassword", 
+   host = "localhost" 
+ }
+ archive_expires_after = "1w"
```
luto, line 286: This is already in the default config, no need to set them again. Please remove it to keep the guide shorter.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+ s2s_timeout = 300
+ ssl = {
+   dhparam = "/home/isabell/etc/prosody/certs/dhparam-4096.pem";         
+   cafile = "/etc/pki/tls/certs/ca-bundle.trust.crt";
+   ciphers = "EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA+SHA384:EECDH+ECDSA+SHA256:EECDH+aRSA+SHA384:EECDH+aRSA+SHA256:EECDH:EDH+aRSA:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS:!RC4:!SEED:!AES128:!CAMELLIA128";
+   options = { "no_sslv2", "no_sslv3", "no_tlsv1"; "no_ticket", "no_compression", "cipher_server_preference", "single_dh_use", "single_ecdh_use" };
+ }
+ authentication = "internal_hashed"
+ pidfile = "/home/isabell/var/lib/prosody/prosody.pid";
+ daemonize= false;
```
luto, line 277:
```suggestion
 daemonize = false;
```

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+ [...] 
+ [isabell@stardust prosody-0.11.5]$ 
+
+Configuration
+=============
+
+Generate SSL dhparam file
+-------------------------
+
+.. note:: This is going to take a long time!
```
luto, line 208:
```suggestion
.. note:: This is going to take a long time! You can start configuring the server while it runs, but don't start it yet.
```

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+   host = "localhost" 
+ }
+ archive_expires_after = "1w"
+ log = { info = "*console" }
+ certificates = "/home/isabell/etc/certificates/"
+ https_certificate = "/home/isabell/etc/certificates/upload.isabell.uber.space.crt"
+ http_ports = { }
+ https_ports = { FILEUPLOAD-PORT }
+
+ ----------- Virtual hosts -----------
```
luto, line 293: Does `VirtualHust "localhost"` stay in the config?

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+
+ * ``xmpp.isabell.uber.space``
+ * ``conference.isabell.uber.space``
+ * ``upload.isabell.uber.space``
+
+Your URL ``xmpp.isabell.uber.space`` and ``upload.isabell.uber.space`` needs to be setup:
+
+::
+
+ [isabell@stardust ~]$ uberspace web domain list
```
luto, line 57: This doesn't ensure that there are actually certificates for the domains, as we fetch them only when HTTPS clients want them. Trigger the generation for each one like so: `curl https://isabell.uber.space`.

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+Install runtime lua-dependencies
+--------------------------------
+
+The following dependencies_ (``luasocket``, ``luaexpat``, ``luafilesystem`` and ``luasec``) are required:
+
+::
+
+ [isabell@stardust ~]$ luarocks install luasocket --local
+ luasocket [...] is now built and installed in [...]
+ [isabell@stardust ~]$ luarocks install luaexpat --local
```
luto, line 125: The server outputs a warning:

> The version of LuaExpat on your system does not support stanza size limits, which may leave servers on untrusted networks (e.g. the internet) vulnerable to denial-of-service attacks. You should upgrade to LuaExpat 1.3.0 or higher as soon as possible. See https://prosody.im/doc/depends#luaexpat for more information.

Is there a way to install the correct version here?

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+
+Web domain
+----------
+
+.. note:: To understand and setup the domains easier I use the following (recommended) subdomains:
+
+ * ``xmpp.isabell.uber.space``
+ * ``conference.isabell.uber.space``
+ * ``upload.isabell.uber.space``
+
```
luto, line 52:
```suggestion

  Keep in mind that since you can't create DNS records for ``.uber.space`` domains, you'll need your own domain like ``isabell.org``.

```

source/guide_prosody.rst
```diff
@@ -0,0 +1,404 @@
...
+ PATH=$HOME/.luarocks/bin:$PATH
+ 
+ export PATH
+
+Additionally we need to provide the paths ``LUA_PATH`` as well as ``LUA_CPATH`` and have to reload to use ``luarocks`` accordingly:
+
+::
+
+ [isabell@stardust ~]$ echo "$(luarocks path)" >> ~/.bash_profile
```
luto, line 112: How about something like this, in case the paths change?
```suggestion
 [isabell@stardust ~]$ echo 'eval "$(luarocks path)"' >> ~/.bash_profile
```

source/guide_prosody.rst
taddydevil: The installed version of LuaExpat is in the version 1.3.3-1. You can check it with ```luarocks list```. I have no idea why this is showing up. I'll investigate to solve this issue.

source/guide_prosody.rst
taddydevil: What do you think when I continue the whole guide with ``isabell.org`` instead of ``isabell.uber.space`` with the following note:

```suggestion
.. note:: Keep in mind that since you can't create DNS records for ``.uber.space`` domains, you'll need your own domain like ``isabell.org``. To understand and setup the domains easier I use the following (recommended) subdomains:

 * ``xmpp.isabell.org``
 * ``conference.isabell.org``
 * ``upload.isabell.uber.space``

Your URL ``xmpp.isabell.org`` and ``upload.isabell.org`` needs to be setup:

::

 [isabell@stardust ~]$ uberspace web domain list
 isabell.org
 xmpp.isabell.org
 upload.isabell.org
 [isabell@stardust ~]$
```

luto: since this cannot work with the default domain at all, sure!

source/guide_prosody.rst
taddydevil: Its a full copy-paste-able config that must adapt on the highlighted lines. I dont know, when I remove all the defaults lines its unknown where the new lines should be added. 
When I split it between ```Server-wide settings``` and ```Virtual hosts``` the guide is going to be longer.
I mentioned the [Example configuratoin file](https://prosody.im/doc/example_config) with all comments. Maybe we can make a backup of the original config and make a new file with all contents I'm written down?

It's quite difficult to find a solution 😅

luto: It's quite common to have a "adapt these settings, add the missing ones" section in lab guides. I think we'll be fine with a code block that contains all the things that need changing.

source/guide_prosody.rst
taddydevil: No, its replaced by your domain.

source/guide_prosody.rst
taddydevil: Seems to be a known issue -> https://issues.prosody.im/1375. I've succesfully tested the version 1.3.0-1 and will adapt the guide.

luto:
thank you! :tada:

PR #692, opened 2020-06-27
Extend ejabberd guide

coderkun:
The guide for “ejabberd” is a good starting point but I find it incomplete for a production instance. This PR adds configuration options for MySQL, the [compliance test](https://compliance.conversations.im/) and several security options and adds some maintenance notes. Additionally it improves the structure a little bit.

### DIFF
--- a/source/guide_ejabberd.rst
+++ b/source/guide_ejabberd.rst
@@ -1,5 +1,6 @@
 .. author:: fm0de
 .. author:: jo-mei
+.. author:: coderkun
 
 .. tag:: XMPP
 .. tag:: Jabber
@@ -29,33 +30,10 @@ ejabberd
   * :manual:`domains <web-domains>`
   * :manual:`firewall ports <basics-ports>`
   * :manual:`web backends <web-backends>`
+  * :manual:`web headers <web-headers>`
+  * :manual:`MySQL <database-mysql>`
 
 This guide is based on the initial `pull request <https://github.com/Uberspace/lab/pull/444>`_ from `fm0de <https://github.com/fm0de>`_ and the `work from clerie <https://blog.clerie.de/ejabberd-auf-uberspace-installieren/>`_.
-
-Prerequisites
-=============
-
-Domain
-------
-
-Your ejabberd domain ``isabell.example`` needs to be setup. ejabberd then defaults to the subomains
-``upload.isabell.example``, ``conference.isabell.example`` , ``pubsub.isabell.example`` and ``proxy.isabell.example``
-::
-
- [isabell@stardust ~]$ uberspace web domain add isabell.example
- [isabell@stardust ~]$ uberspace web domain add conference.isabell.example
- [isabell@stardust ~]$ uberspace web domain add upload.isabell.example
- [isabell@stardust ~]$ uberspace web domain add pubsub.isabell.example
- [isabell@stardust ~]$ uberspace web domain add proxy.isabell.example
-
-Also we need the SSL Certificates. Visit the domains in a browser or run:
-::
-
- [isabell@stardust ~]$ curl https://isabell.example
- [isabell@stardust ~]$ curl https://conference.isabell.example
- [isabell@stardust ~]$ curl https://upload.isabell.example
- [isabell@stardust ~]$ curl https://pubsub.isabell.example
- [isabell@stardust ~]$ curl https://proxy.isabell.example
 
 
 Installation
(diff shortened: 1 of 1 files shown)


coderkun:
There is no include for web-headers yet, is there?

luto:
No, there isn't. Web headers are quite universal so I am not quite sure what that include would contain :sweat_smile:

source/guide_ejabberd.rst
```diff
@@ -132,81 +129,170 @@ Provide location of your keys and certificates for TLS transport:
...
+to match the domain and your user folder:
 
 .. code-block:: ini
- :emphasize-lines: 2-4
+  :emphasize-lines: 4,10
+
+  modules:
+    …
+    mod_http_upload:
+      put_url: "https://xmpp.@HOST@/upload"
```
luto, line 257: Should this be `xmpp.` or `upload.`? https://github.com/Uberspace/lab/commit/b0158b24cb07f2f06df09f795a9d22c929ce5d9f

coderkun: I like to use `xmpp.` because—in this guide—it is used for all HTTP-based features like BOSH, web sockets, file uploads etc. and not only uploads.

source/guide_ejabberd.rst
```diff
@@ -29,100 +30,96 @@ ejabberd
...
 
- [isabell@stardust ~]$ mkdir $HOME/src/
- [isabell@stardust ~]$ cd $HOME/src/
- [isabell@stardust src]$ wget https://github.com/processone/ejabberd/archive/20.03.tar.gz
- [isabell@stardust src]$ tar xf 20.03.tar.gz
- [isabell@stardust src]$ cd ejabberd-20.03/
- [isabell@stardust ejabberd-20.03]$ ./autogen.sh
- [isabell@stardust ejabberd-20.03]$ ./configure --enable-user=$USER --prefix=$HOME/opt/ejabberd --enable-mysql --enable-new-sql-schema
- [isabell@stardust ejabberd-20.03]$ make install
+  [isabell@stardust ~]$ cd ~/downloads/
```
luto, line 53: There is no `~/downloads` directory on a standard uberspace. Please stick to `~/src` as it's more unixy or just drop the files directly in home.

coderkun: I changed `~/downloads` to the home directory `~/`.

coderkun:
Thanks for your feedback. I updated the PR:
* rebased ontop of master
* changed `~/downloads` to `~/`
* removed reference to non-existing rst file for web headers
* fixed some whitespaces

luto:
awesome, thank you! :)

luto:
Normally we only send out cups for new guides, but this is quite the overhaul. If you'd like one, just ping us at hallo@uberspace.de :)

beatstick:
Hi, 
The letsencrypt ssl certificates are renewed automatically on uberspace 7. After the renewal clients refuse to connect to ejabberd before a restart (or a config restart via ejabberdctl). Would it be possible to make the renewal script editable by users, so we can restart running services as well? Or maybe there is another workaround if that is not possible.

luto:
@beatstick you can restart ejabberd once every 30 or 60 days in a cronjob. On average you get a new cert every 60 days.

beatstick:
@luto: I have set it up that way now, but would ejabberd not have to be restarted right after the cert renewal for it to work?

coderkun:
Would it work to watch the certificate files via “inotify” or systemd and restart the ejabberd service when the files change, maybe with a grace period of a few seconds?

luto:
@beatstick certificates get renewed 30 days before they actually become invalid. That means you have 29-ish days to get the restart in automatically. It works great :)

@coderkun inotify works, sure. A cronjob is easy to understand for users, cannot miss any renewals and is easy to debug.

beatstick:
@luto: I think the clients complained the certificate had changed, not about it being invalid. So maybe inotify is the way to go? I already started looking into it some time ago, but it seemed a big hurdle to get this running on uberspace.

coderkun:
Just for reference: [systemd.path](https://www.freedesktop.org/software/systemd/man/systemd.path.html)

luto:
The clients will complain about a certificate change regardless of your mechanism (inotify, cron), because the certificate actually changes every ~60 days. It will be like that with any let's encrypt setup and there isn't much you can do about that, sorry.

Just a quick hint: please open new issues for problems you encounter with a guide. Thanks :)

beatstick:
Ah, ok, but gajim and conversations both stopped complaining about changed certificates after restarting ejabberd.  So maybe there would still be a use for the inotify script. Especially some older family members are easily irritated by all kinds of error messages. And restarting the service by hand is not a very satisfying or elegant solution. 
Maybe I just wait for a month or two and see how the cronjob works out. 

P.S.: I will open a new issue next time.  Thanks for the help.

PR #693, opened 2020-06-28
[Tiny Tiny RSS] Add guide for Tiny Tiny RSS

escuco:
Tiny Tiny RSS is an open source multiuser web-based news feed reader and aggregator (RSS/Atom). TTRSS is a server side AJAX powered application, the user only needs a web browser or one of the available Android clients. It supports multiple ways to share stuff, many plugins and themes, a JSON-based API and much more. The background service polls for subscribed web feeds at regular intervalls.


luto:
We had guides for TTRSS in both the old U6 wiki and U7 lab, but removed them since we don't think the developers of this tool have [views compatible to our own](https://github.com/Uberspace/lab/pull/137#issuecomment-394680305). So we cannot accept this PR, sorry :/ As a thank you for your work, you will get a cup anyway, though :)

Anyone looking for an RSS reader can use one of the [three alternatives](https://lab.uberspace.de/tags/rss) already documented.

escuco:
> We had guides for TTRSS in both the old U6 wiki and U7 lab, but removed them since we don't think the developers of this tool have [views compatible to our own](https://github.com/Uberspace/lab/pull/137#issuecomment-394680305). So we cannot accept this PR, sorry :/ As a thank you for your work, you will get a cup anyway, though :)

Ok, I understand - I didn't knew that. That's a shame, because the software is really nice and helpfull.

So, my very first pull request ever is virtually going into the trashcan ;)

But I agree with you - this kind of behavior doesn't correspond my attitude. I'm happy with the cup and I've learned a little bit reST :)

PR #609, opened 2020-04-04
[bitwarden_rs] add guide for bitwarden_rs

pxlfrk:
Hi, i've written the guide for [Bitwarden_rs](https://github.com/dani-garcia/bitwarden_rs), an [open source password manager](https://bitwarden.com/). Looking forward to Feedback from you! :tada:

closes #401

PR #694, opened 2020-06-29
[bitwarden_rs] update guide to fix cargo build issue

pxlfrk:
Hi!

in #609 @luto [told me that](https://github.com/Uberspace/lab/pull/609#issuecomment-629205490) Uberspace will support Rust in the Version 7.6.2, which is indeed correct - but only the `stable `release channel. **Bitwarden_rs** depends on the nightly release channel, so as described right now in the guide `cargo build -j 1 --release --features sqlite` will fail.

> Compiling pear_codegen v0.1.4
> error: failed to run custom build command for `pear_codegen v0.1.4`
> 
> Caused by:
>   process didn't exit successfully: `/home/passres/bitwarden_rs/target/release/b                               uild/pear_codegen-eb17718987529976/build-script-build` (exit code: 101)
> --- stderr
> Error: Pear requires a 'dev' or 'nightly' version of rustc.

Therefore I re-added the rust-installation section to the guide to fix this issue.

Another solution would be to include also the `nightly `and/or `dev `release channels to the default uberspace installation.

Looking forward for your Feedback to solve this issue.

### DIFF
--- a/source/guide_bitwarden.rst
+++ b/source/guide_bitwarden.rst
@@ -57,6 +57,75 @@ If you want to use Bitwarden_rs with your own domain you need to setup your doma
 Installation
 ============
 
+Install Rust
+-------------------
+
+.. note:: Since we need the rust nightly release channel for the installation, but Uberspace currently only provides the :manual:`stable release <lang-rust>`, we have to install it ourselves.
+
+To compile the project, we need to install the `rust toolchain`_. Install it via rustup and use the following configuration when asked:
+
+ * Enter ``2`` to customize the installation.
+ * Press ``Enter`` to use the host triple as default.
+ * Type ``nightly`` as default toolchain as it is required for bitwarden_rs.
+ * Type ``default`` to use the default install profile. 
+ * Enter ``y`` to add rust  to the PATH.
+ * Enter ``1`` to proceed with the installation
+
+.. code-block:: console
+ :emphasize-lines: 16,21,24,27,30,42
+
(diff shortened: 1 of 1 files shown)


luto:
sigh :sweat_smile: do you have any recommendations on which of the nightly releases to provide? is it fine to provide the latest nightly that actually builds the relevant tools as a "nightly" channel?

pxlfrk:
As far as i can tell i would recomment using rustup for this guide as it's mentioned in the official setup documentation; the latest nightly-release-channel should also work. At least i think so..

pxlfrk:
cc #708 for further information.

jonmz:
> cc #708 for further information.

Oh, thanks, I just scrolled through the open _issues_ regarding bitwarden, not through open PRs. Tbh I am not really keen on providing nightly builds unless absolutely necessary, and for that I would like to get at least some basic understanding _why_ a nightly build (even one from 2018) is needed for Pear but a stable build from 2020 would not do it... we are strictly sticking with stable versions on other programming languages as well. If possible I would prefer Rust being no exception, but will keep this PR open for now.

pxlfrk:
I absolutely agree with you, in theory the stable release should work fine. I'd prefer waiting for the result of the issue you referenced.

luto:
postponed until we've got some replies in the issue.

luto:
This should be fixed in a future bitwarden/pear release: https://github.com/SergioBenitez/Pear/issues/23#issuecomment-660393674. Thanks for the PR, though!

PR #701, opened 2020-07-05
Replace /home/isabell/ with $HOME

MaxG87:
Using $HOME instead of '/home/isabell/' makes snippets more friendly for copy-paste workflows.
The replacement was done only in snippets of Bash code.


luto:
nice, thank you!

PR #702, opened 2020-07-05
Fix put_url in configuration

thisven:
The value for put_url has to contain "upload." in front of @HOST@ to apply to subdomain configuration in the beginning of this guide. Otherwise uploading media files won't work.


luto:
thanks!

PR #715, opened 2020-07-20
[Jenkins] Update guide to work with Java 14

Snapstromegon:
Jenkins officially does not support Java 14 as they only allow Java 8 and 11.
This is not going to change in the near future, but they have a flag to allow running Jenkins with the newest Java version.
This change introduces this flag to the guide.

### DIFF
--- a/source/guide_jenkins.rst
+++ b/source/guide_jenkins.rst
@@ -111,7 +111,8 @@ We create the service file ``~/etc/services.d/jenkins.ini`` and fill it with:
 
  [program:jenkins]
  directory=%(ENV_HOME)s/Jenkins/Jenkins_home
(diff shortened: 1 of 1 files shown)


source/guide_jenkins.rst
```diff
@@ -111,8 +111,8 @@ We create the service file ``~/etc/services.d/jenkins.ini`` and fill it with:
 
  [program:jenkins]
  directory=%(ENV_HOME)s/Jenkins/Jenkins_home
- command=java -jar ../jenkins.war
-
+ command=env JENKINS_HOME=%(ENV_HOME)s/Jenkins/Jenkins_home java -jar %(ENV_HOME)s/Jenkins/jenkins.war --httpPort=8080 --enable-future-java
```
luto, line 114: Please use `environment=JENKINS_HOME="%(ENV_HOME)s/Jenkins/Jenkins_home"` [like so](http://supervisord.org/subprocess.html#subprocess-environment).

Snapstromegon: Will change tomorrow.

luto:
nice! thank you :)

Issue #696, opened 2020-07-02
Add guide for Ackee


tobimori:
[Ackee is a node.js analytics software](https://github.com/electerious/Ackee)

tobimori:
Began working on this: https://github.com/tobimori/uberspace-lab/tree/ackee

tobimori:
Added in #716

PR #716, opened 2020-07-21
Feature/ackee

felixfoertsch:
#696 I installed it today for my own analytics purposes and finished the guide based off @tobimori's branch. Maybe @tobimori can do the first review. 😁


source/guide_ackee.rst
```diff
@@ -34,47 +32,81 @@ Ackee_ is released under the `MIT License`_.
...
-  remote: Counting objects: 100% (1528/1528), done.
-  remote: Compressing objects: 100% (884/884), done.
-  remote: Total 7991 (delta 1016), reused 1083 (delta 641), pack-reused 6463
-  Receiving objects: 100% (7991/7991), 1.97 MiB | 3.95 MiB/s, done.
-  Resolving deltas: 100% (5471/5471), done.
-  [isabell@stardust ~]$ 
+  ACKEE_MONGODB=mongodb://<admin>_mongoroot:<password>@localhost:27017/admin
+  ACKEE_USERNAME=<ackee_username>
+  ACKEE_PASSWORD=<ackee_password>
+  ACKEE_ALLOW_ORIGIN="https://my-domain-other-than-isabell-on-stardust.de"
```
tobimori, line 77: `https://my-domain-other-than-isabell-on-stardust.de` seems like a weird choice here. Ackee doesn't  automatically allow `isabell.uber.space` for CORS (or something else). You always need to supply a domain, so I'd use another placeholder.

source/guide_ackee.rst
```diff
@@ -34,47 +32,81 @@ Ackee_ is released under the `MIT License`_.
...
+  [isabell@stardust Ackee]$ yarn
+
+Configuration
+=============
+
+Ackee Config
+------------
 
-Clone the `GitHub repository of Ackee`_:
+Create a new, empty ``.env`` file in the root of the Ackee_ directory. Add the following content, replacing the placeholders. The environment variables ``ACKEE_USERNAME`` and ``ACKEE_PASSWORD`` define the credentials to log into the web interface. If you want to use Ackee_ to analyze a domain other than ``isabell.uber.space``, you have to add it under ``ACKEE_ALLOW_ORIGIN`` (read more about the CORS headers `in the Ackee documentation <https://github.com/electerious/Ackee/blob/master/docs/CORS%20headers.md>`_).
```
tobimori, line 70: You always need to add a CORS header for every site you want to use the tracking script. No matter if it's your standard Uberspace domain or not.

source/guide_ackee.rst
```diff
@@ -34,47 +32,81 @@ Ackee_ is released under the `MIT License`_.
...
+  ACKEE_PASSWORD=<ackee_password>
+  ACKEE_ALLOW_ORIGIN="https://my-domain-other-than-isabell-on-stardust.de"
+
+
+Web Backend Config
+------------------
+
+.. note::
+
+    Ackee_ is running on port 3000.
```
tobimori, line 85: Maybe add a hint on how to change your port in `.env` as `3000` may be already used.

source/guide_ackee.rst
```diff
@@ -34,47 +32,81 @@ Ackee_ is released under the `MIT License`_.
...
+  command=yarn start
+  autostart=yes
+  autorestart=yes
+
+.. include:: includes/supervisord.rst
+
+Finishing installation
+======================
+
+Go to ``https://isabell.uber.space``. Login using the credentials defined in section `Ackee Config`. Add the domains you want to analyze in the web interface: After adding, it shows a tracking snippet you can add to these pages. If data does not show up after implementing the tracking script, remember that cross site tracking always takes extra consideration of `CORS <https://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`_.
```
tobimori, line 107: I wouldn't specify the domain here. The user decided that while setting up the Web backend.

source/guide_ackee.rst
```diff
@@ -34,47 +32,81 @@ Ackee_ is released under the `MIT License`_.
...
 
 .. _Ackee: https://ackee.electerious.com/
-.. _GitHub repository of Ackee: https://github.com/electerious/Ackee
+.. _MongoDB: https://mongodb.com
 .. _MIT License: https://github.com/electerious/Ackee/blob/master/LICENSE
 
 ----
```
tobimori, line 113: An update section is missing.

source/guide_ackee.rst
felixfoertsch: As far as I understand it, the baseline for all projects is always a new Uberspace, so that should not be the case. If somebody hosts multiple projects on one space, they should account for the ports.

source/guide_ackee.rst
felixfoertsch: The default for projects is usually `/`. This is why this route should be correct.

source/guide_ackee.rst
felixfoertsch: I have not yet made an update to Ackee, so I haven't done an update yet. Maybe you can add it?

source/guide_ackee.rst
felixfoertsch: Are you sure? Shouldn't it be possible to analyze the own host? E.g. test.isabell.uber.space?

source/guide_ackee.rst
tobimori: Stopping the service and running git pull should work.

source/guide_ackee.rst
tobimori: No, that's not possible. You'd need to specify some kind of wildcard for that.

source/guide_ackee.rst
tobimori: That's not what I meant. I wouldn't specify the *domain* here, also not the route. Users may see that as a general link under which all Ackee instances on the world can be found etc.

felixfoertsch: If I understand you correctly, you would remove:

> Go to ``https://isabell.uber.space``.

I disagree on this, since the guide sets up Ackee on the root, so this link leads to the web interface. As far as I understand it, setting up a system on root is the baseline the guides.

tobimori: Seems like there is no correct behaviour there. Either a placeholder domain is uaed, but not as an anchor tag, or isabell.uber.space is used as an example, but with the indication to please use your specified domain.

source/guide_ackee.rst
felixfoertsch: I updated the guide accordingly. It should make more sense now. I added a note for multple domains.

source/guide_ackee.rst
```diff
@@ -74,15 +74,15 @@ Create a new, empty ``.env`` file in the root of the Ackee_ directory. Add the f
   ACKEE_MONGODB=mongodb://<admin>_mongoroot:<password>@localhost:27017/admin
```
tobimori, line 74: Do you think it would make sense to emphasize the lines here to be changed?

felixfoertsch: Totally. Is it possible to highlight individual words in a code block? I only found `:emphasize-lines: X` but that--unsurprisingly--only emphasizes whole lines.

tobimori: Nope. You'd need to emphasize all line where something needs to be changed.

felixfoertsch: Yeah, too bad. I think we can leave out the emphasize then, since every line in this block has to be changed in some way.

```
ACKEE_MONGODB=mongodb://<admin>_mongoroot:<password>@localhost:27017/admin
ACKEE_USERNAME=<ackee_username>
ACKEE_PASSWORD=<ackee_password>
ACKEE_ALLOW_ORIGIN="https://isabell.uber.space"
```

tobimori: Yes, that's why I asked :)

luto:
nice one :) thank you! :tada:

PR #719, opened 2020-07-23
[nextcloud] add script for command line based updater

EV21:
Since the web-based updater is still prone to errors and slow, the command line tool should be used.
In order to save some time for the user I provide useful commands as a script.


luto:
thank you :tada:

PR #720, opened 2020-07-23
[STYLE.md] explain includes and update boilerplate

EV21:
To prevent contributors copy-paste the boilerplate with redundant content I give an example and set includes in the boilerplate.


luto:
whoops, that doc seems a bit out-of-date. I'll take a closer look later :)

PR #721, opened 2020-07-24
[ethercalc] [isso] fix building warning

taddydevil:
again a suggestion to fix the build warnings I got when building the lab local.


luto:
thanks!

PR #724, opened 2020-07-29
[Lychee] Update for version 4.x

jce-zz:
Updated for version 4.x of Lychee by putting Lychee data outside the DocumentRoot and creating a Symlink to the "public" subfolder as required.


luto:
thank you! :)

PR #727, opened 2020-08-03
Update guide_redis.rst

Hemmho:
[redis] Compilation section removed and set reference for preinstalled binaries.


luto:
nice! thank you :)

PR #730, opened 2020-08-06
[PostgreSQL] Fixes a spelling mistake

Fryboyter:
"Replace replace" makes no sense. Therefore changed to "Please replace" as it is also shown in other parts of the manual.


luto:
Thanks!

PR #732, opened 2020-08-06
[CONTRIBUTING.md] add note on commit feeds

EV21:
> If you want to stay informed of changes to your guide or a guide in your interest you can subscribe to its commit feed.  
> e.g. for Nextcloud it would be: https://github.com/Uberspace/lab/commits/master/source/guide_nextcloud.rst.atom

I think some contributors would like to watch changes to a specific guide as they may not like the default flood of watch notifications by GitHub.

I'm not sure if this is the right place for this information. Maybe each guide could also have such an information like the "View Changelog" link.


luto:
Thanks!

PR #733, opened 2020-08-08
[prosody] fix muc certificate issue

taddydevil:
Add certificate generation for ``conference.isabell.org``


luto:
whoops! thank you!

PR #740, opened 2020-08-26
Add grocy guide

### LINKED ISSUE #464: Guide for Grocy
It would be nice to have a guide for grocy https://github.com/grocy/grocy

grocy is a web-based self-hosted groceries & household management solution for your home.

jfowl:
Should close #464


source/guide_grocy.rst
```diff
@@ -0,0 +1,139 @@
...
+
+.. tag_list::
+
+Grocy is a web-based self-hosted groceries & household management solution for your home.
+
+----
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * Git_
```
EV21, line 24: Currently you don't use any git command in your guide. So I don't know why it is a need-to-know.

There is a way, but I don't think this is the way the target users should use it as it requires some other steps.

Grocy README.md:
> Alternatively clone this repository (the release branch always references the latest released version, or checkout the latest tagged revision) and install Composer and Yarn dependencies manually.

source/guide_grocy.rst
```diff
@@ -0,0 +1,139 @@
...
+  .. image:: _static/images/grocy.svg
+      :align: center
+
+####
+Grocy
+####
+
+.. tag_list::
+
+Grocy is a web-based self-hosted groceries & household management solution for your home.
```
EV21, line 18:
```suggestion
Grocy_ is a web-based self-hosted groceries & household management solution for your home.
```

EV21 (commented):
It's an interesting app, this is something I was missing. 😃 

quick review, not tested

source/guide_grocy.rst
```diff
@@ -0,0 +1,139 @@
...
+ [isabell@stardust isabell]$ 
+
+Step 2
+------
+
+Unpack the downloaded grocy release
+
+::
+
+ [isabell@stardust isabell]$ unzip -q grocy_latest.zip
```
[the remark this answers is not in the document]
jfowl: This is a great idea, did not think about this. Makes it a lot cleaner. Thanks :)

source/guide_grocy.rst
jfowl: Fixed.

source/guide_grocy.rst
```diff
@@ -0,0 +1,144 @@
...
+Step 3
+------
+
+If you want to only host grocy on your Uberspace (best practice since Uberspace 7) you should link your
+DocumentRoot to grocy's public folder:
+
+::
+
+ [isabell@stardust isabell]$ cd /var/www/virtual/$USER/
+ [isabell@stardust isabell]$ rm -r html
```
luto, line 78:
```suggestion
 [isabell@stardust isabell]$ rmdir html
```

Guides should assume a fresh uberspace, which means an empty `html` dir. Please change this to `rmdir`, so it fails if there is something in the directory. Users can then still choose to delete the contents on their own.

source/guide_grocy.rst
```diff
@@ -0,0 +1,144 @@
...
+Please note that this will require further tweaking in the config file.
+
+Step 4
+------
+
+Copy the default config file to ``data/config.php`` and edit it to your liking
+
+::
+
+ [isabell@stardust isabell]$ cp config-dist.php data/config.php
```
luto, line 98: Both the shell prompt and the last `cd /var/www/virtual/$USER/` drop us in `/var/www/virtual/$USER/`, but there is no `config-dist.php` in that directory. Is this missing a `cd /home/$USER/grocy`? Please adjust the shell prompt to match as well.

source/guide_grocy.rst
```diff
@@ -0,0 +1,144 @@
...
+ [isabell@stardust isabell]$ nano data/config.php
+ [isabell@stardust isabell]$ 
+
+
+You will probably want to change the ``CURRENCY`` to ``EUR`` and ``CULTURE`` to ``de``.
+
+If you decided to host your grocy installation in a sub folder, also change the ``BASE_URL`` to your full url, e.g. ``http://isabell.uber.space/grocy/``.
```
luto, line 105:
```suggestion
If you decided to host your grocy installation in a sub folder, also change the ``BASE_URL`` to your full url, e.g. ``http://isabell.uber.space/``.
```

The default installation assumes a link from `/var/www/virtual/$USER/html` to `/home/$USER/grocy`. Please use the matching URL here or provide both options.

source/guide_grocy.rst
```diff
@@ -0,0 +1,144 @@
...
+
+.. include:: includes/web-domain-list.rst
+
+Installation
+============
+
+Step 1
+-----
+
+Download grocy to your home directory.
```
luto, line 44: Installing php code into the home directory will not work because the web server is only allowed to access files in `/var/www/virtual/...`:

```
[Mon Aug 31 13:29:44 2020] [error] [pid 18613] request.c(1204): [client 2a02:8388::] AH00037: Symbolic link not allowed or link target not accessible: /var/www/virtual/grocy2/html
```

Did you actually test this? :sweat_smile: 

You can install the files into `/var/www/virtual/$USER/grocy` instead.

luto (changes requested):
Assuming grocy is installed in `/var/www/virtual/...`, this works great :) I left a couple of change requests and fixed a few formatting issues.

source/guide_grocy.rst
EV21: Before I tested it I mentioned it as an option in my review but then I ran into exactly that issue :laughing: - So I removed that passage quickly. I should have mentioned that. 
This is a working structure like mentioned before:
```
/var/www/virtual/isabell/
├── grocy
├── someStaticFiles.example.com
└── html -> grocy/public
```

re-add this line before doing the magic and everything should work:
```
[isabell@stardust ~]$ cd /var/www/virtual/$USER/
```

luto:
fixed the remaining comments and merged manually :) thank you! :tada: https://lab.uberspace.de/guide_grocy.html

@jfowl I can't find your mail address in the guide or on your profile. Please drop us a quick message at hallo@uberspace.de to get your thank-you-gift :)

PR #741, opened 2020-08-26
Update to PostgreSQL 12.4

qbi:
I updated the document to use the current version (12.4) of PostgreSQL. This uses another auth method and also the output of the initdb changed a bit.


luto:
thank you! :tada: :)

PR #743, opened 2020-08-28
[pleroma] add instructions for erlang version

taddydevil:
New dependencies in the pleroma version 2.1.0 needs minimum OTP22.


luto:
oh, thanks very much! :)

PR #746, opened 2020-09-03
do not assume ~/html to be empty

luto:
Because it will contain an explanatory `index.html` in future releases. So do not delete it, but just move it away.

Alternatively change this to `rm html/index.html; rmdir html` to make it clearer what is happening?


noave:
because we always assume a fresh uberspace account, I would rather go with the `rm htm/index.html` even when this could mean loss of data in very exceptional cases (where users did something wrong and did not take care for executing a `rm` command). But the `mv` solution would mean to create some trash data.

luto:
v7.7.8 (rolling out starting today) will no longer place a index.html overriding a (pre) existing index.php.

PR #749, opened 2020-09-09
[fluentd] add fluentd guide

andipabst:


source/guide_fluentd.rst
```diff
@@ -0,0 +1,192 @@
...
+ Installed /home/isabell/fluentd/fluent.conf.
+ [isabell@stardust ~]$
+
+Edit this configuration and comment out the following lines:
+
+.. code-block::
+
+ # match tag=system.** and forward to another fluent server
+ <match system.**>
+   @type forward
```
luto, line 120: This receives data from another flutend and pushes it along, right? What do you think about a more illustrative example like sending `~/logs/supervisord.log` to another fluentd?

andipabst: Yes, that's correct. This is part of the default fluentd configuration, created by `fluentd --init`. I assumed, that the user does not have another fluentd instance set up and thus included the instruction to comment out this part of the configuration. Should I make this more clear by adding a `#` in front of each line?

luto: oh, whoops. I totally didn't get that while reading it, so probably, yes :sweat_smile:

luto:
thank you! :tada:

PR #751, opened 2020-09-12
Update guide_radicale.rst

mackeprm:
Added a short note on how to setup radicale in a subfolder. Took me an hour to figure this out, so maybe future versions of me can be helped here :)


luto:
thanks!

PR #753, opened 2020-09-13
[postgres] guide improvements.

thigg:
Mostly rewording.
A small add for unix domain sockets.

### DIFF
--- a/source/guide_postgresql.rst
+++ b/source/guide_postgresql.rst
@@ -34,7 +34,7 @@ Create a working directory.
  [isabell@stardust ~]$ cd ~/postgres/
  [isabell@stardust ~]$
 
-Download version 9.6.10. A list of supported major and beta releases can be found on the PostgreSQL `download server`_.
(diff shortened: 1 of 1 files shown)


source/guide_postgresql.rst
```diff
@@ -251,6 +248,17 @@ Edit ``~/opt/postgresql/data/postgresql.conf`` and set the key values ``listen_a
                                         # (change requires restart)
 
 
+Because postgres defines the socket as abstract, you cannot see it directly in the filesystem.
```
luto, line 251: Hm, I can see the socket just fine using `ls -la`. Which utility isn't able to see it? Or is this new in more recent pg versions?

luto:
Thanks! I left a small note, otherwise this looks very good to me :)

source/guide_postgresql.rst
thigg: my mistake, didnt try -la because thats usually my default anyway. gonna change the text

thigg:
fixed it

luto:
Thanks!

PR #755, opened 2020-09-15
fix for 404 error in css

michi-zuri:
### Before: 
image
### After:
image

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -81,7 +81,7 @@ nav li {
     float: right;
     width: 30%;
     height: 30%;
-    shape-outside: url(/en/_images/atom.svg);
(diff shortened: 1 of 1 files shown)


luto:
Thanks! The page already looks like the "after" screenshot for me :thinking: Your browser doesn't seem to follow the initial 301 response. Which one are you using?

michi-zuri:
To be clear: this only applies to the local development copy. In the deployed version on netlify, this is not an issue, but localhost:8000/en/_images/atom.svg returns a 404 page not found error. 

I'm seeing this on macOS with Firefox 80.0.1
The effect on the layout on Safari and Chrome is hardly noticable, but still the 404 error is returned by the dev server.

luto:
oh! I see. Thanks!

PR #756, opened 2020-09-15
add starship guide

michi-zuri:
my first contribution, please let me know if there are style issues etc.


Hafpaf:
Great work, Starship seems like a pretty cool shell. I can see a few minor things.

1. To make your author link work, it can be fixed with something like this:
```rst
`Python 
<http://www.python.org/>`_
```

2. Should there perhaps be installation instructions for Nerdfonts, or is this outside the scope of the guide?
If not, its totally fine for it to appear in _Prerequisites_.

4. In end of the guide, just before `.. author_list::`, add which version of Starship and Uberspace version it was tested with. It could be like this
```rst
Tested with Starship v0.44.0 and Uberspace version 7.7.7.0
```
  Your Uberspace version can be found in the welcome message connecting to the server using SSH.

michi-zuri:
Thanks for the review. 
Re point 1: I couldn't figure out how to make the author link work and kept the same style as all other guides in the uberlab.

luto:
Thank you!

PR #757, opened 2020-09-19
Update guide_bitwarden.rst

cblte:
corrected some typos and updated the versions of the web-vault as well rewrote the update section a little bit to make it less error prone.


luto:
Thanks! :tada:

PR #761, opened 2020-09-21
[nextcloud] Add reference to Nextcloud release feed on Github

dejIO:
While there is an official RSS feed, it's very noisy and doesn't only contain information about new releases. The Github releases feed serves this purposes (but it is very bare and also includes unstable releases). A more advanced solution might use the output of "occ update:check" to send an email to the user whenever there is a new release.


luto:
Thank you :)

PR #762, opened 2020-09-22
Add note for additional document root

duese:


source/guide_framadate.rst
```diff
@@ -150,6 +150,10 @@ To enable URL rewriting, to have links in the form ``https://isabell.uber.space/
 
   If you choose not to do this, set ``const URL_PROPRE = false;`` in ``~/html/app/inc/config.php`` or you won't be able to access your polls.
 
+.. note::
+
+  If you're using a subdomain, you need to add ``RewriteBase /`` to the ``.htaccess`` file as described `in the section for additional DocumentRoots <https://manual.uberspace.de/web-documentroot.html#additional-documentroots>`_.
```
luto, line 155: Please us a tag like this
```rst
:manual_anchor:`npm <lang-nodejs.html#npm>`
```
Thanks! :)

luto (changes requested):

source/guide_framadate.rst
EV21: I think _subdomain_ is the wrong word in this context or I don't understand.
```suggestion
  If you're using an :manual_anchor:`additional DocumentRoot <web-documentroot.html#additional-documentroots>` , you need to modify the ``.htaccess`` file within the DocumentRoot by adding ``RewriteBase /`` to it.
```

PR #763, opened 2020-09-22
Update guide_mailman-3.rst

mohs8421:
changing pip version to 3.8, as 3.6 no longer worked
removing names from commands to ease copying

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -68,7 +68,7 @@ Install Mailman 3 and its dependencies via pip.
 
 ::
 
- [isabell@stardust ~]$ pip3.6 install --user mailman hyperkitty postorius mailman-hyperkitty whoosh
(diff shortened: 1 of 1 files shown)


noave:
thank you very much for the `pip3.8`, it actually works now. But there are some issues here:

- considering our [style guide](https://github.com/Uberspace/lab/blob/master/STYLE.md) the `[isabell@stardust ~]$` should be kept, even if its make harder to use copy paste. 
- also, please mind the naming of your commits, see https://github.com/Uberspace/lab/blob/master/STYLE.md#add-your-changes-to-the-uberspace-lab
- and more importantly, the use of python/pip3.8 needs to be used in all commands, otherwise it wont work

Again, I thank you very much for this little but important fix, please tell me if you can update your PR with the suggestions mentioned here :)

noave:
The fixes have been merged with #772 thanks for your input

PR #765, opened 2020-09-23
added a note for failed mailman installation

jonashoechst:
some package required by mailman (`flufl.i18n`) requires a newer version of `setuptools`, which leads to non-trivial failure of the installation. This commit includes a hint how to update the respective package.

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -66,6 +66,15 @@ Get Mailman 3
 -------------
 Install Mailman 3 and its dependencies via pip.
 
+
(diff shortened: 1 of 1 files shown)


noave:
thanks, but did you see #763 already? did this problem also occur with pip3.8 for you?

jonashoechst:
I'll give it a try, will probably work. It seems there is work required at #763 to pull the changes. After confirming I could create a respective pull request, if wanted.

noave:
yeah would be happy if someone works on it :tada:

PR #766, opened 2020-09-23
[mailman3] updated python version, added hyperkitty installation hints

jonashoechst:
These commits update the guide to use python3.8, as proposed in #763, and include an example configuration for mailman hyperkitty. It fixes #765 by using the more recent python installation and makes #764 obsolete, by making the guide usable again.

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -27,8 +27,6 @@ Mailman 3
   * **HyperKitty**; the web-based archiver
   * **Mailman client**; the official Python bindings for talking to the Core’s REST administrative API.
 
-.. note:: This guide only focuses on mail distribution and is missing information on how to archive emails. Please see the `hyperkitty documentation <https://hyperkitty.readthedocs.io/en/latest/install.html#connecting-to-mailman>`_ for this and create a PR 🎉.
-
 ----
 
 .. note:: For this guide you should be familiar with the basic concepts of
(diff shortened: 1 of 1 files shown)


jonashoechst:
There's an potential problem that I came across: `includes/install-uwsgi.rst` contains a pip3.6 reference, which cannot be updated, since it is also included in the guides `etesync-server`, `babybuddy` and `django`.

noave:
thank you very much for your PR, I will check it today. 

> There's an potential problem that I came across: `includes/install-uwsgi.rst` contains a pip3.6 reference, which cannot be updated, since it is also included in the guides `etesync-server`, `babybuddy` and `django`.

I would say it is okay to just use the include text as plaintext in the mailman3 guide with fixed 3.8 version for now. Then create an issue that we have to check the other guides if they work with 3.8 so we can update the include file and use it again in maillman3.

noave:
I could test the changes now (sorry for the delay) and hypekitty is so far working, that I can read the messages sent via mailinglist. But were you also able to reply via hyperkitty and start a new thread? That is yet not working for me, the without any error the data is not saved to db or sent to the mailinglist.

jonashoechst:
It works in one of my recently configured instances, just checked. 
I've had this problem before in another configuration but it was an error in my configuration. Could you check, wether the smtp config in both `mailman-suite/settings.py` as well as var/etc/mailman.cfg is correct?

noave:
no this was fine, because the mailing works quite well. But there was another little trap I found when I did some digging in the code:

```
# hyperkitty/lib/postings.py
if not settings.DEBUG:
        msg.send()  # Don't send mail in debug mode, just in case...
```

so, apart from the main debug settings in `settings.py` where you can pipe mails to the console while debugging, hyperkitty here completely disables the creating of mails when its in Debug mode, so much about testing hyperkitty when debug mode is enabled :)

Anyway, I wanted to change the Debug management in the mailman3 guide, but so far your PR works :tada:

noave:
I would merge it into main(master), but because this will be incomplete without fixing the uwsgi problem you mentioned, could you add this to the PR like I proposed here (see below)? 

> I would say it is okay to just use the include text as plaintext in the mailman3 guide with fixed 3.8 version for now. Then create an issue that we have to check the other guides if they work with 3.8 so we can update the include file and use it again in maillman3.

PR #771, opened 2020-09-27
note to delete uberspaces default index.html to enable Wordpress frontend

paradonym:
Note to delete the default index.html because it prevents wordpress index.php to load the frontend (browsers prefer html over php).


luto:
Thanks! added WP to https://github.com/Uberspace/lab/pull/746.

PR #777, opened 2020-09-29
Rocket.Chat guide update

d-sko:
Hi,

I've updated my Rocket.Chat guide because Rocket.Chat has released version [3.7.0](https://github.com/RocketChat/Rocket.Chat/releases/tag/3.7.0), which has my pull request https://github.com/RocketChat/Rocket.Chat/pull/17012 included, so we can now set `TMPDIR` and the workaround in my original guide is not longer necessary.


luto:
Awesome! Thank you :)

PR #797, opened 2020-10-16
[gitlab] Create a gitlab guide

MaelStor:
I created a guide for GitLab in a recent version (v13.4.2). Any feedback on this would be welcome :)

EDIT:
Since I faced some crashes using the gitlab ci I needed to rearrange things quiet a bit and marked this PR as draft.

### DIFF
--- a/source/guide_gitlab.rst
+++ b/source/guide_gitlab.rst
@@ -1,0 +1,1657 @@
+.. author:: Michael Strobler <maelstor@posteo.de>
+
+.. tag:: lang-ruby
+.. tag:: lang-yaml
+.. tag:: audience-developers
+.. tag:: continuous-integration
+.. tag:: automation
+
+.. highlight:: console
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/gitlab-logo.png
+      :align: center
+
+#############
+GitLab
+#############
+
+.. tag_list::
+
+`GitLab`_ is a web-based DevOps lifecycle tool that provides a Git-repository
+manager providing wiki, issue-tracking and continuous integration and deployment
+pipeline features, using an open-source license, developed by GitLab Inc. [#f1]_
+
+----
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * :manual:`supervisord <daemons-supervisord>`
+  * :manual:`web backends <web-backends>`
+  * :manual:`ssh <basics-ssh>`
+  * :manual:`resources <basics-resources>`
+  * :manual:`ports <basics-ports>`
+  * :manual:`ruby <lang-ruby>`
+  * :manual:`nodejs <lang-nodejs>`
+  * :manual:`perl <lang-perl>`
+  * :manual:`gcc <lang-gcc>`
+
+
+Prerequisites
+=============
+
+Requirements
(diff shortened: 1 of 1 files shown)


MaelStor:
Solved the crashes during usage of the gitlab ci by moving the postgresql installation to the host with the sidekiq installation. I also limited the puma webserver memory usage with a lower threshold for the puma memory killer.

MaelStor:
Solved the issue that no emails were sent by gitlab because the assets weren't compiled on the sidekiq host.

luto:
Thanks for your submission! It's clear that you put a lot of work into this.

Managing three separate apps as well as their various dependencies also seems like a big task. Maintaining this guide in the future seems like an even bigger one. The fact that they don't even fit into one uberspace doesn't make the situation easier. Additionally, Gitlab instances on U6 caused us (and users) quite a few headaches in the past :/ So, I am not sure how to proceed here. I guess I'll discuss with the team next week and get back to you.

Please open an issue before starting a 2.000 line guide, as documented in [CONTRIBUTING.md](https://github.com/Uberspace/lab/blob/master/CONTRIBUTING.md) so we can discuss things like this beforehand without potentially wasting many hours of work.. :/

MaelStor:
Hi @luto. First of all thanks for your feedback and I'm really sorry that I came straight out to it instead of creating an issue in the first place. I just worked on it and after days! I solved all the problems and was so happy with it, that I wrote it all down in a uberlab guide without thinking too much about anything else. Honestly I overlooked the `Contribute.md` and thought about the pull request like an issue. I had problems on U6 too with my GitLab instance but on U7 and after working out the solutions, GitLab is running stable. Although it needs 2 accounts for that, the second one doesn't need to be dedicated to gitlab alone, if that helps. I admit that it takes up to a day to install everything into the right places but actually it's fun and maybe the right thing to do on a rainy sunday ;) The installation is meant for a small development team, definitely not a 100 users but I hope other ubernautes following the guide see that after reading through the Prerequisites section what should avoid headaches later. See you next week.

luto:
I am very sorry, but we cannot accept this guide for the following reasons:

1. It requires users to create two uberspaces. Each guide should be able to run in just one.
2. A guide that takes an entire day to set up does not seem practical for users trying this and, more importantly, us testing the guide. Users are free to use their time in whatever way they please, but we have limited resources.
3. Such a complex setup will create loads of problems for us providing support to customers who run into problems.
4. The [official requirements](https://docs.gitlab.com/ee/install/requirements.html) are 4 cores as well as 4GB of RAM. The latter exceeds the limits of even two uberspaces.
5. There are much lighter alternatives offering similar features, such as [gitea](https://lab.uberspace.de/guide_gitea.html).

Still, I recognize that you put a lot of labor and love into this :/ As a small token of thanks, we'd like to offer to two [cups](https://github.com/Uberspace/lab/blob/master/CONTRIBUTING.md#reward) instead of the usual one (you need two uberspaces, after all!), even though we won't include the guide. Please send a quick message to hallo@uberspace.de, if you are interested.

I hope this doesn't discourage you from further work on the lab. To prevent this in the future, please talk to us (and the community) by opening an issue beforehand.

PR #804, opened 2020-10-26
[akaunting] Remove PHP version from Akaunting guide

felurx:
https://akaunting.com/docs/requirements
Akaunting requires "PHP 7.2.5 or higher". Since 7.2 is the oldest available version, I think we can just drop the notice about the requirement. I could also update it to 7.4 and mention that anything 7.2 or above should work, if that'd be better.


luto:
Thanks!

PR #809, opened 2020-10-28
Update ezmlm guide

### LINKED ISSUE #800: install ezmlm with existing lists
if you install ezmlm and take over lists from an old system installation, you might get the error `ezmlm-make: fatal: unable to stat /etc/ezmlm/de: file does not exist`. Then you have to remove `/etc/ezmlm/de` from ~/ezmlm/LIST/ezmlmrc`

we could put a warning considering this in the lab, especially thats the case for users moving from U6 to U7

noave:
cleaned up ezmlm guide, but still not include a hint for the ezmlrc problem with existing lists fro U6 to solve #800

### DIFF
--- a/source/guide_django.rst
+++ b/source/guide_django.rst
@@ -57,7 +57,7 @@ Install django
 
 ::
 
- [isabell@stardust ~]$ pip3.6 install django --user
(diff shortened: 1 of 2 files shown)


luto:
feel free to reopen, once this is ready.

PR #812, opened 2020-11-02
[mosquitto] Add guide on Mosquitto configuration

transistortim:
Mosquitto is an open source MQTT message broker written in C.

We finally managed to update our guide to the preinstalled mosquitto (see also https://github.com/Uberspace/lab/pull/657).


source/guide_mosquitto.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Step 1
+------
+
+Copy the default (preinstalled) configuration to ``~/etc/mosquitto``.
+
+.. code-block:: console
+ :emphasize-lines: 1
+
+ [isabell@stardust ~]$ cp /etc/mosquitto/mosquitto.conf ~/etc/mosquitto/
```
luto, line 61: Please create that directory beforehand, so the `cp` doesn't fail.

```suggestion
 [isabell@stardust ~]$ mkdir -p ~/etc/mosquitto/
 [isabell@stardust ~]$ cp /etc/mosquitto/mosquitto.conf ~/etc/mosquitto/
```

source/guide_mosquitto.rst
```diff
@@ -0,0 +1,171 @@
...
+Luckily, Mosquitto is preinstalled on Uberspace hosts.
+We just need to configure it (see below).
+
+
+Configuration
+=============
+We're setting up the broker to be reachable via a dedicated TCP port (as opposed to MQTT over WebSockets).
+The communication will be secured using TLS encryption and password authentication.
+
+Step 1
```
luto, line 53: Please name your steps, e.g. "Copy default config", "Open firewall port", etc. This makes them much easier to find, in case you are only reading part of the guide, are continuing an earlier attempt or are debugging an existing installation. I am aware that older guides use numbered steps, but the has [been updated](https://github.com/Uberspace/lab/commit/9a55c0c79f520ecedd8964fb630e90903c192de) for a couple of months now.

luto (changes requested):
Thanks for the guide. It looks great! I left two small suggestions.

source/guide_mosquitto.rst
transistortim: Thanks for your feedback. We updated our guide accordingly.

PR #819, opened 2020-11-06
Update and improve taskd guide

ebroda:
Updated and partly improved the taskd guide as I was installing it for me.
Tried to let it run with the web backends but didn't work; so I kept it with a port.


luto:
Thank you!

PR #818, opened 2020-11-06
Fix typos in guides

ebroda:
I re-ran some spell checking on the lab and corrected some typos.

PR #820, opened 2020-11-06
Spell checking for Sphinx

ebroda:
As raised by @luto in #818, here the current version of my spell-checking setup and dict.
There are still some false-positives; I didn't include all words in the dict. 

see also the old commit #608
Unfortunately last time there were some problems with netlify; I'm not sure if they have been resolved.
https://github.com/Uberspace/lab/pull/608#issuecomment-621274531


luto:
Netlify seems to be fine this time around. Thank you! :)

PR #821, opened 2020-11-06
Remove build warning for prosody

ebroda:
The guide for prosody was missing the one space identation in one code block, created warnings during build and weren't rendered correctly.


luto:
Thanks!

PR #824, opened 2020-11-08
[open-port] Fix syntax highlighting

transistortim:
Upon writing a guide we noticed syntax highlighting in the "open-port" include differs from other code blocks. It uses code-block [bash](https://pygments.org/docs/lexers/#pygments.lexers.shell.BashLexer), which is intended for bash scripts, but not for bash sessions. This PR changes it to [console](https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer).


luto:
Thanks!

PR #827, opened 2020-11-09
Update guide_prosody.rst

thisven:
[Documentation](url=https://prosody.im/doc/dns#srv_records) says:
"The target domain MUST be an existing A record of the target server, it cannot be an IP address, and cannot be a CNAME record." So target domain xmpp.isabell.org must be specified as A and AAAA record in DNS configuration.

I came across this when I investigated DNS configuration after executing `prosodyctl check`. Unfurtunately output is still like this:
```
Checking config...
Done.

Checking DNS for component conference.isabell.org...

Checking DNS for component upload.isabell.org...
    No _xmpp-server SRV record found for upload.isabell.org, but it looks like you need one.
    Host upload.isabell.org does not seem to resolve to this server (IPv4/IPv6)
    No targets for upload.isabell.org appear to resolve to this server.
    DNS records are necessary if you want users on other servers to access this component.

Checking DNS for host isabell.org...

For more information about DNS configuration please see https://prosody.im/doc/dns

[...]

Problems found, see above.
```


luto:
thanks! :)

PR #833, opened 2020-11-09
[TeamCity] Added guide for TeamCity

### LINKED ISSUE #828: Add guide for TeamCity
I've recently discovered TeamCity as a CI Runner and was able to set it up without any problems on Uberspace. I'd like to share my experiences and add a guide on how to setup a TeamCity server.

proohit:
Added the guide for TeamCity

closes #828

### DIFF
--- a/source/guide_teamcity.rst
+++ b/source/guide_teamcity.rst
@@ -28,6 +28,8 @@ TeamCity_ is a build management tool by JetBrains. It is not open source and req
 
   * :manual:`MySQL <database-mysql>`
   * :manual:`domains <web-domains>`
+  * :manual:`web backends <web-backends>`
+  * :manual:`supervisord <daemons-supervisord>`
 
 License
 =======
--- a/source/guide_teamcity.rst
+++ b/source/guide_teamcity.rst
@@ -14,9 +14,9 @@
       :target: https://www.jetbrains.com/company/brand/logos/
 
 
-##########
+########
 TeamCity
-##########
+########
 
 .. tag_list::
 
@@ -73,7 +73,7 @@ Installation
 ============
 
 Download and Extract
-------
+--------------------
 First, we're going to download TeamCity and extract it.
 
 Download the latest version (TeamCity-Download_ page) of the tgz archive and extract it. Note that the download link on the page is actually a redirect. So instead of the link behind the button, use the direct link e.g.: ``https://download.jetbrains.com/teamcity/TeamCity-2020.1.5.tar.gz``.
@@ -86,6 +86,7 @@ Download the latest version (TeamCity-Download_ page) of the tgz archive and ext
  [isabell@stardust ~]$
 
 Start Server
+------------
 We will configure two services for supervisor: One for the actual TeamCity server and one for the first build agent. You can use the following configuration file as a template but don't forget to change the paths! Save them in ``~/etc/services.d/``
 
 .. code-block:: text
@@ -107,12 +108,12 @@ We will configure two services for supervisor: One for the actual TeamCity serve
 
 .. include:: includes/supervisord.rst
 
-Configure Backend
-------
+Configure Web Backend
+---------------------
 .. include:: includes/web-backend.rst
 
 Initialize TeamCity Server
-------
+--------------------------
 We visit the page in the browser of our choice on isabell.uber.space.
 
 1. Type in the data directory. You can also use the default one at /home/isabell/.BuildServer
--- a/source/guide_teamcity.rst
+++ b/source/guide_teamcity.rst
@@ -86,23 +86,26 @@ Download the latest version (TeamCity-Download_ page) of the tgz archive and ext
  [isabell@stardust ~]$
 
 Start Server
-------
-Start the TeamCity Server. The startup scripts lie in TeamCity/bin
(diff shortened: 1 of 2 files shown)


source/guide_teamcity.rst
```diff
@@ -0,0 +1,162 @@
...
+
+Step 1
+------
+First, we're going to download TeamCity and extract it.
+
+Change into your local bin folder. Then we're going to download the latest version (TeamCity-Download_ page) of the tgz archive and extract it. Note that the download link on the page is actually a redirect. So instead of the link behind the button, use the direct link e.g.: ``https://download.jetbrains.com/teamcity/TeamCity-2020.1.5.tar.gz``.
+
+::
+
+ [isabell@stardust ~]$ cd ~/bin
```
luto, line 83: The bin-folder is for binaries only. Please extract teamcity into `~/TeamCity`.

source/guide_teamcity.rst
```diff
@@ -0,0 +1,162 @@
...
+Your URL needs to be setup:
+
+.. include:: includes/web-domain-list.rst
+
+You can also setup your own :manual_anchor:`domain <web-domains.html>`. TeamCity will run as a backend serving HTML files.
+
+Installation
+============
+
+Step 1
```
luto, line 75: Please name your steps, e.g. "Download and Extract", "Configuration", etc. This makes them much easier to find, in case you are only reading part of the guide, are continuing an earlier attempt or are debugging an existing installation. I am aware that older guides use numbered steps, but the has [been updated](https://github.com/Uberspace/lab/commit/9a55c0c79f520ecedd8964fb630e90903c192de) for a couple of months now.

source/guide_teamcity.rst
```diff
@@ -0,0 +1,162 @@
...
+ Agent home directory is /home/isabell/bin/TeamCity/buildAgent
+ Agent Launcher Java runtime version is 1.8
+ Lock file: /home/isabell/bin/TeamCity/buildAgent/logs/buildAgent.properties.lock
+ Using no lock
+ Done [28997], see log at /home/isabell/bin/TeamCity/buildAgent/logs/teamcity-agent.log
+ [isabell@stardust bin]$
+
+Step 3
+------
+Now we should configure a web backend for TeamCity.
```
luto, line 110: Please use the the [`web-backend.rst` include](https://github.com/Uberspace/lab/blob/master/source/includes/web-backend.rst) here.

source/guide_teamcity.rst
```diff
@@ -0,0 +1,162 @@
...
+ [isabell@stardust bin]$ rm teamcity.tgz
+ [isabell@stardust bin]$
+
+Step 2
+------
+Start the TeamCity Server. The startup scripts lie in TeamCity/bin
+
+::
+
+ [isabell@stardust bin]$ ./TeamCity/bin/runAll.sh start
```
luto, line 95: All services on uberspace should be started within supervisord. This ensures that they stay online, if a server gets a reboot or is having temporary problems.

luto (commented):
Thanks for the guide! I left a couple of comments.

I am not sure how to proceed here, because teamcity comes with a collection of scripts, which are supposed to manage the running services. As-is, the guide will not survive a reboot, resulting in a dead teamcity instance, once we restart our serves for updates.

Normally, services are managed using [supervisord](https://manual.uberspace.de/daemons-supervisord.html), which takes care of restarting broken services and starting them again after a server reboots. To do that, supervisord needs a process that stays in the foreground. Sadly, neither the `teamcity-server.sh` nor the `runAll.sh` script provide an option to do that.

So I'm afraid we've hit a dead end here, sorry :/

proohit:
> Thanks for the guide! I left a couple of comments.
> 
> I am not sure how to proceed here, because teamcity comes with a collection of scripts, which are supposed to manage the running services. As-is, the guide will not survive a reboot, resulting in a dead teamcity instance, once we restart our serves for updates.
> 
> Normally, services are managed using [supervisord](https://manual.uberspace.de/daemons-supervisord.html), which takes care of restarting broken services and starting them again after a server reboots. To do that, supervisord needs a process that stays in the foreground. Sadly, neither the `teamcity-server.sh` nor the `runAll.sh` script provide an option to do that.
> 
> So I'm afraid we've hit a dead end here, sorry :/

Thanks for the quick review!

As for the service problem: AFAIK you can start the server and agents independently, even in attached mode. Will this do eventually?
Server start: TeamCity/bin/teamcity-server.sh run
Agent start: TeamCity/buildAgent/bin/agent.sh run

If we write two service configurations, supervisor should be able to monitor them:
```
[program:build-agent]
command=bash /home/isabell/bin/TeamCity/buildAgent/bin/agent.sh run
autostart=yes
autorestart=yes
```
```
[program:teamcity-server]
command=bash /home/isabell/bin/TeamCity/bin/teamcity-server.sh run
autostart=yes
autorestart=yes
```

Although I've noticed that supervisor restarts the server quite often

luto:
Oh, I see! Those scripts actually seem to stay attached, yes. nice :) The process takes quite long to actually start up, so please set `startsecs=120` or so. Also, make sure the `supervisorctl stop` command is able to actually stop the service and set `stopsignal` and/or `stopasgroup` accordingly, if it isn't.

proohit:
I applied all the suggestions :)

luto:
Thank you! :tada:

PR #835, opened 2020-11-11
Update guide_prosody.rst

thomas-kammerer:
Adding SRV record for upload.isabell.org - according to `prosodyctl check`


luto:
Thanks!

PR #845, opened 2020-11-17
[coturn] add warning

nichtmax:


luto:
Move it down a bit, please: https://deploy-preview-845--uberlab.netlify.app/guide_coturn.html

thisven:
Can't we have this guide hosted anyway (with a warning) for those working with it? I was wondering why I can't find upgrade instructions anymore before I found this issue. What about the link that [luto mentioned above](https://github.com/Uberspace/lab/pull/845#issuecomment-728803753), will it stay available?

thisven:
I see. Many thanks for the static link. =)

PR #848, opened 2020-11-20
Add initial Part-DB Guide

927589452:
Add guide
Add image


source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+Installation
+============
+
+``cd`` to your :manual:`document root <web-documentroot>`, then download the latest release of Part-DB and extract it:
+
+.. note:: The link to the latest version can be found at the `release page <https://github.com/Part-DB/Part-DB/releases`_.
+
+::
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
```
luto, line 53: There is a symlink in the homedir, so this can be much shorter.
```suggestion
 [isabell@stardust ~]$ cd html/
```

source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                 Dload  Upload   Total   Spent    Left  Speed
+  100 37.5M  100 37.5M    0     0  5274k      0  0:00:07  0:00:07 --:--:-- 6062k
+ [isabell@stardust html]$
+
+To install all needed libraries you run `composer install`Ö
+
+::
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
```
luto, line 64: Since we're just continuing the last step, no need to cd again.
```suggestion
```

source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+ [isabell@stardust html]$ curl --location https://github.com/Part-DB/Part-DB/archive/master.tar.gz | tar -xzf - --strip-components=1
+  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                 Dload  Upload   Total   Spent    Left  Speed
+  100 37.5M  100 37.5M    0     0  5274k      0  0:00:07  0:00:07 --:--:-- 6062k
+ [isabell@stardust html]$
+
+To install all needed libraries you run `composer install`Ö
```
luto, line 60: typo
```suggestion
To install all needed libraries you run `composer install`:
```

source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+::
+
+ [isabell@stardust ~]$ mysqldump ${USER}_partdb > partdb.sql
+ [isabell@stardust ~]$
+
+Remove the old installation:
+
+::
+
+ [isabell@stardust ~]$ rm -rf ~/html/
```
luto, line 141: This works, but is tricky. If you omit the trailing slash, the `rm -rf` deletes the `~/html` symlink only but leaves the actual directory as-is. Please add a `.. note::` box above this to tell users that the trailing slash is necessary here.

source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+ [isabell@stardust ~]$
+
+Follow the installation procedure and restore the data directory
+::
+
+ [isabell@stardust ~]$ mv -Z ~/data ~/html/data
```
luto, line 147: our `mv` is an `mv -Z` by default:
```
[user@halimede ~]$ alias mv
alias mv='mv -Z'
```
```suggestion
 [isabell@stardust ~]$ mv ~/data ~/html/data
```

source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+The installation wizard will prompt you to initialize the database and you will have to change the admin password afterwards.
+
+Updates
+=======
+
+To update Part-DB you copy the data folder to your home directory:
+
+::
+
+ [isabell@stardust ~]$ cp --recursive ~/html/data ~/
```
luto, line 127: After we copy this directory, the source is deleted anyway. What do you think about changing this to a `mv` instead of `cp`? will be faster, too :)
```suggestion
 [isabell@stardust ~]$ mv ~/html/data ~/
```

source/guide_part-db.rst
```diff
@@ -0,0 +1,161 @@
...
+ [isabell@stardust ~]$
+
+Create an backup of your database:
+
+::
+
+ [isabell@stardust ~]$ mysqldump ${USER}_partdb > partdb.sql
```
luto, line 134:
```suggestion
 [isabell@stardust ~]$ mysqldump ${USER}_partdb > partdb_update_backup.sql
```

luto (changes requested):
nice and short :) I left a few comments to streamline the process a bit. Thanks for the guide!

source/guide_part-db.rst
927589452: I would personally keep it as a backup 

And I am thinking about splitting the whole backup part into an own section

927589452: Oh wait, it's the other direction

source/guide_part-db.rst
927589452: Would it be better to work directly with the full `/var` based path here? no chance to delete the symlink then. 

Or alternatively use the git version and tags; this would remove the old files automatically AFAIK
Or would it be a bad idea to replace the document root html directory entirely with the git repository?

source/guide_part-db.rst
927589452: I wasn't sure, if the symlink would get the same SE_Linux mode and basically copied this step from the nextcloud guide

source/guide_part-db.rst
luto: Sure, specify the whole `/var/...` path here.

source/guide_part-db.rst
```diff
@@ -0,0 +1,198 @@
...
+
+----
+
+Tested with Part-DB 0.5.9, Uberspace 7.7.10.0
+
+.. author_list::
+
+Sources:
+    * :lab: guide_satis
+    * :lab: guide_nextcloud
```
luto, line 198: image
This doesn't really work as expected and looks a bit funny. Structure and fragments are copied freely inside the lab all the time - feel free to just skip this section.

source/guide_part-db.rst
```diff
@@ -0,0 +1,198 @@
...
+
+::
+
+ [isabell@stardust ~]$ mysqldump ${USER}_partdb > partdb_update_backup.sql
+ [isabell@stardust ~]$
+
+Remove the old installation:
+
+.. note::
+   If the trailing `/` is missing from the command below, the folder will be deleted and you will need to restore it using `mkdir /var/www/virtual/${USER}/html`.
```
luto, line 173: That's not true. The `html` directory will be deleted with or without slash.
```
[testt@stardust ~]$ ll /var/www/virtual/${USER}/html
-rwxr-xr-x. 1 testt testt 1171 Nov 22 17:41 nocontent.html

[testt@stardust ~]$ rm -rf /var/www/virtual/${USER}/html/
[testt@stardust ~]$ ll /var/www/virtual/${USER}/html
ls: cannot access /var/www/virtual/testt/html: No such file or directory

[testt@stardust ~]$ mkdir /var/www/virtual/${USER}/html

[testt@stardust ~]$ rm -rf /var/www/virtual/${USER}/html
[testt@stardust ~]$ ll /var/www/virtual/${USER}/html
ls: cannot access /var/www/virtual/testt/html: No such file or directory
```
Please have a look at the suggestion below and remove this note.

source/guide_part-db.rst
```diff
@@ -0,0 +1,198 @@
...
+
+::
+
+ [isabell@stardust ~]$ rm -rf /var/www/virtual/${USER}/html/
```
luto, line 177: To fix the comment above.
```suggestion
 [isabell@stardust ~]$ rm -rf /var/www/virtual/${USER}/html/
 [isabell@stardust ~]$ mkdir /var/www/virtual/${USER}/html/
```

source/guide_part-db.rst
```diff
@@ -0,0 +1,198 @@
...
+
+ [isabell@stardust ~]$ mv ~/html/data ~/
+ [isabell@stardust ~]$
+
+Create an backup of your database:
+
+::
+
+ [isabell@stardust ~]$ mysqldump ${USER}_partdb > partdb_update_backup.sql
+ [isabell@stardust ~]$
```
luto, line 168: The entire backup process is duplicated here. What do you think about just referring to the section above?
```suggestion
To update Part-DB you create a backup as described above.
```

luto (changes requested):
Thanks for the fixes! I added a couple comments regarding the new sections :)

source/guide_part-db.rst
927589452: `mv` was quicker (as of your first suggestion) but yeah removing it is cleaner

luto:
woop! :tada: thank you!

I can't find your mail address. Please ping hallo@uberspace.de to claim reward :)

PR #853, opened 2020-11-22
[privatebin] add guide for privatebin

### LINKED ISSUE #852: add guide for PrivateBin
https://privatebin.info/
https://github.com/PrivateBin/PrivateBin

First-Timer here, feedback appreciated.

Nepomacs:
Add guide and image. Closes #852 

First Timer here. Feedback appreciated.


source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Installation
+============
+
+Download the source
+-------------------
+
+Check Github_ for the `latest release`_ and copy the download link to the .tar.gz file.
+Then ``cd`` to your :manual:`DocumentRoot <web-documentroot>` and use ``wget`` to download it. Replace the URL with the one you just copied.
+In the example the version is 1.2.4 which does not exist.
```
luto, line 64: We had fake versions for the longest time, but it just ended up causing more harm than joy. We've updated the style guide since to say

> If possible use a download URL that points to the latest version (e.g. latest.zip on some platforms). If such an URL is not available, use the newest version instead.

Please update the guide to point to version 1.3.4.

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+
+.. code-block:: console
+ :emphasize-lines: 2
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/
```
luto, line 69: This seem a bit convoluted. First we `cd` into `html/..` to then just extract into `html` using `-C html/`. What do you think about changing this to `cd /var/www/virtual/$USER/html` (or rather `cd ~/html`, which is much shorter) and skipping the `-C`?

Please make sure to also update the prompts for the following lines to say `html]$` instead of `isabell]$`.

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+ [isabell@stardust isabell]$ wget https://github.com/PrivateBin/PrivateBin/archive/1.2.4.tar.gz
+ […]
+ Saving to: ‘1.2.4.tar.gz’
+
+ 100%[=================================================>] 3,172,029   3.45MB/s   in 0.9s
+
+ 2020-11-22 16:27:44 (8.32 MB/s) - ‘1.2.4.tar.gz’ saved [523648]
+ [isabell@stardust ~]$
```
luto, line 77: Since there isn't another `cd` here, the prompt stays the same and doesn't revert back to `~`.

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+
+
+Activate the .htaccess file
+-----------------------------
+PrivateBin provides a .htaccess file, which blocks some known robots and link-scanning bots. Activate it by renaming it from ``.htaccess.disabled`` to ``.htaccess``.
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ cd /var/www/virtual/$USER/html
```
luto, line 95: If you accept my previous suggestion, we're already in `html` and you can skip this line entirely.
```suggestion
```

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+ [isabell@stardust html]$ mv .htaccess.disabled .htaccess
+ [isabell@stardust ~]$
```
luto, line 97: prompt stays the same, like above.

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+Moving files outside of DocumentRoot
+------------------------------------
+It is recommended_ to move the configuration, data files, templates and PHP libraries outside of your document root.
+To do that, create a Folder  ``privatebin-data`` in ``/var/www/virtual/isabell/`` and move the folders to the new location.
+If not already there, go to the ``html`` directory before running ``mv``.
+
+.. code-block:: console
+
+ [isabell@stardust html]$ mkdir /var/www/virtual/$USER/privatebin-data
+ [isabell@stardust html]$ cd /var/www/virtual/$USER/html
```
luto, line 109: the prompt and prior steps indicate that we're already in `html`. No need to change the directory here.
```suggestion
```

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Moving files outside of DocumentRoot
+------------------------------------
+It is recommended_ to move the configuration, data files, templates and PHP libraries outside of your document root.
+To do that, create a Folder  ``privatebin-data`` in ``/var/www/virtual/isabell/`` and move the folders to the new location.
+If not already there, go to the ``html`` directory before running ``mv``.
+
+.. code-block:: console
+
+ [isabell@stardust html]$ mkdir /var/www/virtual/$USER/privatebin-data
```
luto, line 108: What do you think about using a relative path to keep the command short?
```suggestion
 [isabell@stardust html]$ mkdir ../privatebin-data
```

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+ [isabell@stardust html]$ cd /var/www/virtual/$USER/html
+ [isabell@stardust html]$ mv -t /var/www/virtual/$USER/privatebin-data cfg/ lib/ tpl/ vendor/
```
luto, line 110: same as above, relative?
```suggestion
 [isabell@stardust html]$ mv -t ../privatebin-data cfg/ lib/ tpl/ vendor/
```

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+ [isabell@stardust ~]$
```
luto, line 111: prompt stays the same :) just adding a note to each so I keep track of them ;)

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+Now edit ``/var/www/virtual/$USER/html/index.php``  to inform PrivateBin about to the new location of the folders. 
+
+.. warning:: Replace ``<username>`` with your username!
+
+.. code-block:: php
+ :emphasize-lines: 3
+
+ [...]
+ // change this, if your php files and data is outside of your webservers document root
+ define('PATH', '/var/www/virtual/<username>/privatebin-data/');
```
luto, line 125: Maybe this also works with a relative path like above? This would remove the additional step of replacing `<username>`.

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+ [...]
+
+Setting permissions
```
luto, line 128: The permissions look fine without this step. Did you encounter permission problems while creating the guide?

source/guide_privatebin.rst
```diff
@@ -0,0 +1,204 @@
...
+
+
+
+Configuration
+=============
+
+Configure your PrivateBin Instance
+----------------------------------
+
+.. note:: You are not forced to change any of the default settings as they are mostly secure.
```
luto, line 149: "forced" is a bit of a strong word :sweat_smile: 
```suggestion
.. note:: You don't need to change any of the default settings as they are mostly secure.
```

luto (changes requested):
Awesome, thank you :tada: I added some comments. Mostly tiny changes to make the guide more streamlined. Let me know what you think! Once those are resolved, I'll take the guide for a spin myself.

source/guide_privatebin.rst
Nepomacs: The style-guide did confuse me here as it says "Always end your console code blocks with [isabell@stardust ~]$."  :-)

source/guide_privatebin.rst
Nepomacs: it works 👍 Nice!

source/guide_privatebin.rst
Nepomacs: There were no problems but the privatebin-data folder has 775 instead of 755 and the files 664 instead of 644. Is that a problem? I guess not, so I removed the chapter.

source/guide_privatebin.rst
Nepomacs: I agree :-)

Nepomacs:
Thanks for reviewing my changes.

luto:
https://pri.uber.space/?ee511327f2a5ac1f#3ZULxxj6Qmkn7fNSJVLL1hqFnWTWrYd3fxaFJNGrHYxj

luto:
Thanks for the guide! :tada: 

I can't find an email address on your github profile. Please ping us at hallo@uberspace.de to claim your reward :)

PR #864, opened 2020-12-01
[etesync-server] add warning about v2 incompatibility

Nepoh:
Etesync Server version [0.5.0](https://github.com/etesync/server/releases/tag/v0.5.0) seems to introduce some major changes. It looks like this guide no longer works for the latest releases. So I removed all the parts that promote using a version after 0.3.0 until the guide gets updated.


luto:
thanks!

PR #870, opened 2020-12-08
[bitwarden_rs] Added create data folder instructions & note to web-vault

### LINKED ISSUE #843: [bitwarden_rs] data folder and clearer instructions for web vault
I followed the guide but I had to add the `bitwarden_rs/data` folder manualy. Otherwise the bitwarden server crashed with "error creating keys". Maybe this step should be added to the guide?

Another thing I struggled with was the following note regarding the web vault:

> If you don't want to use the web-vault feature (web-app access to your vaults) for any security reasons you can skip this step.

It should be more clear that without the web app an invited user can't join because the link sent with the invitation mail leads to a 404 site. Or is there any other way to add a user?

thurti:
fixes #843

### DIFF
--- a/source/guide_bitwarden.rst
+++ b/source/guide_bitwarden.rst
@@ -83,7 +83,8 @@ In order to build bitwarden_rs  successfully you'll need to set an environment v
 
 .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


luto:
oh, thanks! this bit us in support once or twice already.

PR #875, opened 2020-12-10
[adminer] add guide for adminer

### LINKED ISSUE #874: Guide for Adminer
Hey everybody!

I know there are instances of [Adminer](https://mysql.uberspace.de/adminer/) and [phpMyAdmin](https://mysql.uberspace.de/phpmyadmin/) available globally, but still I decided to run my own version of Adminer. Mainly to apply a custom theme :D
Just in case I'm not alone with this I will create a guide.

nikolasdas:
Hey!

Here is my guide for Adminer :)

Fixes #874


luto:
Thanks for you submission! I'm sorry, but I won't find time to review it until mid-January :/

source/guide_adminer.rst
```diff
@@ -0,0 +1,140 @@
...
+  [isabell@stardust ~]$ VERSION=4.7.8
+  [isabell@stardust ~]$ cd ...
+  [isabell@stardust ~]$
+
+Option 1: Install Adminer without Plugin Support
+------------------------------------------------
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ wget -O index.php https://github.com/vrana/adminer/releases/download/v$VERSION/adminer-$VERSION.php
```
luto, line 53: This will install adminer into the home directory, instead of the [document root](https://manual.uberspace.de/web-documentroot/). Maybe `-O html/index.php`?

source/guide_adminer.rst
```diff
@@ -0,0 +1,140 @@
...
+  [isabell@stardust ~]$
+
+That's it!
+
+Option 2: Install Adminer with Plugin Support
+---------------------------------------------
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ wget -O adminer.php https://github.com/vrana/adminer/releases/download/v$VERSION/adminer-$VERSION.php
```
luto, line 63: same as above, PHP files need to go into the document root.

source/guide_adminer.rst
```diff
@@ -0,0 +1,140 @@
...
+
+You can find a list of available Plugins `here <https://www.adminer.org/plugins/>`_.
+
+Example Plugin: Themes
+----------------------
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ THEME_VERSION=1.7
+  [isabell@stardust ~]$ wget -O theme.zip https://github.com/pematon/adminer-theme/archive/v$THEME_VERSION.zip
```
luto, line 98: same as above, PHP files need to go into the document root.

source/guide_adminer.rst
```diff
@@ -0,0 +1,140 @@
...
+    return new AdminerPlugin($plugins);
+  }
+
+  include "./adminer.php";
+
+This will include all plugins located in the ``plugins`` folder. To use a plugin you also need to initialize it inside the ``$plugins`` array.
+
+You can find a list of available Plugins `here <https://www.adminer.org/plugins/>`_.
+
+Example Plugin: Themes
```
luto, line 92: What do you think of only providing this section, skipping the theoretical "Install Adminer with Plugin Support" one and the "Install Adminer without Plugin Support" one? Users, who don't need plugins, can use the adminer provided by us. Users can adapt the theme example to their needs. It would also keep the guide nice and short, without any duplicated code or commands.

luto (changes requested):
Thanks for the guide! I am sure a custom adminer is a nice addition for users, who would like to install additional plugins.

source/guide_adminer.rst
nikolasdas: I changed the code to cd into the html directory first

source/guide_adminer.rst
nikolasdas: Valid point. I removed the part without plugins

luto:
Thank you! :tada:

PR #880, opened 2020-12-12
[nextcloud] disable backup in update script

### LINKED ISSUE #851: [Nextcloud] Disable backups before update via  ~/bin/nextcloud-update script
There is already an update script and a workaround for storage capacity problems in the uberspace documentation. 
I think it would be best to disable the automatic backups that the nextcloud updater generates as they are really not needed on the uberspace and take up unnecessary time and resources. 

The updater provides no flag to disable automatic backups, so the only workaround at the moment seems to edit the nextcloud/updater/index.php. 

Three lines need to be commented out:
//	case 3:
//		$updater->createBackup();
//		break;

Cmp.: https://github.com/nextcloud/updater/issues/145#issuecomment-612727675

I think this could be handled by issuing a sed command during the nextcloud-update script on uberspace.

EV21:
closes #851 
also add new database optimisation command (NC20)

- [x] tested

deploy preview: [guide_nextcloud.html#updates](https://deploy-preview-880--uberlab.netlify.app/guide_nextcloud.html#updates)

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -62,7 +62,7 @@ Installation
 
 ::
 
- [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
+ [isabell@stardust ~]$ cd html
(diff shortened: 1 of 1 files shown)


luto:
thanks!

PR #882, opened 2020-12-15
fix nodejs tags

EV21:
Some tag variants have been renamed in favor of the [popular software link for nodejs in the manual](https://manual.uberspace.de/lang-nodejs/#popular-software).


luto:
Thanks!

PR #885, opened 2020-12-18
Create guide_django_venv.rst

Spidiffpaffpuff:
A guide to settign up two instances of Django, each in their own virtual environment, on uberspace.


luto:
I am sorry, but we can't include this guide. Please open an issue before creating a guide, as outlined in [CONTRIBUTING.md](https://github.com/Uberspace/lab/blob/master/CONTRIBUTING.md) to prevent this in the future.

Each guide should assume that it occupies its uberspace alone, or rather, alone together with its dependencies. We recommend using one uberspace per project to keep things clean. This also eliminates the need for a venv.

Spidiffpaffpuff:
Hey, sorry for the inconvenience. I am a total github noob. I just wanted to write this guide on my own account and wasn't even aware that I had already done this in a way where you would notice. :)
So, are you suggesting that if I have 3 projects, I just open 3 uberspace accounts?

luto:
yes! you can split your desired price as needed.

PR #886, opened 2020-12-19
Fix ejabberd installation path and compilation command

coderkun:
Align the installation path in the ejabberd guide with the description. Fix the compilation to run “make” before “make install” which is required for existing installations.


luto:
Thanks! :)

PR #892, opened 2020-12-30
[ghost] Remove line 'Do you want to start Ghost?: Answer No.'

pbru87:
The question "Do you want to start Ghost?" does not show up at the console (and we already use the '--no-start' parameter at ghost install) Therefore, we can remove the mentioned line from the doc.


luto:
Thanks! :tada:

PR #894, opened 2020-12-30
[ghost] Updated section 'Prerequisites'

pbru87:
We should update the mentioned Node.js version from v10 to v12, because v12 is the new Uberspace default (see https://manual.uberspace.de/lang-nodejs/) and it is also the recommended version of Ghost.org (see https://ghost.org/docs/faq/node-versions/).


luto:
Thank you!

PR #901, opened 2021-01-06
[wsgidav] Fixed host ip address

nrekretep:
Explicitly set host ip address to 0.0.0.0

For Wsgidav to be usable as a web backend it needs to listen on 0.0.0.0:8080 and not localhost:8080, which is default.

Tested with wsgidav 3.1.0, Uberspace 7.8.1.0


luto:
Thanks! :)

PR #905, opened 2021-01-12
[wordpress] Replace explicit mention of isabell by $USER

MaxG87:
Using $USER in two places decreases the number of location the user of
the manual has to adapt to his needs.


MaxG87:
I edited this merge request because the original attempt became obsolete. I noted that the manual already includes a command to create a new MYSQL database, so my new paragraph in the manual is not needed.

luto:
Thanks!

PR #912, opened 2021-01-25
[Invoice Ninja] updated to use v4.5.31

### LINKED ISSUE #899: Invoice Ninja Guide not working for Updates
The update guide results in an error.
https://lab.uberspace.de/guide_invoiceninja.html#updates

...invoiceninja]$ composer install
[03-Jan-2021 11:18:00 Europe/Berlin] PHP Fatal error:  Class UpdateHelper\ComposerPlugin contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (Composer\Plugin\PluginInterface::deactivate, Composer\Plugin\PluginInterface::uninstall) in /var/www/virtual/YYY/invoiceninja/vendor/kylekatarnls/update-helper/src/UpdateHelper/ComposerPlugin.php on line 11

Fatal error: Class UpdateHelper\ComposerPlugin contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (Composer\Plugin\PluginInterface::deactivate, Composer\Plugin\PluginInterface::uninstall) in /var/www/virtual/YYY/invoiceninja/vendor/kylekatarnls/update-helper/src/UpdateHelper/ComposerPlugin.php on line 11

@tilldeeke

DaniW42:
- got rid of deprecated php 7.1
- composer is not working so we simply use the zip build
- for convenience we are using the default html link in the users directory
- changed to use the pre compiled phantomjs binary (https://twitter.com/ubernauten/status/325582191822249985)
- since we are using the zip build we can now use the internal updater
- should also close #899

@luto (hallo@uberspace.de mail 4 Dec 2020, 22:06) Coffee mug incoming (Please)? ;-)

### DIFF
--- a/source/guide_invoiceninja.rst
+++ b/source/guide_invoiceninja.rst
@@ -127,8 +127,8 @@ To automatically send recurring invoices and reminders we need to setup some cro
 
 ::
 
- 0 8 * * * /usr/bin/php /var/www/virtual/$USER/artisan ninja:send-invoices
- 0 8 * * * /usr/bin/php /var/www/virtual/$USER/artisan ninja:send-reminders
+ 0 8 * * * /usr/bin/php /var/www/virtual/$USER/html/artisan ninja:send-invoices
+ 0 8 * * * /usr/bin/php /var/www/virtual/$USER/html/artisan ninja:send-reminders
 
 You can learn more about cronjobs in the :manual:`uberspace manual cron article <daemons-cron>`.
 
--- a/source/guide_invoiceninja.rst
+++ b/source/guide_invoiceninja.rst
@@ -164,7 +164,7 @@ After that you can :manual_anchor:`add a new supervisord service <daemons-superv
 
 ::
 
- /usr/bin/php /var/www/virtual/$USER/artisan queue:work --daemon
+ /usr/bin/php /var/www/virtual/$USER/html/artisan queue:work --daemon
 
 as the command.
 
--- a/source/guide_invoiceninja.rst
+++ b/source/guide_invoiceninja.rst
@@ -1,6 +1,7 @@
 .. highlight:: console
 
 .. author:: Till Deeke <hallo@tilldeeke.de>
+.. author:: Daniel Weber <https://github.com/DaniW42>
 
 .. tag:: lang-php
 .. tag:: accounting
@@ -39,14 +40,6 @@ License
 Prerequisites
 =============
 
-We're using :manual:`PHP <lang-php>` in the stable version 7.1:
-
-::
-
- [isabell@stardust ~]$ uberspace tools version show php
- Using 'PHP' version: '7.1'
- [isabell@stardust ~]$
-
 .. include:: includes/my-print-defaults.rst
 
 Your domain needs to be set up:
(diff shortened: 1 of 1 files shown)


source/guide_invoiceninja.rst
```diff
@@ -136,8 +127,8 @@ To automatically send recurring invoices and reminders we need to setup some cro
 
 ::
 
- 0 8 * * * /usr/bin/php /var/www/virtual/$USER/invoiceninja/artisan ninja:send-invoices
```
luto, line 139: Are you sure about that? The prevous `[isabell@stardust ~]$ mv ninja/* html/` step installs invoice ninja into `/var/www/virtual/$USER/html`. Where is the `artisan` script supposed to come from?

luto (changes requested):
Thanks for the guide! Just one kind of fundamental question. Once this is resolved, I'll take a closer look and take the new version for a spin.

source/guide_invoiceninja.rst
DaniW42: Mea culpa, I forgot the html/. Will change that

DaniW42: I double checked by doing the whole thing on ninjatmp@gale.uber.space. Feel free to have a look. Will remove the account after successful merge

source/guide_invoiceninja.rst
```diff
@@ -157,22 +164,27 @@ After that you can :manual_anchor:`add a new supervisord service <daemons-superv
 
 ::
 
- /usr/bin/php /var/www/virtual/$USER/invoiceninja/artisan queue:work --daemon
+ /usr/bin/php /var/www/virtual/$USER/artisan queue:work --daemon
```
luto, line 167:
```suggestion
 /usr/bin/php /var/www/virtual/$USER/html/artisan queue:work --daemon
```
based on the cronjob fix, this should read as above?

DaniW42: absolutely yes

DaniW42 (commented):
@luto did a great job fixing my updated code from @tilldeeke. Looks good to me even though it now contains the weird php downgrading and updating stuff. without that php flipping it unfortunately ends in an error 500 by the webserver. any help is greatly appreciated. in the end it might be not worth the effort. invoice ninja runs under php 7.3 too but will soon face an EOL why we might consider reviewing this in the future.

DaniW42 (commented):
You were absolutely right. I still don't really get it with the html directory in /home and /var/www. But I changed any occurrence of this.

Edit:
I tried to review that open change request but GitHub still asks me to do it. Sorry, I can not find where to do this...

luto:
>  I still don't really get it with the html directory in /home and /var/www

The actual document root is `/var/www/virtual/$USER/html`. To make your life easier (or harder, in this case), there is a symlink from `/home/$USER/html` to that directory.

tilldeeke:
Hi, I just noticed the notification for this PR and had a quick look at it. Thank you for updating the guide! There are two things I noticed:

- The internal updater is only available in the new v5 version of Invoice Ninja. The old v4 of this guide does not have the ability to update itself, you need to manually follow the steps outlined in the original guide. 

- Is there a reason you removed the symlink step in the installation? The /public directory should be the root directory for the served site. It is designed this way, so sensitive files like the .env file are never exposed to the web, since the only directly accessible files are in the /public directory. With the updated instructions it might be possible to access sensitive files in the application itself.

And another small thing: Isn't Uberspace already enforcing the usage of HTTPS, so the redirect in the .htaccess file isn't needed?

source/guide_invoiceninja.rst
```diff
@@ -126,6 +125,14 @@ User Details
 
 After you save the settings, you will be redirected to the login for the admin panel.
 
+We now can change back to PHP Version 7.4:
```
luto, line 128: Can you point me to some documentation, which says that 7.4 is supported? The [official docs](https://docs.invoiceninja.com/install.html) say only <= 7.2 is supported. Some github issues (https://github.com/invoiceninja/invoiceninja/issues/4221#issuecomment-717618019, https://github.com/invoiceninja/invoiceninja/issues/3503#issuecomment-599262500) claim support for 7.3, but explicitly rule out support for 7.4. Additionally, 7.4 seems to cause errors similar to the ones mentioned in the linked issues:
```
[16-Feb-2021 18:51:28] WARNING: [pool www] child 12282 said into stderr: "[16-Feb-2021 17:51:28 UTC] PHP Deprecated:  Unparenthesized `a ? b : c ?: d` is deprecated. Use either `(a ? b : c) ?: d` or `a ? b : (c ?: d)` in /var/www/virtual/invninja/html/app/Models/Account.php on line 1126"
```
What do you think about just sticking to 7.3 for now?

Jaglag: Even with 7.3 it has some problems.
I have it running here with 7.3 and my interface is broken, and I can't view the already generated invoices.
@DaniW42 4.5.31 isn't supported and don't receive any update in the future, so it would probably just better to focus on v5 because they are already in r2 and people use it in Production.
I tried to install v5 but the composer throwing out an error:
Script vendor/bin/snappdf download handling the post-install-cmd event returned with error code 126

luto:
Considering that invoice ninja v4 does not work with recent PHP versions, I guess I'll have to close this. We'd be happy about a v5 guide, though!

PR #913, opened 2021-01-25
[pleroma] update guide

taddydevil:
*Small* typo in config that caused the value not to be accepted
Update postgresql setup

### DIFF
--- a/source/guide_pleroma.rst
+++ b/source/guide_pleroma.rst
@@ -50,13 +50,6 @@ We’re using :manual:`PostgreSQL <database-postgresql>` in the version 13. Plea
  Using 'postgresql' version: 13
  [isabell@stardust ~]$
 
-Please make sure you set ``listen_addresses`` in the config.
-
-.. code-block:: console
- :emphasize-lines: 1
-
- listen_addresses = '*'         # what IP address(es) to listen on;
-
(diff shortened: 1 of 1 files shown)


source/guide_pleroma.rst
```diff
@@ -42,13 +42,21 @@ Your URL needs to be setup:
...
+ [isabell@stardust ~]$ uberspace tools version show postgresql
+ Using 'postgresql' version: 13
  [isabell@stardust ~]$
 
+Please make sure you set ``listen_addresses`` in the config.
+
+.. code-block:: console
+ :emphasize-lines: 1
+
+ listen_addresses = '*'         # what IP address(es) to listen on;
```
luto, line 58: isn't [the default value localhost](https://www.postgresql.org/docs/13/runtime-config-connection.html#GUC-LISTEN-ADDRESSES) enough here since we only care about access within the uberspace?

luto (changes requested):
Thank you!

source/guide_pleroma.rst
taddydevil: I tested it and it worked with ``#listen_addresses = 'localhost'`` too. PSQL still listen on localhost with the default config.
Thought psql dont listen on any interface when this line is comment out by a ``#``.

luto:
Thanks! :tada:

PR #922, opened 2021-02-02
[mongodb] fix typo in version

taddydevil:
MongoDB version 12 does not exist.
Adapted ``uberspace tools version use mongodb`` accordingly with the entry in the manual.


luto:
whoops! that's copy-pasta from the pgsql guide. sorry m)

PR #923, opened 2021-02-02
bitwarden_rs - Updated versions and corrected one Command

m0veax:
Node.js 13 is not available anymore. Further i updated the webinterface version and fixed an uncomplete command.


luto:
sweet! thanks :)

PR #924, opened 2021-02-03
Update guide_passbolt.rst

momo-aux:
since composer 2 is not compatible downloading and using composer1 is needed to install


luto:
thank you! :tada:

PR #925, opened 2021-02-04
add tutorial for beehive

momo-aux:
Added a tutorial to install Mueslis beehive


source/guide_beehive.rst
```diff
@@ -0,0 +1,124 @@
...
+
+  * :manual:`Go <lang-go>`
+  * :manual:`Web Backends <web-backends>`
+  * :manual:`Supervisord <daemons-supervisord>`
+  * :manual:`Domains <web-domains>`
+
+License
+=======
+
+Beehive is released under the `AGPL-3.0 License`_.
```
luto, line 41: This link [does not seem to work](https://deploy-preview-925--uberlab.netlify.app/guide_beehive.html#license). Maybe because there is an extra ` ` in front of `AGPL` in the link list at the bottom?

source/guide_beehive.rst
```diff
@@ -0,0 +1,124 @@
...
+Install ``Beehive``:
+
+.. code-block:: console
+
+  [isabell@stardust ~]$ cd
+  [isabell@stardust ~]$ mkdir beehive
+  [isabell@stardust ~]$ wget https://github.com/muesli/beehive/releases/download/v0.4.0/beehive_0.4.0_Linux_x86_64.tar.gz
+  [isabell@stardust ~]$ cd beehive
+  [isabell@stardust ~]$ tar -xzvf beehive_0.4.0_Linux_x86_64.tar.gz
+  [isabell@stardust ~]$ rm beehive_0.4.0_Linux_x86_64.tar.gz
```
luto, line 64: It seems like there is some confusion about the current directory here :sweat_smile: the `tar -xzvf` won't find the archive file after we `cd beehive`. How about this?
```suggestion
  [isabell@stardust ~]$ mkdir beehive
  [isabell@stardust ~]$ cd beehive
  [isabell@stardust beehive]$ wget https://github.com/muesli/beehive/releases/download/v0.4.0/beehive_0.4.0_Linux_x86_64.tar.gz
  [isabell@stardust beehive]$ tar -xzvf beehive_0.4.0_Linux_x86_64.tar.gz
  [isabell@stardust beehive]$ rm beehive_0.4.0_Linux_x86_64.tar.gz
```

luto: Actually, after looking at the contents of the tar.. what do you think about this? It would also simplify the supervisord config to just `command=beehive` since the binary sends up in `~/bin` where supervisord looks for it.
```suggestion
  [isabell@stardust ~]$ wget https://github.com/muesli/beehive/releases/download/v0.4.0/beehive_0.4.0_Linux_x86_64.tar.gz
  [isabell@stardust ~]$ tar -C ~/bin -xf beehive_0.4.0_Linux_x86_64.tar.gz beehive
  [isabell@stardust ~]$ rm beehive_0.4.0_Linux_x86_64.tar.gz
```

source/guide_beehive.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Configuration
+=============
+
+Beehive will generate default config on startup
+
+Web Backend Config
+------------------
+
+.. warning:: Beehive_ will bind to locahost per default. There is no login. Please make sure to secure it before using it with external interface.
```
luto, line 75: How do you secure it? All lab guides should produce a at least reasonably secure setup, if you follow all of the instructions.

source/guide_beehive.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Prerequisites
+=============
+
+Nothing special
+.. warning:: But be careful, Beehive has no login and defaults to bind to localhost. So to connect you should use ssh tunnel or use any kind of .htaccess Basic Auth or similar.
+
+::
+
+
```
luto, line 51: If there are no prerequisites, skip the section, please. The security is handled down below.

```suggestion
```

luto (changes requested):
Thanks for the guide! it's nice and short :) Do you mind having a look at the comments?

source/guide_beehive.rst
momo-aux: It is secure because it binds to localhost. So no external connection possible but if started with -bind "0.0.0.0:8181" then you need to take care of access. is there a reverse proxy guide somewhere that I could link?

I don't know how to add htaccess on uberspace if directly binding web access to the service. If you can point me to the right direction, I'll get into it and update but I could not find something about securing services.

source/guide_beehive.rst
momo-aux: Hope i updated the links right. Can I check the preview without bothering you everytime when I update? could not find something here: https://github.com/Uberspace/lab/blob/main/STYLE.md

momo-aux (commented):
Hope I adressed all issues. Thanx for helping me to commit.

source/guide_beehive.rst
luto: Right now there is no way to add basic authentication to a web backend, if the app itself does not support it. If beehive has no authentication built-in, I guess the only way to run it securely is to only use it inside of SSH port forwarding.

momo-aux: should I just remove the bind and add a little tutorial for Chrome socks proxy?

momo-aux: or maybe just link here: https://www.chromium.org/developers/design-documents/network-stack/socks-proxy
and here https://securitymusings.com/article/462/how-to-set-up-a-socks-proxy-using-putty-ssh

momo-aux (commented):
If wanted, I can add additional Information for "Tunnel with Chromium" or putty but I think thats something not directly for this tutorial, right?

source/guide_beehive.rst
luto: hm, I removed the sections that expose beehive to the internet. Still a helpful guide nonetheless! :)

luto:
Thanks for the guide and sorry for the wait!

PR #928, opened 2021-02-10
Added Troubleshooting for "output_buffering" error

TalusUnheil:
Source: https://help.nextcloud.com/t/php-configuration-option-output-buffering-must-be-disabled-nextcloud-20-0-1-ubuntu/97918

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -443,7 +443,21 @@ Here is an example you probably don't want to keep on your Uberspace. To get rid
 .. _`Nextcloud App Store`: https://apps.nextcloud.com
 .. _SELinux labels: https://wiki.gentoo.org/wiki/SELinux/Labels#Introduction
 
+Fehlermeldung "PHP configuration option output_buffering must be disabled" beheben
+----------------------------------------------------------------------------------
 
+Wenn nach einer Aktualisierung der Nextcloud der Fehler "PHP configuration option output_buffering must be disabled" gemeldet wird, dann muss zunächst die zur Behebung erforderliche PHP-Konfiguration mit folgender Kommandozeile ergänzt werden:
+
+.. code-block:: console
+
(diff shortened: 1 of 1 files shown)


EV21:
Hi @TalusUnheil ,
thank you for your pull request.

Output-buffering is already mentioned as an initial step in this guide.
see https://lab.uberspace.de/guide_nextcloud.html#output-buffering
If you think it is important (for old installations) you can just add a short hint in troubleshooting with a reference to the output-buffering section. I don't think there should be duplicated content in the guide. 
You can also use your changes and change the guide `output_buffering=0` --> `output_buffering=Off` to improve the readability. There is always room for improvements.

Please read the [CONTRIBUTION.md](https://github.com/Uberspace/lab/blob/main/CONTRIBUTING.md) and [STYLE.md](https://github.com/Uberspace/lab/blob/main/STYLE.md)
You missed at least these points:
> * Use English language. You don't have to be a native speaker or a poet.

> * If there are files to edit, don't do stuff like cat > ~/.npmrc <<__EOF__, just tell the user to edit the file.

You added your content below the link references. Please make sure you keep the structure as suggested by the boilerplate from the Style-Guide.

```
.. These are link references at the end of the document,
.. so every contributor can find them easily by scrolling to the bottom:
.. _Loremipsum: https://en.wikipedia.org/wiki/Lorem_ipsum
.. _feed: https://github.com/lorem/ipsum/releases.atom
.. Please do not add guide content below this section!

----

Tested with Loremipsum 1.22.1, Uberspace 7.1.1

.. author_list::
```

TalusUnheil:
I actually overlooked the given solution as an initial step in this guide because I only stumbled across the error after updating an older version, so I went straight to "Troubleshooting". Next time I will first "Ctrl+F" the problem. I also still have the German language in my fingers from Uberlab 6 DokuWiki times.

All in all, not a brilliant performance; I vow to do better!

PR #930, opened 2021-02-14
Guide to install searx on an Uberspace

maitrenageur:
Guide to install searx on an Uberspace.


source/guide_searx.rst
```diff
@@ -0,0 +1,242 @@
...
+.. code-block:: console
+
+ [isabell@stardust ~]$ git clone https://github.com/searx/searx.git ~/opt/searx/
+
+Step 2 - Python Module Installation
+-----------------------------------
+
+----
+
+.. note:: Here we use python 3. This is the reason, that these guide consider this version aspect to the installation tool like ``pip`` and later to ``python`` it self.
```
luto, line 60: I am not quite sure what the 2nd sentence means :sweat_smile: How about this?
```suggestion
.. note:: Here we use python 3, instead of the default python 2. For this reason, all ``pip`` and ``python`` commands need a version postfix ``3`` like below.
```

source/guide_searx.rst
```diff
@@ -0,0 +1,242 @@
...
+
+Some python modules are necessary and will be installed in your uberspace with ``pip3`` and the option ``--user``:
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ pip3 install --user requests pyyaml pygments werkzeug babel flask flask_babel lxml langdetect python-dateutil
```
luto, line 66: This does not ensure the correct versions are installed. Luckily, searx comes with a [`requirements.txt` file](https://github.com/searx/searx/blob/master/requirements.txt) specifying the needed packages as well as their exact versions. Using that one instead of documenting the list here would make the guide more resilient to future updates and easier to use.

What do you think?
```suggestion
 [isabell@stardust ~]$ pip3 install --user ~/opt/searx/requirements.txt
 ```

File: source/guide_searx.rst [hunk 3]
```diff
@@ -0,0 +1,242 @@
...
+
+ [isabell@stardust ~]$ cp ~/opt/searx/utils/templates/etc/searx/use_default_settings.yml ~/etc/searx/settings.yml
+
+Now it's time to change some entries in the configuration file:
+
+1. Searx requires for the own instance a secret key. This random number will be created with openssl (16 digits) and placed direct in the config file via sed, the stream editor.
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ sed -i -e "s/ultrasecretkey/`openssl rand -hex 16`/g" ~/etc/searx/settings.yml
```
luto, line 89: While `sed` calls make the guide easier to use right now, it's harder to understand what's going on. Especially because `ultrasecertkey` is not a configuration key, but just an example value, which can randomly change in the future. Please adapt this section to generate a token first and insert it into the file manually, like `ADMIN_TOKEN` in the [bitwarden_rs guide](https://lab.uberspace.de/guide_bitwarden.html#install-bitwarden-rs).

source/guide_searx.rst
```diff
@@ -0,0 +1,242 @@
...
+At first we must create the service file ``~/etc/services.d/searx.ini`` and the following content:
+
+.. warning:: The PATH variable consider the installed Python modules regarding to the version 3.6 at this time point. Please have a look in the directory ``~/.local/lib/`` to be sure which version is the actual ones.
+
+.. code-block::
+ :emphasize-lines: 2
+
+[program:searx]
+environment =
+ PATH="%(ENV_HOME)s/opt/searx/:%(ENV_HOME)s/.local/lib/python3.6/:%(ENV_PATH)s",
```
luto, line 139: `PATH` specifies where to find binaries to run, but  `~/.local/lib/python3.6` and `~/opt/searx` both don't contain any binaries. Are you sure this is needed?

source/guide_searx.rst
```diff
@@ -0,0 +1,242 @@
...
+ SEARX_SETTINGS_PATH="%(ENV_HOME)s/etc/searx/settings.yml"
+autostart=yes
+autorestart=yes
+command=python3 %(ENV_HOME)s/opt/searx/searx/webapp.py
+
+
+We must report to supervisord that there is a new ini file to consider:
```
luto, line 146: Please use the [supervisord include](https://github.com/Uberspace/lab/blob/main/source/includes/supervisord.rst) here, like in the [bitwarden guide](https://lab.uberspace.de/guide_bitwarden.html#setup-daemon).

source/guide_searx.rst
```diff
@@ -0,0 +1,242 @@
...
+
+[isabell@stardust ~]$ uberspace web domain list
+isabell.example
+searx.isabell.example
+isabell.uber.space
+
+Step 6 - Web Backend Setup
+--------------------------
+
+This step is important, that your running searx instance is reachable from outside.
```
luto, line 187: Please use the [web backend include](https://github.com/Uberspace/lab/blob/main/source/includes/web-backend.rst) here, like in the [1up guide](https://lab.uberspace.de/guide_up1.html#configure-web-server).

source/guide_searx.rst
```diff
@@ -0,0 +1,242 @@
...
+.. note:: Please be sure, that your new subdomain is configured at your domain provider too.
+
+If you want to find out that everthing is maintained for the web server on your Uberspace account, use the command:
+
+.. code-block:: console
+
+[isabell@stardust ~]$ uberspace web domain list
+isabell.example
+searx.isabell.example
+isabell.uber.space
```
luto, line 182: The example domain in all guides is `isabell.uber.space`, except for guides that need multiple domains. Please remove this section and use `isabell.uber.space` everywhere else.

luto (changes requested):
Thanks for the guide! :tada: it's quite comprehensive :) I left a few comments to make to future proof and more consistent with the other guides in the lab. Let me know what you think!

source/guide_searx.rst
maitrenageur: Yes, you have right. I considered it in the guide update.

source/guide_searx.rst
maitrenageur: Yes, you have right. But it was necessary to consider additional options too.

source/guide_searx.rst
maitrenageur: Yes, you have right. I considered it in the guide update.

source/guide_searx.rst
maitrenageur: I had in a Python project a problem with environment settings to find the external modules (libraries). But in this case everthing is fine without the PATH.

source/guide_searx.rst
maitrenageur: Ok.

source/guide_searx.rst
maitrenageur: Ok.

source/guide_searx.rst
maitrenageur: @luto All comments by you are in a new version of my guide and here as attachment available:
[guide_searx.rst.txt](https://github.com/Uberspace/lab/files/5983661/guide_searx.rst.txt)

luto: thanks! you upload the changes here so they show up in the pull request? https://github.com/maitrenageur/lab/edit/main/source/guide_searx.rst

maitrenageur: @luto I'm not so familiar with GitHub, but I updated my repository with the new version of the guide and a logo image.

maitrenageur: @luto Is somthing else to do from my side?

luto:
Thanks for the guide! :tada:

maitrenageur:
@luto Thank you for your support and patience.

beatstick:
Hi, thanks for the guide! I used it to install and update for a couple of years now,  but searx is no longer maintained and probably should not be installed on any uberspace anymore. 

Maybe the guide could be updated to migrate to searxng?

PR #931, opened 2021-02-16
Revert changes in SMTP configuration block

quadratmuede:
Change partially reverted so that the SMTP block can be displayed again.

### DIFF
--- a/source/guide_bitwarden.rst
+++ b/source/guide_bitwarden.rst
@@ -118,7 +118,6 @@ Generate an ``openssl -base64`` key now and save it temporarily, you'll need it 
 Use your favourite editor to create ``~/bitwarden_rs/.env`` with the following content:
 
 .. code-block:: ini
-  :emphasize-lines: 1,4,5,8,9,11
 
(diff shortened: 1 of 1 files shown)


source/guide_bitwarden.rst
```diff
@@ -118,7 +118,6 @@ Generate an ``openssl -base64`` key now and save it temporarily, you'll need it
 Use your favourite editor to create ``~/bitwarden_rs/.env`` with the following content:
 
 .. code-block:: ini
-  :emphasize-lines: 1,4,5,8,9,11
 
  ADMIN_TOKEN=PASTE YOUR TOKEN GENERATED ABOVE HERE
  ROCKET_PORT=62714
```
EV21, line 123: Hi @quadratmuede, 
thanks for mentioning this bug :bug: 
The indentation level does mismatch. If you just delete one space character in front of `:emphasize-lines:` everything will be displayed as all code block content has the same level. Alternatively you can adapt the code block content by adding a space in front of each line. 
I think fixing a simple bug is better than deleting a nice but broken feature :wink:

luto:
fixed. Sorry about that!

PR #934, opened 2021-02-20
Use USER-variable for wp core install command as well

programmieraffe:


luto:
`isabell.uber.space` is the example domain used in all guides. It's meant to be replaced by whatever domain the user would like to use. Thanks for the patch, but everything is fine here.

PR #935, opened 2021-02-21
[mattermost] add directory to service file

dingensundso:
Changing the relevant paths in the configuration to absolute ones is unnecessary.
I think it's easier to just start the service in the right directory.


luto:
That's nice :) thank you!

PR #938, opened 2021-02-23
Delete guide_coturn.rst

### LINKED ISSUE #844: [coturn] most of the time installations don't work
In our experience in support, the guide only works sometimes and often does not work. We can only roughly isolate the problem and assume that it has something to do with the number of ports used.

Guides that only work sometimes and sometimes don't work do not provide a good experience for the users. If it's not clear why they don't work and the logs don't provide enough information we can't support users and the guide has to be removed.

@jo-mei as this is your guide do you have any idea how to go forward or add information to help debugging?

SalocinHB:
no fix after more than two months
closes #844


thisven:
I've written a guide for [eturnal](https://eturnal.net) #1300 to compensate for a missing STUN/TURN server on Uberspace. It runs stable since five days now and seems like a promising replacement.

PR #943, opened 2021-02-27
Correct backup command, adding timestamp to backups

quadratmuede:
Changing the command for sqlite3 backup command since the tilde "~" in the backup file path generates the error on uberspace. Using ``$HOME`` instead. Also providing an alternative version of making backups with timestamp as file name (i.e. no file overwritten in cronjobs) if this meets user's need better.

### DIFF
--- a/source/guide_bitwarden.rst
+++ b/source/guide_bitwarden.rst
@@ -207,9 +207,15 @@ You can create a backup of the database manually. ``cd`` to your project folder,
 
  [isabell@stardust ~]$ cd ~/bitwarden_rs/data
  [isabell@stardust data]$ mkdir db-backup
- [isabell@stardust data]$ sqlite3 ~/bitwarden_rs/data/db.sqlite3 ".backup '~/bitwarden_rs/data/db-backup/backup.sqlite3'"
+ [isabell@stardust data]$ sqlite3 ~/bitwarden_rs/data/db.sqlite3 ".backup '$HOME/bitwarden_rs/data/db-backup/backup.sqlite3'"
 
-.. note ::  You could run this command through a CRON job everyday - note that it will overwrite the same backup.sqlite3 file each time.
+.. note ::  You could run this command through a CRON job everyday - note that it will overwrite the same backup.sqlite3 file each time. If you want to save every version of the backup, please read further.
+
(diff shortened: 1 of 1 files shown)


source/guide_bitwarden.rst
```diff
@@ -207,9 +207,15 @@ You can create a backup of the database manually. ``cd`` to your project folder,
 
  [isabell@stardust ~]$ cd ~/bitwarden_rs/data
  [isabell@stardust data]$ mkdir db-backup
- [isabell@stardust data]$ sqlite3 ~/bitwarden_rs/data/db.sqlite3 ".backup '~/bitwarden_rs/data/db-backup/backup.sqlite3'"
+ [isabell@stardust data]$ sqlite3 ~/bitwarden_rs/data/db.sqlite3 ".backup '$HOME/bitwarden_rs/data/db-backup/backup.sqlite3'"
```
luto, line 210: I am not sure what's going on here. `~` is replaced by the shell, not sqlite so this should work fine. It also seems like it does:

```
[root@stardust ~]# echo sqlite3 ~/bitwarden_rs/data/db.sqlite3 ".backup '$HOME/bitwarden_rs/data/db-backup/backup.sqlite3'"
sqlite3 /root/bitwarden_rs/data/db.sqlite3 .backup '/root/bitwarden_rs/data/db-backup/backup.sqlite3'
```

Which client and which shell are you using?

quadratmuede: I am using Windows Terminal with Powershell. It is weird that ~ outside the quote can be replaced correctly but not the one inside the quote. Also tried WSL and bash on Mint Linux, still no luck, always got the same error:

```
[user@adrastea ~]$ sqlite3 ~/bitwarden_rs/data/db.sqlite3 ".backup '~/bitwarden_rs/data/db-backup/$(date +%Y-%m-%d).sq
lite3'"
Error: cannot open "~/bitwarden_rs/data/db-backup/2021-03-04.sqlite3"
```

With $HOME instead I could create backups successfully. 

I am not sure whether ~ is interpreted by sqlite3 as a database operator before it gets replaced by shell?

luto:
Thank you!

PR #973, opened 2021-03-26
Fixes #953: Set data directory to a writable place

### LINKED ISSUE #953: redis guide issue: Save not possible
@SalocinHB I encountered this issue when trying to install Redis for nextcloud. I started getting `500`s . The nextcloud log contains this message repeatedly:

```
: "RedisException",
    "Message": "MISCONF Redis is configured to save RDB snapshots, but it is currently not able to persist on disk. Commands that may modify the data set are
disabled, because this instance is configured to report errors during writes if RDB snapshotting fails (stop-writes-on-bgsave-error option). Please check the
Redis logs for details about the RDB error.
```

I do not have access to the redis logs under `/var/log/redis` but it seems to indicate, that it can't write to the file. I can replicate the error in the redis client.

Not sure if this is a config issue on my side or on the server side.

schoeke:
See https://github.com/Uberspace/lab/issues/953
Thanks to @riweckr for the help.

### DIFF
--- a/source/guide_redis.rst
+++ b/source/guide_redis.rst
@@ -47,6 +47,7 @@ Now create the config file ``~/.redis/conf`` with an editor of your choice and e
  unixsocket /home/<user>/.redis/sock
  daemonize no
  port 0
+ dir /home/<user>/.redis/
 
 Setup daemon
 ------------


luto:
Thank you! This fixes the problem, but I'd expect a default redis installation to not persist any data at all. I added a `save ""` line to prevent it from dumping the contents to the disk.

PR #975, opened 2021-03-30
Remove trailing "s" in supervisord setup

matthiasbeyer:
I bet that's an typo, is it @thisven (original author of this file)?


matthiasbeyer:
It seems it is not. :confused:

luto:
The `%(NAME)s` placeholder syntax is [required by supervisord](http://supervisord.org/configuration.html#environment-variables) and defined by python. So it's correct, yes.

Issue #997, opened 2021-04-13
[pelican] new guide


k11h-de:
[Pelican](https://docs.getpelican.com/en/latest/) is a python-based tool to generate static websites/blogs.
The most prominent example is probably [kernel.org](https://www.kernel.org/pelican.html).

I set up my website [k11h.de](https://k11h.de/) with pelican in only a few minutes and documented the steps required on an uberspace [here](https://github.com/k11h-de/k11h-website).
It is also a good example, how pelican can be used.

I would love to provide a more detailed guide to the uberspace lab.
Please let me know what would be your prefered way of contribution.

Cheers from Berlin
Karsten

k11h-de:
Thanks for merging, I hope some users find the guide helpful.

Cheers Karsten

PR #1003, opened 2021-04-17
New guide pelican

k11h-de:
As mentioned in the [issue](https://github.com/Uberspace/lab/issues/997), I have prepare a guide for pelican (a python-based static website generator).
I hope you find it useful.

Cheer Karsten


source/guide_pelican.rst
```diff
@@ -0,0 +1,182 @@
...
+
+  ::
+
+    pip3 install --upgrade --user -e "git+https://github.com/getpelican/pelican.git#egg=pelican"
+
+
+Generate your website
+=====================
+
+**Step 1:**
```
luto, line 62: Please give your steps clear names instead of numbers and make them proper headings. Numbered steps are very cumbersome when the guide is edited later.
```suggestion
Initialize
----------
```

source/guide_pelican.rst
```diff
@@ -0,0 +1,182 @@
...
+  > Do you want to generate a tasks.py/Makefile to automate generation and publishing? (Y/n) <enter>
+  > Do you want to upload your website using FTP? (y/N) <enter>
+  > Do you want to upload your website using SSH? (y/N) <enter>
+  > Do you want to upload your website using Dropbox? (y/N) <enter>
+  > Do you want to upload your website using S3? (y/N) <enter>
+  > Do you want to upload your website using Rackspace Cloud Files? (y/N) <enter>
+  > Do you want to upload your website using GitHub Pages? (y/N) <enter>
+  Done. Your new project is available at /home/isabell/pelican
+
+  [isabell@stardust isabell]$
```
luto, line 97: Since there was no additional `cd`, the prompt will stay the same.
```suggestion
  [isabell@stardust pelican]$ 
```

source/guide_pelican.rst
```diff
@@ -0,0 +1,182 @@
...
+  Done. Your new project is available at /home/isabell/pelican
+
+  [isabell@stardust isabell]$ 
+
+
+.. note:: In case you want to use a custom domain other than isabell.uber.space, you can enter it here or easily change it later in ``~/pelican/pelicanconf.py``
+
+**Step 2:**
+
+You need to modify ``~/pelican/Makefile`` to make sure the publish process is working correctly on your uberspace.
```
luto, line 104: Modifying the `OUTPUTDIR` varliable seems cleaner and easier to me, what do you think?
```
OUTPUTDIR=/var/www/virtual/$(USER)/html
```

luto (changes requested):
Thanks for the guide! It's nice and short :) I added a few comments to streamline it a bit

source/guide_pelican.rst
k11h-de: Hi Luto,

thanks for your review and suggestions.
I was about to implement the same, but there is a pelican [setting](https://docs.getpelican.com/en/stable/settings.html) named  "DELETE_OUTPUT_DIRECTORY"
From the docs:

> Delete the output directory, and all of its contents, before generating new files. This can be useful in preventing older, unnecessary files from persisting in your output. However, this is a destructive setting and should be handled with extreme care.

So I decided to go with the additional rsync as this allows to specify excludes like `--exclude /.htaccess`.

luto: Deleting everything seems fine to me. All guides in the lab should assume an empty uberspace. If you'd like, add a `.. warning::` block above the `make publish` step saying
> This step will delete all content at `~/html`, if any.

The default pelican step does not seem to generate a `.htaccess` file and the default uberspace does not come with one, so I am not quite sure which `.htaccess` you'd like to ignore or preserve?

k11h-de: I have added a custom .htaccess for http-> https redirect and securtiy headers. 
I am fine with adding the suggested warning and change. My [commit](https://github.com/Uberspace/lab/pull/1003/commits/ad722b0b4cf44a90fdb4478e6bd4108ec377f350) was added before I read your last comment.

luto: I see! HTTP->HTTPS is handled by us globally, so no HTTP connection will ever reach your uberspace. Security headers should be set using [`web header`](https://manual.uberspace.de/web-headers/).

k11h-de:
Let me know if you need any additional changes from my end. 
From my perspective, I am happy with the resulting guide.

luto:
Thank you! :tada:

PR #1006, opened 2021-04-26
[fastapi] update guide after fix in dependency

cmacht:
An [issue in a dependency](https://github.com/MagicStack/httptools/issues/59) previously limited usage to Python 3.8. Now that it's fixed, you can use the latest, Python 3.9.

I tested my way through the whole guide again to confirm it does work.


luto:
thanks!

PR #1014, opened 2021-05-01
Update umami guide

pReya:
* mysql password is not needed on the given operation
* `npm run build-mysql-client` is automatically included in `npm run build`


luto:
Thanks!

PR #1015, opened 2021-05-04
Update URL and a hint

ungeschneuer:
Update of documentation URL and hint concerning the configuration of the web server.

### DIFF
--- a/source/guide_miniflux.rst
+++ b/source/guide_miniflux.rst
@@ -119,7 +119,7 @@ Configure web server
 
 .. note::
 
-    Miniflux is running on port 9000.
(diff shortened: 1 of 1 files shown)


luto:
This is already [documented in the manual](https://manual.uberspace.de/web-backends/#specific-path). There are lots of possible ways to run any given software. The lab should document the most simple, standard way. Running tools in sub-paths also comes with lots of problems and bugs, so we tend to not document this at all.

luto:
applied the URL part here: https://github.com/Uberspace/lab/commit/68850f54055f2a41cb65a440ece822834a3c610a Thanks :)

ungeschneuer:
Good to know and now worries :)

Issue #1020, opened 2021-05-05
[Xdebug] Add guide for Xdebug


groetzner-net:
I just started to write a guide on how to set up [Xdebug](https://xdebug.org/)

luto:
https://twitter.com/ddiesenreither/status/1559437603295854592

PR #1021, opened 2021-05-05
[Xdebug] Add Guide for Xdebug

groetzner-net:
#1020 
Guide based on a support ticket[488303] from uberspace

### DIFF
--- a/source/guide_xdebug.rst
+++ b/source/guide_xdebug.rst
@@ -1,0 +1,149 @@
+.. highlight:: console
+
+.. author:: Sascha Groetzner <https://groetzner.net>
+
+.. categorize your guide! refer to the current list of tags: https://lab.uberspace.de/tags
+.. tag:: lang-php
+
+.. sidebar:: Logo
+
+  .. image:: _static/images/xdebug.svg
+      :align: center
+
+######
+Xdebug
+######
+
+
+.. tag_list::
+
+`Xdebug <https://xdebug.org>`_ is an extension for PHP, and provides a range of features to improve the PHP development experience.
+
+----
+
+.. note:: For this guide you should be familiar with the basic concepts of
+
+  * :manual:`PHP <lang-php>`
+
+License
+=======
+
+Xdebug is released under `The Xdebug License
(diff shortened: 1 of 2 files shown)


source/guide_xdebug.rst
```diff
@@ -0,0 +1,149 @@
...
+
+Installation
+============
+
+export PHP Path
+---------------
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ export PATH="/opt/remi/php80/root/bin:$PATH"
```
luto, line 56: This path is subject to change and we'd like to not document it. However, we can expose `phpize` binary just like we already do with the `php` one. Are there any other binaries that are needed to build the extension?

Once that change is released, I'll take your guide for a spin :) Will take a few weeks, though, because we are focusing on a different project right now.

groetzner-net: > Are there any other binaries that are needed to build the extension?
I don`t think so. `phpize` should be the only one

luto: phpize will land in the next U7 release next week-ish :)

luto: @groetzner-net `phpize` has been available without special paths for some time now, would you like to update the guide? :)

groetzner-net: Hi, I have tried it with phpize, however I fail now at the problem that php-config from PHP development headers is also required.

nichtmax: we'll expose `php-config` as well, I'll keep you posted.

brutus: Will be part of the next release (probably later this week).

brutus: started [rolling out](https://github.com/Uberspace/manual/pull/344/) today

nichtmax:
@groetzner-net There we go, you can use `php-config` without the full path now: https://manual.uberspace.de/changelog/#v7-11-4-short

luto:
https://twitter.com/ddiesenreither/status/1559437603295854592

PR #1027, opened 2021-05-12
bitwarden_rs renamed to Vaultwarden

sruf:
bitwarden_rs was renamed Vaultwarden to avoid confusion with the official server implementation. Paths within the repository were also updated, so these instructions failed.


luto:
Thanks! :tada:

PR #1029, opened 2021-05-22
Fixed missing Statping URL

nishad:
Fixed missing URL for Statping in the intro session. Added https://statping.com/ as the main Statping URL since the Github link already points to the repository.


luto:
Thanks!

PR #1030, opened 2021-05-24
Update guide_radicale.rst

robhuen:
* Use web backends instead of RewriteRule via .htaccess 
* updated tested versions

### DIFF
--- a/source/guide_radicale.rst
+++ b/source/guide_radicale.rst
@@ -161,29 +161,15 @@ For every following user use it without ``-c``:
 Configure web server
 --------------------
 
-In order for your Radicale instance to be reachable from the web, you need to put a file called ``.htaccess`` into your ``~/html`` folder (or any other DocumentRoot, see the :manual:`document root <web-documentroot>` for details), with the following content:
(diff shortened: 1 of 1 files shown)


luto:
The guide uses RewriteRules instead of web backends, because radicale does not (or did not?) support authenticating requests itself. If you'd like change this to use backends, which I welcome very much :grin:, we'll need new authentication somewhere. Right now it looks like radicale will just be exposed to the web without any auth?

robhuen:
According to the documentation (and the source) radicale nowadays supports authentication:

Doc: https://radicale.org/3.0.html#tutorials/basic-configuration/authentication

Source: https://github.com/Kozea/Radicale/tree/master/radicale/auth

In the current version of the guide the authentication was already enabled via the config file (see https://lab.uberspace.de/guide_radicale.html#create-config-files) using the same htpasswd file as the webserver. Even bcrypt instead of md5 was already set in the current guide. So nothing else in the guide needs to be changed.

luto:
ha, interesting. Thank you! :tada:

PR #1038, opened 2021-06-01
[jekyll] udpate guide to current versions

pascoda:
[isabell@stardust website]$ bundle install --path vendor/bundle
returns:

[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path 'vendor/bundle'`, and stop using this flag

so I also adapted that part (to `config set`, followed by `install`), and (successfully) tested the adapted steps on another fresh uberspace.

### DIFF
--- a/source/guide_jekyll.rst
+++ b/source/guide_jekyll.rst
@@ -43,12 +43,12 @@ All relevant legal information can be found here
 
 Prerequisites
 =============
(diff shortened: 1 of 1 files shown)


luto:
Thanks! :tada:

PR #1040, opened 2021-06-01
[kirby] check compatibility with current versions

pascoda:
worked just fine, updated the versions presented in the guide.


luto:
thanks! :tada:

PR #1043, opened 2021-06-09
[FoundryVTT] New Guide for FoundryVTT

revilowaldow:
Creation of a new guide for Foundry VTT, a self hosted virtual tabletop web application.


source/guide_foundryvtt.rst
```diff
@@ -0,0 +1,221 @@
...
+
+.. code-block:: ini
+
+ [program:foundry]
+ command=node %(ENV_HOME)s/foundryvtt/resources/app/main.js –dataPath=%(ENV_HOME)s/foundrydata
+ autostart=yes
+ autorestart=yes
+
+When you are finished press **Ctrl + O** then **Enter** to confirm.
+You have now saved the file and closed nano.
```
luto, line 187:
```suggestion
```
See above.

source/guide_foundryvtt.rst
```diff
@@ -0,0 +1,221 @@
...
+
+
+Start Foundry by first updating the service, and then telling it to start:
+
+::
+
+ [isabell@stardust ~]$ supervisorctl update
+ [isabell@stardust ~]$ supervisorctl start all
+ foundry: started
+ [isabell@stardust ~]$
```
luto, line 197:
```suggestion
.. include:: includes/supervisord.rst
```

source/guide_foundryvtt.rst
```diff
@@ -0,0 +1,221 @@
...
+======================
+
+Prepare installation folders
+-----------------------------
+
+``mkdir`` is the command we will use to create two new folders **foundryvtt** and **foundrydata**. 
+
+``mkir myfolder`` creates the folder *"myfolder"*.
+
+Press enter after each line:
```
luto, line 102:
```suggestion
Create the required directories:
```
Same as above. To keep them short, guides can assume a basic level of linux knowledge.

luto (changes requested):
Thank you & sorry for the detail. I left a few comments :)

source/guide_foundryvtt.rst
```diff
@@ -0,0 +1,221 @@
...
+  `Foundry Virtual Tabletop <https://foundryvtt.com/>`_ is an alternative to Roll20 and many other platforms that enable you to play tabletop role playing games online.
+  Foundry is not free software, but you can host it on your own server.
+  
+  The first step is to purchase a license from their `website <https://foundryvtt.com/>`_.
+  
+  If you pay by SEPA transfer the payment can take up to a week, faster payment methods are available.
+
+----
+
+.. note:: This guide assumes you are completely unfamiliar with the basic concepts of linux
```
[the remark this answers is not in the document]
revilowaldow: Thank you for your comments, no worries about the delay.

The whole guide is obviously targeted at novice users.
The SSH introduction included in the guide is deliberately brief so that only the information the user needs to get this specific application set-up is included.

Once the application is running the user would likely never need to use the terminal again (except to maybe restart the service).

If at all possible, keeping the SSH guide would be preferable.

source/guide_foundryvtt.rst
```diff
@@ -0,0 +1,221 @@
...
+
+To do this we start with a config file in the folder ``etc/services.d/foundry.ini``
+
+Run the command to create a file in the right folder with the text editor **nano**:
+
+::
+
+[isabell@stardust ~]$ nano etc/services.d/foundry.ini
+
+Then type or paste the following lines:
```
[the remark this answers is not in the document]
revilowaldow: Similarly to the comment above, the guide is targeted at novices with no experience.
I'm familiar with how divisive editor recommendation and preferences can be, I respect UberSpace's decision to keep out of it.

In this specific context, the user is expected to have never used SSH before, and the recommendation is made with the specific intent to minimise commands and complexity. They also will not have yet formed an opinion on editors and so this should not attract the kind of negative feedback that may come from enthusiastic users proselytising their preference.

Would the inclusion of a note that other text editors are available be sufficient to enable this to be included?
I note you haven't linked a manual page in your comment; can you recommend a starting reference that gives only the relevant information on creating and editing a file in 5 lines as this proposed guide does?

source/guide_foundryvtt.rst
revilowaldow: See above.

source/guide_foundryvtt.rst
revilowaldow: As above, this guide specifically does not want to assume that level of knowledge; it seeks to _concisely_ provide only the needed information to get this use case started.

SnortsAlot: I see your intent Luto, and having a specific concrete example for people to work from is helpful.

For people do indeed have a preference on their ssh/cli text editor, then they most likely have experience enough to interpret the commentary to use within that editor (after all preference suggests comparison and evaluation)

Snargle13 (commented):
As someone who came in with no ideal how to program or coding in general I would state that you need  ssh eliza@dolittle.uberspace.de, that you need to use the linix/node version, and that you EV_Home is your spaceuber username (atleast those were the points i mess up on when making the server)

nichtmax:
Explaining detailed steps in every single guide doesn't help. Just as SSH is explained in the [manual](https://manual.uberspace.de/basics-ssh/) (feel free to link to it), maybe a short introduction of different editors in the manual could be useful, probably in the "YOUR FIRST DAY" section. Then all guides could link to the articles in the manual and we only have to update one article and not every guide.

Explaining the respective author's favourite editor in each and every guide will lead to bloated guides and confused users who are presented with minimal instructions on how to use nano in this guide, vi in another guide etc.

We gave our [styleguide](https://github.com/Uberspace/lab/blob/main/STYLE.md) much thought and want to keep all guides consistent: 

> If there are files to edit, don't do stuff like cat > ~/.npmrc <<__EOF__, just tell the user to edit the file. Don't mention an editor like vi or nano.

revilowaldow:
> We gave our [styleguide](https://github.com/Uberspace/lab/blob/main/STYLE.md) much thought 

Updated to comply with the style guide

revilowaldow:
Line 99 needs to change to have the correct port number. It should be 30000 not 3000.

nichtmax:
Please get in touch via hallo@uberspace.de for your cup ☺️

nichtmax:
Good catch @revilowaldow, I quick fixed the port number: https://github.com/Uberspace/lab/commit/3109932ba59841213e3e22be07664ad94b68413c

PR #1044, opened 2021-06-10
Guide for matterbridge

thisven:
I've setup [matterbridge](https://github.com/42wim/matterbridge) to use IRC with Matrix (Synapse) over a bridge and came up with a lab guide.


thisven:
How to check why a process is killed? It seems like matterbridge is getting killed after some minutes.

thisven:
I'm now monitoring the matterbridge process using a bash script found [here](https://knoxxs.github.io/programming/linux/ops/2015/08/31/linux-find-out-how-the-process-got-killed/) to investigate this problem.

thisven:
Regarding ressources the process seems to behave normal:
```
Size    Resid.  Shared  Data    %
188912  5801    3320    176754  0.1
```
Resident set size is ranging from 5778 to 6102. Does anyone know if this is a problem?

thisven:
I just opened an [issue](https://github.com/42wim/matterbridge/issues/1514) in the matterbridge repository.

thisven:
It seems like the matterbridge process gets killed by the kernel (exit code: 137) due to an out of memory error as @42wim mentioned in the [related issue](https://github.com/42wim/matterbridge/issues/1514#issuecomment-867994223) and suggests to check `dmesg`, but only superuser can run that command.

How to proceed now?

luto:
Each uberspace can use up to 1.5GB of RAM. If the process gets killed automatically, it exceeded that amount. In that case matterbridge probably won't run properly on an uberspace. There is nothing we can do about that, sorry :/

42wim:
fyi, matterbridge does not take 1.5GB of ram, my own instance takes 30MB

thisven:
I can confirm that. It's just about [23 MB on Uberspace](https://github.com/42wim/matterbridge/issues/1514#issuecomment-871670666).

thisven:
I [compiled the sources](https://github.com/42wim/matterbridge#building) locally as this step takes more than 1.5 GB of memory and used that binary, but I can't run it:
> ./matterbridge: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./matterbridge)

thisven:
To decrease memory usage for building [I used tags](https://github.com/42wim/matterbridge/issues/1514#issuecomment-874504257) to exclude some bridges and was able to compile the sources, but it still gets killed.

PR #1046, opened 2021-06-12
Add new Overview page

tomaculum:
I really like to browse the lab guides as a source for inspiration. In its current form it is pretty time consuming to click on every entry of the side bar, since most of the names are not self explanatory.

To solve this issue, I created a new overview page, which lists all guides with its short description. The existing description now has a `.. abstract::` directive, which outputs the text in the guide as expected but makes it also accessible for the new overview page.

image

I never heard of sphinx before starting this one day ago and its documentation is not very beginner friendly. Therefore I used a lot of code from the existing `authorships` package and most likely there is a more efficient way to achieve this.

There are a few things I want to improve, but I hope someone with actual knowledge about sphinx could guide me a bit:

- [ ]  fix references from guides which point to an headline

In almost every guide there is a link to itself in the first sentence, like for example: 

```rst
.. abstract::
  Ackee_ is a self-hosted open-source web analytics tool built with :manual:`Node.js <lang-nodejs>`.
```

External links do work as well as the link in the title but the `Ackee_` reference does not work, since it tries to open some kind of id:
image

I guess one could replace every link as such `*overview.html#id1` with the correct `.html` link after generating the html code. Is there an easier or more elegant way to do this?

- [ ] Are there some kind of dividers or something better to divide each entry visually from each other?

- [ ] It would be possible to add authors and tags to this overview page. 

I am a bit torn if this is a good idea or if it defeats the purpose for a quick overview. 

*I enabled the "Allow edits by maintainers" checkbox, if you want to edit/improve something.*


luto:
Including big changes like this without prior discussion is pretty hard :/ I'm sorry, but I won't be able to merge this as-is. In the future, please first open an issue describing your idea, work with us and other users to find a solution that works for everyone and then start working. Thank you!

PR #1047, opened 2021-06-20
Create User CLI usage for Gitea changed

Ryuno-Ki:
According to https://docs.gitea.io/en-us/command-line/#admin it's `user create` instead of `create-user` now.


luto:
Thanks! :)

PR #1049, opened 2021-06-26
Fixes for pixelfed guide

amenk:
* Fix formatting
* Config editing has to take place before key:generate, otherwise it will fail because of the missing redis connection
* Put MySQL Password in Quotes for more compatibility


luto:
Thanks! :)

PR #1050, opened 2021-06-27
[Prosody] Use a suitable example domain

magicbrothers:
One should use domains like example.org or example.com in documentation, because they explicitly allow that kind of usage.
Other domains (like isabell.org) don't.


luto:
Thanks! :tada:

PR #1052, opened 2021-06-27
[Prosody] Use file-based storage

magicbrothers:
Prosody developers discourage the use of MySQL.
Additionally, using file-based storage simplifies the installation process.

### DIFF
--- a/source/guide_prosody.rst
+++ b/source/guide_prosody.rst
@@ -1,6 +1,7 @@
 .. highlight:: console
 .. author:: Arian Malek <https://fetziverse.de>
 .. author:: fapsi
+.. author:: magicfelix
 
 .. tag:: Instant Messaging
 .. tag:: Jabber
@@ -28,7 +29,6 @@ XMPP is an open and free alternative to commercial messaging and chat providers.
   * :manual:`Domains <web-domains>`
   * :manual:`Firewall Ports <basics-ports>`
   * :manual:`HTTPS <web-https>`
-  * :manual:`MySQL <database-mysql>`
   * :manual:`supervisord <daemons-supervisord>`
 
 This guide is based on the initial `pull request`_ from fapsi_.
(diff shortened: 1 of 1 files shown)


luto:
Using the file system as a database creates a lot of unnecessary I/O, so I'd like to avoid this. The [official docs](https://prosody.im/doc/storage) list (My)SQL as a standard storage adapter and the [example config](https://prosody.im/doc/example_config) also documents it as an option. I can't see any warnings anywhere. Can you point me to the correct place in the docs, please?

magicbrothers:
What I heard from @Zash is, that MySQL is often a pain to work with.
For small servers he recommends using the file backend. But if that's not an option for you, maybe go with SQLite (or PostgreSQL, but this again adds complexity).

Zash:
> What I heard from @Zash is, that MySQL is often a pain to work with.

Yes, MySQL / MariaDB has been the source of a number of painful bugs. If it weren't the most popular choice for SQL (according to a survey a few years back) we would most likely have dropped support for it.

> Using the file system as a database creates a lot of unnecessary I/O

`[citation needed]`

> The official docs list (My)SQL as a standard storage adapter [...]

The docs also lists the `null` and `memory` storage backends. Being available as an option is not meant as a recommendation.

> I can't see any warnings anywhere. Can you point me to the correct place in the docs, please?

"Don't use MySQL" is my personal recommendation, not (yet) an official stance of the Prosody project. The problems with MySQL is primarily documented in the [issue tracker](https://issues.prosody.im/?q=sql) and various commit messages.

The official recommendation is to stay with the default `storage = "internal"` unless you have good reason otherwise.

luto:
> > Using the file system as a database creates a lot of unnecessary I/O
>
> [citation needed]

This statement might be wrong for prosody, yes! However, other apps like Nextcloud significantly increase system load when  using sqlite.

> The official recommendation is to stay with the default `storage = "internal"` unless you have good reason otherwise.

We'll go with that then. Thanks for providing additional info on this! It would be great to have this reflected in the official docs somewhere.

PR #1055, opened 2021-07-10
[vaultwarden] Add instructions for easy install/prebuilt binaries

pc-coholic:
I've been using vaultwarden for some time and always got annoyed by just how much space is taken by cargo, rust and all the other tiny bits and pieces...

I will admit: Using the prebuilt binaries, especially when it comes to storing ones passwords, is probably questionable at least.

But then - on the other hand - it's not much worse than just running the docker container itself, which is also one of the recommended ways to using vaultwarden...

I've split off the last part of the original installation-section and moved it to a new subsection ``Basic configuration``, since the steps explained therein are identical for both install variants.

### DIFF
--- a/source/guide_bitwarden.rst
+++ b/source/guide_bitwarden.rst
@@ -1,6 +1,7 @@
 .. highlight:: console
 
 .. author:: Achim | pxlfrk <hallo@pxlfrk.de>
+.. author:: Martin | pc-coholic <martin@pc-coholic.de>
 
 .. tag:: passwordmanager
 .. tag:: rust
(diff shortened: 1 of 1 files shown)


luto:
> always got annoyed by just how much space is taken by cargo, rust and all the other tiny bits and pieces...

rust should be preinstalled on uberspace, so I am not sure what is taking up space here? the temporary build files and the source? I'd be fine with adding an optional deletion step to the guide to save some space. What do you think?

Other than that I'd prefer to stay with the currently documented docker method because it seems less hacky to me and doesn't required any 3rd-party tools.

nichtmax:
@pc-coholic I, as @luto, see your point: it would be nice to have prebuild binaries. Extracting those from docker containers though is rather hacky and makes things complicated. I suggest to [ask vaultwarden](https://github.com/dani-garcia/vaultwarden/pull/1170) for prebuild binaries. 

I'll close this PR until there are binaries from vaultwarden and I hope you're not disappointed 😞

pc-coholic:
Hi Moritz, luto,

Sorry for not reacting any faster - I must have overlooked the first notification somehow...

Absolutely no issue with closing this - but to answer lutos interjection: I might have perhaps had a brainfart and just wrongly imagined that since the original guide was based on [this guide](https://vigonotion.com/blog/install-bitwarden-rs-on-uberspace/), I would also end up with all the build-env... I should have RTFM better ;-)

PR #1056, opened 2021-07-14
Update guide_gitea.rst

MatthiasMerk:
Just a minor fix. I wanted to upgrade gitea and couldn't import the key from keys.gnupg.net:
According to gitea [installation from binary docs](https://docs.gitea.io/en-us/install-from-binary/) the keyserver changed to keys.openpgp.org. The previously used keyserver keys.gnupg.net seems to be offline (gpg: keyserver receive failed: No name).
Also bumped gitea version to current release 1.14.4

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -56,10 +56,10 @@ Check current version of Gitea at releases_ page.
 
 .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


luto:
Thanks!

PR #1060, opened 2021-07-18
feat: update umami guide how to limit nextjs ram

pReya:
* add a warning regarding the `next build` process ram usage, which will make it fail occasionally


luto:
Thanks1

PR #1063, opened 2021-07-20
[directus] inital guide for directus 9 (node)

### LINKED ISSUE #1061: [directus] add guide
To date there is no guide on how to set up [directus ](https://github.com/directus/directus) 9 (node) on uberspace. As I sucessfully deployed it on uberspace I would like to offer to write up a guide for others.

j3n57h0m45:
This is the first draft of the guide, to get started with directus.

Closes #1061

On another note: any ideas whether the ``Error: /lib64/libc.so.6: version GLIBC_2.25`` is an uberspace or a node module issue?


nichtmax:
Hey, sorry for the late answer. We'll take a look if we can do something about `Error: /lib64/libc.so.6: version GLIBC_2.25` and then I'll review your guide. Thanks!

nichtmax:
CentOS 7 only provides `glibc 2.17`, that's the reason for `Error: /lib64/libc.so.6: version GLIBC_2.25`. We can't fix that at the moment but since you found and documented a workaround that's fine.

Would be nice to mention the reason for the error in your guide.

I'll review tomorrow.

j3n57h0m45:
Alles klar, Hinweis ist aufgenommen. Letztlich scheint das ein bekannter "Fehler" zu sein (z.B: directus/directus#7251)

source/guide_directus.rst
```diff
@@ -0,0 +1,154 @@
...
+==============
+
+Web backend
+-----------
+
+.. warning:: This exposes Directus_ to the global internet and allows anyone to query your directus database. Not recommended without authentication!
+
+.. note::
+
+    Directus_ is running on port 8055. The ``--remove-prefix`` parameter is needed if you want to run Directus_ under a sub URI like ``/directus`` instead of the root URI ``/``.
```
nichtmax, line 128: Please make it as simple as possible. [Don't mention sub domains or folders](https://github.com/Uberspace/lab/blob/main/STYLE.md):

nichtmax (changes requested):
Looks good, just a few things.

source/guide_directus.rst
```diff
@@ -0,0 +1,154 @@
...
+.. note::
+
+  As CentOS 7 (uberspace) only provides ``glibc 2.17``, the ``argon2`` module on needs to be manually recompiled. This will avoid the ``Error: /lib64/libc.so.6: version GLIBC_2.25``.
+    
+  .. code-block:: bash
+    
+    [isabell@stardust ~]$ cd ~/directus
+    [isabell@stardust ~]$ npm install @mapbox/node-pre-gyp
+    [isabell@stardust ~]$ npm install argon2 --ignore-scripts
+    [isabell@stardust ~]$ npx node-pre-gyp rebuild -C ./node_modules/argon2
```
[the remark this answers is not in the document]
j3n57h0m45: so i guess ``npx``  is not part of a verbatim uberspace?

source/guide_directus.rst
j3n57h0m45: removed section

source/guide_directus.rst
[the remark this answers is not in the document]
j3n57h0m45: To get around the ``npx`` issue and avoid pollution the global modules I rewrote the procedure. However if ``npx`` is available, this could simplify the procedure.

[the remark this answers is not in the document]
j3n57h0m45: added note on perquisites to have it available

source/guide_directus.rst
```diff
@@ -0,0 +1,154 @@
...
+=======
+
+All relevant legal information can be found here in the GitHub_ repository of the project:
+
+  * https://github.com/directus/directus/blob/main/license
+
+Prerequisites
+=============
+
+n/a
```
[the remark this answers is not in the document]
j3n57h0m45: ok - sure

j3n57h0m45:
> Looks good, just a few things.

ready for round 2 :)

source/guide_directus.rst
```diff
@@ -0,0 +1,163 @@
...
+
+  * https://github.com/directus/directus/blob/main/license
+
+Prerequisites
+=============
+
+Node and npm
+------------
+
+We're using :manual:`Node.js <lang-nodejs>`, its package manager :manual_anchor:`npm <lang-nodejs.html#npm> as well as ``npx``:
```
nichtmax, line 44: I Think you're missing a ` here. Check out the preview, the link is broken: https://deploy-preview-1063--uberlab.netlify.app/guide_directus.html

source/guide_directus.rst
```diff
@@ -0,0 +1,163 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ uberspace tools version show node
+ Using 'Node.js' version: '16'
+ [isabell@stardust ~]$ npm --version
+ 7.20.0
+ [isabell@stardust ~]$ npx --version
```
nichtmax, line 52: As a new user this command fails. I still don't know what `npx` is here. Leave it out if we have to install it via `npm`.

source/guide_directus.rst
```diff
@@ -0,0 +1,163 @@
...
+ [isabell@stardust directus]$
+
+.. note::
+
+  As CentOS 7 (uberspace) only provides ``glibc 2.17``, the ``argon2`` module on needs to be manually recompiled. This will avoid the ``Error: /lib64/libc.so.6: version GLIBC_2.25``.
+    
+  .. code-block:: bash
+    
+    [isabell@stardust ~]$ cd ~/directus
+    [isabell@stardust directus]$ npm install @mapbox/node-pre-gyp argon2 directus
```
nichtmax, line 73: Can we install `npx` via `npm`? Should this be...

`npm install npx @mapbox/node-pre-gyp argon2 directus`?

nichtmax (changes requested):
We always check if new guides work on a newly created account and I suggest you do the same. Accounts are free for the first month ☺️ 

I have honestly no idea what npx is and how to get it, my guess would be an alternative to npm? But as a new user this guide does not work for me but I think it's fairly easy to fix.

source/guide_directus.rst
j3n57h0m45: fixed

source/guide_directus.rst
j3n57h0m45: Actually added as a global dependency. So this should work.

j3n57h0m45:
Ok tested and verified with a new verbatim uberspace account. Maybe third time is the charm :)

nichtmax:
Looks good, thank you! I only removed the absolute path to npm since it's not necessary. Accidentally hit return while writing the commit message so that's only `Don` 😅 

Hit us up via hallo@uberspace.de for your well earned cup 🎉

PR #1070, opened 2021-07-28
Show scrollbar for Navigation / Sidebar

kraftner:
https://twitter.com/tkraftner/status/1420304313667268608

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -27,6 +27,10 @@ footer {
 
 /* --- Navigation / Sidebar ------------------------------------------------- */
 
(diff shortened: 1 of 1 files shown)


kraftner:
Just to clarify: With a mouse without a mousewheel you currently can't scroll at all. In any case on Desktop you can't even see that there is more you can scroll down to.

brutus:
Good idea. The huge scrollbar might be somewhat "invasive" though. How about [something like this](https://github.com/Uberspace/lab/pull/1073)?

This also keeps the search and main links in focus.

kraftner:
closed in favour of #1073

PR #1071, opened 2021-07-28
Notice the memory limit in PHP.ini

Ryuno-Ki:
As I ran to this issue today, I want to see it documented. I'm open for additional wording suggestions.
My setup migrated from U6 to U7. Formerly, I only had a php.ini with the content of Output Buffering and Memory Limit to be able to run Nextcloud.

The instructions here made many errors go away.
I only ran into some troubles with well known URLs but could resolve them. Do we want to document them here, too?
(Basically redirecting to index.php well-known URLs)

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -103,6 +103,12 @@ Nextcloud `recommends 512 MB <https://docs.nextcloud.com/server/latest/admin_man
 .. code-block:: ini
 
  memory_limit=512M
+ 
+If you still see an error in the web UI, edit ``~/etc/php.d/php.ini`` with the following content:
+
+.. code-block:: ini
+
+ memory_limit=512M
 
 Output Buffering
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


source/guide_nextcloud.rst
```diff
@@ -100,6 +100,12 @@ PHP Memory
 
 Nextcloud `recommends 512 MB <https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html>`_ of memory for an installation. To set the PHP memory limit to 512 MB, create the file ``~/etc/php.d/memory_limit.ini`` with the following content:
 
+.. code-block:: ini
+
+ memory_limit=512M
+ 
+If you still see an error in the web UI, edit ``~/etc/php.d/php.ini`` with the following content:
```
EV21, line 107: It looks like that file only exist because you migrated from U6. You may add this to the troubleshooting section. The guide assumes to create a new U7 with default settings.

Ryuno-Ki: Fine with me!
I mark this as a todo because I'm about to call it a day for today.

Ryuno-Ki: Moved.

nichtmax:
@Ryuno-Ki thanks for documenting the fix and @EV21 thanks for reviewing! 💞

PR #1073, opened 2021-07-29
:lipstick: only show scrollbars for guide section in side nav

brutus:
adds scrolling [as suggested](https://github.com/Uberspace/lab/pull/1070), but only for the guides section instead the whole side navigation.

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -74,6 +74,24 @@ footer {
   border-color: #84000f;
 }
 
+/* make (only) guide section scrollable  */
+
+.wy-body-for-nav,
+.wy-grid-for-nav
+.wy-nav-side,
+.wy-side-scroll,
+.wy-menu-vertical
+{
+  display: flex;
+  flex-direction: column;
+  min-height: 0;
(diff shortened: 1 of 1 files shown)


kraftner:
Yeah, sure, why not. Maybe making the scrollbar a little less invasive is an option too: https://css-tricks.com/the-current-state-of-styling-scrollbars/

brutus:
Yes! definitly :) Was just a quick test in Firefox to gather feedback… I just noticed now how ugly this is in Chromium :sweat_smile:

kraftner:
( :+1: for using Firefox as primary browser! )

I am glad you're working on it and the approach sounds good to me.  So, I guess you'll do great from here, I'm dropping out here and go back to my work ;)

brutus:
UX opinion came in, not in favour of only scrolling the guides; we now scroll the whole navigation - but not the sidebar head (with logo and search). Also some :lipstick: to bring Chrome (Edge, Safari…) more in line with Firefox scrollbar design.

kraftner:
:heart_eyes:

PR #1074, opened 2021-07-30
Web Backend Config

philipbrembeck:
Made the `Web Backend Config` more clear, with detailed instructions for Ackee


nichtmax:
We [explicitly ask](https://github.com/Uberspace/lab/blob/main/STYLE.md) to include generic snippets to keep all things uniform. I don't see any reason to not include the generic section here, except for your addition of the domain, which we also ask you to not mention to keep guides simple ☺️ 

Feel free to reopen the MR if I missed something.

PR #1076, opened 2021-08-01
[Nextcloud] ✏ update database maintenance commands

EV21:
An anonymous human mailed me and asked for these changes.

I also enabled a command in the update script since NC19 is EOL


nichtmax:
Thanks! Always a pleasure ☺️

PR #1080, opened 2021-08-05
use Ghost-x.yy.z.zip for updates

### LINKED ISSUE #1079: [Ghost] more specific instructions on which zip to use to update Ghost
The `Ghost-x.yy.z.zip` and the source code zip have different directory structures. `Ghost-x.yy.z.zip` should be used.

SalocinHB:
closes #1079

### DIFF
--- a/source/guide_ghost.rst
+++ b/source/guide_ghost.rst
@@ -232,7 +232,7 @@ Updates
 Download and unzip new version
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


noave:
we should be more clear here, either "copy link from" or "replace version numbers in the code". I prefer replacing just version numbers in the code, we should have a consistent style for this over all guides

nichtmax:
Let's see what general approach we can agree on in https://github.com/Uberspace/lab/issues/1084 @noave 

Until then this PR plays nicely with our [style guide](https://github.com/Uberspace/lab/blob/main/STYLE.md).

PR #1091, opened 2021-08-17
Update guide_wekan.rst

logazer:
Fixed the old project URL (wekan.org) because it's offline.

The whole guide would still require an overall update, as the linked [release](https://releases.wekan.team/wekan-4.43.zip) is already non-existent.

Unfortunately, I don't have time for that right now. Maybe someone else could look into it. The project is still actively developed (see [changelog](https://github.com/wekan/wekan/blob/master/CHANGELOG.md)).


luto:
Thanks!

PR #1092, opened 2021-08-20
[uu-notify] some small corrections [non-functional]

franok:
* be more precise about github personal access token scopes
* fixed a link being relative to the lab, should be absolute --> use protocol prefix `https://`


luto:
Thanks!

Issue #1100, opened 2021-08-31
[vaultwarden] guide is broken


schneider-de-com:
The guide for Vaultwarden does not work anymore. After entering the command "cargo build -j 1 --release --features sqlite" build stops with the message that there are modules that are not needed. We added a banner to the guide to signal the broken state.

If you have any insights, please comment on this issue. Pull requests fixing the guide are also highly appreciated. If the guide does not get fixed within two months, we will remove it from the lab, as documented in the [broken guides policy](https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md).

@brutus  @sruf @quadratmuede @pxlfrk FYI.

pxlfrk:
Hey @permanentmedia, thanks for your Contribution and hint regarding the error.
I am the original author of the guide and tested the guide in its current state yesterday on a brand new uberspace.

I can confirm the warning you describe, but it's just a compiler warning, **not an error.**
If you let run the build process after the warning shows the build finishes as expected.

```
[user@host vaultwarden]$ cargo build -j 1 --release --features sqlite
   Compiling vaultwarden v1.0.0 (/home/$USER/vaultwarden)
warning: unused imports: `Build`, `Rocket`, `self`
  --> src/util.rs:7:15
   |
7  |     fairing::{self, Fairing, Info, Kind},
   |               ^^^^
...
10 |     Data, Request, Response, Rocket, Build
   |                              ^^^^^^  ^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: 1 warning emitted

    Finished release [optimized] target(s) in 8m 53s
```

So instead of marking the guide as broken i would prefer adding a hint that a warning may occur.
Some words regarding the needed workaaround with the `async`-branch:
Vaultwarden is build upon/with [Rocket](https://github.com/SergioBenitez/Rocket) that in turn needs a rust nightly or dev version to compile successfully (see #708). Rocket has alredy implemented [rust stable](https://github.com/SergioBenitez/Rocket/issues/19), this changes will be released with Rocket 0.5 which is [mostly finished](https://github.com/SergioBenitez/Rocket/milestone/8). As soon as this new Rocket version is released, vaultwarden [will update its dependencies](https://github.com/dani-garcia/vaultwarden/issues/712#issuecomment-658206248) and our patch and hopefully also this warning will be gone.

permanentmedia:
Hey @pxlfrk 

thank you for analysing the warning. 

[...]

PR #1104, opened 2021-09-07
[Vaultwarden] Update "Update" Section

remarcable:
The steps in the Update section were incomplete.
- Replace `web-vault-feed` link with https://github.com/dani-garcia/bw_web_builds/releases
- Add `git checkout origin/async` to apply new changes from remote

### DIFF
--- a/source/guide_bitwarden.rst
+++ b/source/guide_bitwarden.rst
@@ -288,6 +288,7 @@ Updating bitwarden_rs is really easy. Just stop the server, pull everything and 
  [isabell@stardust ~]$ cd ~/vaultwarden
  [isabell@stardust bitwarden_rs]$ supervisorctl stop vaultwarden
  [isabell@stardust bitwarden_rs]$ git pull
+ [isabell@stardust bitwarden_rs]$ git checkout origin/async
  [isabell@stardust bitwarden_rs]$ mv web-vault web-vault.old && mkdir web-vault
  [isabell@stardust bitwarden_rs]$ wget new-release.tar.gz
  [isabell@stardust bitwarden_rs]$ tar -xvzf new-release.tar.gz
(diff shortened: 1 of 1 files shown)


source/guide_bitwarden.rst
```diff
@@ -288,6 +288,7 @@ Updating bitwarden_rs is really easy. Just stop the server, pull everything and
  [isabell@stardust ~]$ cd ~/vaultwarden
  [isabell@stardust bitwarden_rs]$ supervisorctl stop vaultwarden
  [isabell@stardust bitwarden_rs]$ git pull
+ [isabell@stardust bitwarden_rs]$ git checkout origin/async
```
remarcable, line 291: I wasn't sure what to checkout here. The guide says in line 66 to use `origin/async` because of the missing dependency updates. `origin/async` didn't get any new updates so far, though, so I suppose the whole `Update` section isn't needed until the dependencies are updated and we can use the current version tags.

pxlfrk:
Hi Marc, thanks for your contribution! :) Your Fix with the wrong link (web-feed) is correct.
But i'm not sure if adding another `git checkout origin/async` is necessary since or local branch should still be on that remote branch  and `git pull` fetched the latest commits from that.

https://github.com/Uberspace/lab/issues/1100#issuecomment-912357238
Even since the async-branch hasn't got any updates in the last time it's still necessary to work in the uberspace environment. The official branch (main) depends on rust nightly which isn't available on uberspace - see further information in my linked comment in another issue. Hopefully this gets fixed soon and we can finally get rid of this clumpy workarround :)

luto:
merged: https://github.com/Uberspace/lab/commit/4477bd402cacad552b8293743d78ca47874cbcdb

Thank you! :)

PR #1124, opened 2021-10-28
[uu-notify] add optional instructions for app icon

franok:
changes:
- added section "Optional"
- writing Gotfiy uppercase where it is a name


luto:
thank you!

PR #1125, opened 2021-10-28
[uu-notify] add section "Troubleshooting"

franok:
adding a section "Troubleshooting" to the guide, which links to the project's github issues having a label "workaround". Users might find quick notes about issues in newer releases and potential workarounds. This is a faster channel, than updating the guide for every new bug or resolved bug.


luto:
Thank you :)

PR #1126, opened 2021-10-28
[Mailman 3] make Django Cluster use 4 worker processes

rattenburg:
Each worker process consumes 60MB memory. The default number of worker is equal to CPU count (ie. like 20, which uses way too much memory!!)

The other variables use the default values.


luto:
oh, thanks!

PR #1136, opened 2021-11-09
listmonk installation guide

maitrenageur:
This is a guide line to install listmonk on an Uberspace.

listmonk is a simple web based newsletter and mailing list manager.


SalocinHB:
When I set up listmonk as described in your guide, it tries to load its own logo form `http://localhost:9000/public/static/logo.png`, which fails. Is there some reverse proxy option missing from your guide? Once I'm logged in, all resources appear to load correctly.

maitrenageur:
> When I set up listmonk as described in your guide, it tries to load its own logo form `http://localhost:9000/public/static/logo.png`, which fails. Is there some reverse proxy option missing from your guide? Once I'm logged in, all resources appear to load correctly.

During my installation I didn't noticed this effect. How did you found out the missing file?
And from my point of view the reverse proxy is configured in the right way.

maitrenageur:
> listmonk binary is already executable.

Main idea was to consider the case, that the file is *not* executable. It is like a safety step, to avoid a missunderstanding in case of a missing executable setting.

luto:
Thank you! :tada:

PR #1142, opened 2021-11-17
[OpenSlides] Add a info in upgrade section

Lukas2112:
Important information for OpenSlides instances on Uberspaces that would like to upgrade to v3.4 or higher.

see: https://github.com/OpenSlides/OpenSlides/discussions/6283

### DIFF
--- a/source/guide_openslides.rst
+++ b/source/guide_openslides.rst
@@ -182,6 +182,7 @@ Updates
 
 .. note:: Check the update feed_ regularly to stay informed about the newest version.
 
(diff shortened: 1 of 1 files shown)


SalocinHB:
As I understand the comment, it should be possible, but is untested and probably requires a re-write of the guide.

Lukas2112:
> As I understand the comment, it should be possible, but is untested and probably requires a re-write of the guide.

Upgrading to v3.4+ using Pip will not work. 

We would need a new version of the guide because the respective services are set up manually, see https://github.com/OpenSlides/OpenSlides or the linked discussion.

PR #1143, opened 2021-11-19
[keycloak] Add guide for Keycloak

jlskuz:
Keycloak is an open source identity and access management system which allows single sign-on for web applications and services.


source/guide_keycloak.rst
```diff
@@ -0,0 +1,274 @@
...
+ Press ctrl-d (Unix) or ctrl-z (Windows) to exit
+ Password:
+ [isabell@stardust keycloak-15.0.2]$
+
+Setup daemon
+------------
+
+Use your favourite editor to create the file :file:`~/etc/services.d/keycloak.ini` with the following content. Replace the version in the archive file name with the one you just downloaded.
+
+.. code-block::
```
luto, line 124: The highlight doesn't work here because the language is not set, try this:

```suggestion
.. code-block:: ini
```

source/guide_keycloak.rst
```diff
@@ -0,0 +1,274 @@
...
+ :emphasize-lines: 2
+
+ [program:keycloak]
+ command=/home/smartlab/keycloak-15.0.2/bin/standalone.sh -b 0.0.0.0
```
luto, line 128: To make this fully copy-pasteable use `%(ENV_HOME)s`:
```suggestion
 command=%(ENV_HOME)s/keycloak-15.0.2/bin/standalone.sh -b 0.0.0.0
```

source/guide_keycloak.rst
```diff
@@ -0,0 +1,274 @@
...
+ [isabell@stardust ~]$ rm keycloak-15.0.2.tar.gz
+ [isabell@stardust ~]$
+
+Configuration
+=============
+
+Change the configuration
+------------------------
+
+Find the following block in file :file:`standalone/configuration/standalone.xml` and add the ``proxy-address-forwarding="true"`` attribute to ``<http-listener>`` element under ``<server>``.
```
luto, line 86: Please provide a full path here, including `~/keycloak-15.0.2`.

source/guide_keycloak.rst
```diff
@@ -0,0 +1,274 @@
...
+ [isabell@stardust ~]$
+
+Extract the archive
+-------------------
+
+Use ``tar`` to extract the archive and delete the archive. Replace the version in the archive file name with the one you just downloaded.
+
+::
+
+ [isabell@stardust ~]$ tar -xvzf keycloak-15.0.2.tar.gz
```
luto, line 69: This drops all the files in `~/keycloak-15.0.2`. What do you think about moving them to `~/keycloak`? That way all the paths for config, service file, etc. stay the same across updates. A `mv keycloak-15.0.2 keycloak` after the `tar` would do the job.

source/guide_keycloak.rst
```diff
@@ -0,0 +1,274 @@
...
+
+Find the following block in file :file:`standalone/configuration/standalone.xml` and add the ``proxy-address-forwarding="true"`` attribute to ``<http-listener>`` element under ``<server>``.
+
+.. code-block:: xml
+ :emphasize-lines: 4
+
+ <subsystem xmlns="urn:jboss:domain:undertow:10.0" default-server="default-server" default-virtual-host="default-host" default-servlet-container="default" default-security-domain="other" statistics-enabled="${wildfly.undertow.statistics-enabled:${wildfly.statistics-enabled:false}}">
+     <buffer-cache name="default"/>
+     <server name="default-server">
+         <http-listener name="default" socket-binding="http" redirect-socket="https" enable-http2="true" read-timeout="30000" proxy-address-forwarding="true"/>
```
luto, line 94: This snippet looks different than the one provided in the tarball (I think it's mostly the added `read-timeout` attribute):
```xml
<http-listener name="default" socket-binding="http" redirect-socket="https" enable-http2="true"/>
<https-listener name="https" socket-binding="https" security-realm="ApplicationRealm" enable-http2="true"/>
```
Can you take another look and align them, please?

source/guide_keycloak.rst
luto: Alternatively, add a symlink `~/keycloak-current` which always points to the current version.

luto (changes requested):
Thank you! The guide is nice and easy to follow :) I left a few comments to streamline it.

source/guide_keycloak.rst
jlskuz: I decided for symlinks since the `mv` approach would make the update process more complex as far as I can see.

jlskuz:
Thanks for the review! It should be all fixed/improved now. I hope haven't regressed something else…

luto:
Thank you! :tada:

PR #1147, opened 2021-11-28
use python3.6 -m pip instead of pip3.6

u0nel:
I got this warning:

WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.


luto:
Thanks!

PR #1148, opened 2021-11-29
Add overview of lab guides

kamicater:
This page contains the name and the first paragraph of all the labs so you don't need to click through all pages.

### DIFF
--- a/source/guide__overview.rst
+++ b/source/guide__overview.rst
@@ -8,41 +8,41 @@ Overview
 
 Ackee
 ==========
-.. tag:: #analytics #lang-nodejs #web 
+.. tag:: analytics lang-nodejs web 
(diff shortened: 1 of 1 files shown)


luto:
Thank you for this pull request!

It's true that an overview page would be nice to have, but we will have a hard time keeping this approach up-to-date in the long term. Any change to the existing guide's texts, tags (or existence) would now need to be reflected in the overview file as well. So we cannot merge this, as is. I'm sorry.

For further work please always check our [CONTRIBUTING.md](https://github.com/Uberspace/lab/blob/main/CONTRIBUTING.md):

> To signify intent and prevent duplicate work, please create an issue before working on your pull request.

This is especially true for larger changes like this.

PR #1163, opened 2021-12-14
[vaultwarden] Updated naming and commands

Tschoepel:
Updated bitwarden_rs naming to vaultwarden and updated the git pull command to reflect the checkout of the async branch earlier.


luto:
Thanks!

PR #1172, opened 2021-12-30
[directus] change listening URL, create web backend

Adashi12:
This is an update to the Directus guide to include the change of the listening URL and create a web backend.

### DIFF
--- a/source/guide_directus.rst
+++ b/source/guide_directus.rst
@@ -1,6 +1,7 @@
 .. highlight:: console
 
-.. author:: j3n57h0m45 <https://github.com/j3n57h0m45>, André Lehner <https://github.com/Adashi12>
+.. author:: j3n57h0m45 <https://github.com/j3n57h0m45> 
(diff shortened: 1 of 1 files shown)


source/guide_directus.rst
```diff
@@ -126,6 +126,29 @@ Create ``~/etc/services.d/directus.ini`` with the following content:
...
+Change listening URL
+------------
+
+In ``~/directus/.env`` change the PUBLIC_URL to ``http://0.0.0.0``.
+
+::
+
+  PUBLIC_URL="http://0.0.0.0"
+
+Create web backend
```
nichtmax, line 138: Please use the snippet: https://github.com/Uberspace/lab/blob/main/STYLE.md#rules-for-guides, e.g. https://github.com/Uberspace/lab/blob/main/source/guide_ghost.rst#configure-web-server

nichtmax (changes requested):
Thanks for the update! Just one small thing ☺️

PR #1186, opened 2022-01-24
[Focalboard] Layout: render code-block correctly

ebroda:
Currently in the [Guide](https://lab.uberspace.de/guide_focalboard.html) this codeblock is not rendered, but instead printed as plain text.


noave:
thanks! for the future: pls start your commit message with a `[focalboard]` :slightly_smiling_face:

Issue #1135, opened 2021-11-07
[mailman3] link to experimental automated deployment


cknoll:
I think uberspace guides would be more accessible if they were automated. I experimentally realized this idea for the imho nontrivial mailman3-guide. The current state is *works-for-me*. A reference in the [respective guide](https://lab.uberspace.de/guide_mailman-3.html) might be a reasonable idea for two causes: a) the automation in its current state might already be helpful for others and b) publication of the approach might spawn some discussion on how to automate similar guides.

→ See suggested https://github.com/Uberspace/lab/compare/main...cknoll:autodeploy-mailman3

However, I am unsure whether there is any future for the idea of automating uberspace-guides in general and my approach (see <https://codeberg.org/cknoll/uberspace-autodeploy-mailman3>) in particular.

Thus, before filing a PR I would welcome some feedback

noave:
hey, generally we would like to realize automated deployment in the future by using ansible. But there are still some obstacles we need to work on before we can provide a nice wrapper that makes ansible ready to work together with our userfacts like domains and ports etc. But of course you can already parse cli output etc.

Thta means using ansible for creating automated installations would be more future-safe from current point of view. But if there is like here a working approach with its own supported repository, it would be ok to link it from within the guide. But please make clear, that all issues regarding the scripts should be addressed to your repository :)

cknoll:
> But please make clear, that all issues regarding the scripts should be addressed to your repository :)
I updated my PR accordingly.


As to the general topic:
From my perspective, ansible might pose a too high hurdle for getting started and to implement customization. I tried several times but I found it hard to find out what actually happens on the server. Thus I wrote my [own little library](https://codeberg.org/cknoll/deploymentutils) which tries to combine the the understandability of shell scripts with the flexibility of ansible (using variables and tile templates). I very much prefer this (my) approach over some arcane playbook stuff – but that is obviously a biased view.

Maybe you should ask your users what they are likely to adopt.

PR #1189, opened 2022-01-27
mailman3: add notice and link to (third party) auto-deployment script

cknoll:
see #1135


noave:
hey, I have come back to this here. I unfortunately haven tested your script before merging probably, because its "only" external 3rdparty, but I tried it now. And it did not really work for me, the README was hard to follow, there was no requirements.txt present etc. It looks like it was built for your own purpose and worked, but there where still some of your personal domains within the configs :-) Do you like to work on this again and improve coding and README ? Otherwise I would -for now- remove the linking again until its tested to basically work for new accounts.

PR #1196, opened 2022-02-05
[libreoffice] update installation instructions for udocker and CODE

### LINKED ISSUE #1178: Adapt udocker installation part @ LibreOffice Guide
In the guide for the [Libreoffice Installation](https://lab.uberspace.de/guide_libreoffice.html) the installation of udocker is explained as follows:

`curl https://raw.githubusercontent.com/indigo-dc/udocker/master/udocker.py > $HOME/bin/udocker`

The udocker.py file does not exist anymore in this project so above request will fail. I assume the installation procedure of udocker has been adapted. As far as I can see here: https://indigo-dc.github.io/udocker/installation_manual.html it is proposed as follows:

`curl -L https://github.com/indigo-dc/udocker/releases/download/v1.3.1/udocker-1.3.1.tar.gz > udocker-1.3.1.tar.gz`
`tar zxvf udocker-1.3.1.tar.gz`
```export PATH=`pwd`/udocker:$PATH```

I tried this out in $HOME/bin/ and it works perfectly fine. The only thing I needed to additionally change was in the libreoffice.ini file.
Here I had to change:
command=bash -c '\
    %(ENV_HOME)s/bin/udocker run \

to:
command=bash -c '\
    %(ENV_HOME)s/bin/udocker/**udocker** run \ 

Since the udocker file is after following above procedure contained within the udocker folder.

Everything else worked perfectly fine!

Since in my opinion a new version of this guide makes sense since the current one is not working anymore due to the initial not working curl request I am proposing this here... Hope that is fine :)

//cc @jorsn

### LINKED ISSUE #1192: [libreoffice] guide broken
udocker URL does not work anymore

the more severe thing is that the path `https://xxx/loleaflet/dist/admin/admin.html` does not work.

jorsn:
fixes #1178
fixes #1192
allows changing the wopi host setting after the first run
1. udocker installation changed upstream
2. `collabora/code` container changed upstream, e.g. admin console URL
3.  container `libreoffice/online` seems to be abandoned for the time being
4. `collabora/code` only allows to change the `domain` setting one time. Later changes take no effect. I found a workaround.

### DIFF
--- a/source/guide_libreoffice.rst
+++ b/source/guide_libreoffice.rst
@@ -76,20 +76,37 @@ Setup the domain where your office installation will be accessible.
 
 .. include:: includes/web-domain-list.rst
 
-Then, install the latest stable version of `udocker`_:
+Then, look for the latest version number of `udocker`_ on the
+`github page <https://github.com/indigo-dc/udocker/releases>`_.
+Here, we assume that the version is 1.3.1.
 
 ::
 
-  [isabell@stardust ~]$ curl https://raw.githubusercontent.com/indigo-dc/udocker/master/udocker.py > $HOME/bin/udocker
+  [isabell@stardust ~]$ curl -L https://github.com/indigo-dc/udocker/releases/download/v1.3.1/udocker-1.3.1.tar.gz > udocker.tar.gz
     % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                    Dload  Upload   Total   Spent    Left  Speed
-  100  338k  100  338k    0     0   522k      0 --:--:-- --:--:-- --:--:--  521k
+  100   657  100   657    0     0   1294      0 --:--:-- --:--:-- --:--:--  1293
+  100 75791  100 75791    0     0  64650      0  0:00:01  0:00:01 --:--:-- 64650
+  [isabell@stardust ~]$ tar zxvf udocker-1.3.1.tar.gz
+  [isabell@stardust ~]$
+
+Then, create a wrapper script ``$HOME/bin/udocker`` with the following content:
+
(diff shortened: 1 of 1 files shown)


jorsn:
@therealsimwoe would you like to review the change, especially the udocker fix?

kimdiallo (changes requested):
Please excuse the delay - my query is already from March 15 and I simply neglected to submit it as well.

jorsn:
> Please excuse the delay - my query is already from March 15 and I simply neglected to submit it as well. 

Excused ;) Thanks for the query!

> Although it works as described with the wrapper, I wonder why it is necessary. A symlink does it also, is less error-prone and in addition still faster deposited. Or what do you think?

A symlink does it not, because the udocker python script uses $0 to set the python module search path (try it out, it breaks).

In the udocker installation manual, they simply add the udocker dir to $PATH, but I don't like that, since then the $PATH dirs are cluttered with files that shall not be executed, and one has to modify $PATH permanently.

nichtmax:
Thanks for the fix, I just merged it. 

A quick heads up: We're thinking about removing the guide because it's very complex and we're having many users with questions about the guide and udocker in general. 

Can you get in touch with us via hallo@uberspace.de please? Me might have another little something for you 👕 because this MR took so long...

jorsn:
> Thanks for the fix, I just merged it. 

Thanks!

> A quick heads up: We're thinking about removing the guide because it's very complex and we're having many users with questions about the guide and udocker in general. 

Of course I don't know about those questions, as I only had two (?) support requests so far, but if this can be alleviated by improving the guide, please let me know, e.g. what you/people find unclear about the guide.

PR #1205, opened 2022-02-17
rewrite manual ssh article as howto for lab

noave:


nichtmax:
> After starting your shell app you should see a window with something like this:

How about an actual screenshot so people know they are doing the right thing? Keep in mind they probably have never seen nor used a terminal before.

Here's a screenshot from MacOS, that should be enough for Windows users too because it's pretty similar. But it looks better 😅 

terminal

nichtmax:
@EV21 [I'll try again](https://github.com/Uberspace/lab/pull/1089#issuecomment-914160490): can you please get in touch with us via hallo@uberspace.de?

noave:
ok for me, but can you provide screenshot of a smaller window with not that much of black empty window space :)

noave:
how about renaming the article "SSH connections" to "SSH basics" ?

nichtmax:
Screenshot 2022-02-23 at 13 06 09

noave:
Sorry I have now created another terminal picture because the image looked to "strong" within the guide, the other one is simplier and with the localuser@localhost prompt

https://deploy-preview-1205--uberlab.netlify.app/howto_ssh.html#starting-your-local-shell

for me its ready to merge

nichtmax:
I really think we should keep the screenshots consistent. Personally I like the drop shadow, this is a question of personal taste in the end.

In [the website howto](https://lab.uberspace.de/howto_website.html) I used MacOS screenshots so I'll use these here as well. We can always change that later but should change it everywhere then. And I think your screenshot is not x2, it looks pretty blurry on my screen.

PR #1206, opened 2022-02-18
[conduit] Fix binary path in update section

jfowl:
As per https://github.com/Uberspace/lab/pull/1065#issuecomment-1045011557


jfowl:
Auch gut, Hauptsache gefixt :)

PR #1215, opened 2022-03-08
[Gitea] rework guide

### LINKED ISSUE #1219: [gitea] rework some guide steps
- the pgp verfification is no longer working, update it or maybe just leave it
- choose a single installation way, not a "minimum set and could be done via web"
- move the external renderer to a later configuration option, and explain what it is for
- do not create the admin user password on console, let the user do it her way
- also put the ssh key feature to later configuration
- do not provide subfolder installation, keept it simple
- add startsecs=30 to service config

EV21:
Changes:
- fix GPG key import
- add backup section
- add `gitea-update` script
  tested by updating 1.14.4 to 1.16.3

[👀 deploy preview](https://deploy-preview-1215--uberlab.netlify.app/guide_gitea.html)

- [x] closes #1219
- [x] guide is working with a new asteroid 🪨
  tested via
  ```bash
  bash -c "$(wget -q -O - https://github.com/EV21/uberspace-helper/raw/main/uberspace-gitea-installer.sh)"
  ```
  with
  ```bash
  GITHUB_API_URL=https://api.github.com/repos/$ORG/$REPO/releases/tags/v1.14.4
  GITHUB_API_URL=https://api.github.com/repos/$ORG/$REPO/releases/latest
  ```
- [x] run `make clean html` without relevant warnings
- [x] run `make spelling` without relevant warnings

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -250,7 +250,10 @@ Create a file ``~/etc/services.d/gitea.ini`` for the service ...
 .. code-block:: ini
 
   [program:gitea]
-  command=%(ENV_HOME)s/gitea/gitea web
+  directory=%(ENV_HOME)s/gitea
+  command=gitea web
+  startsecs=30
+  autorestart=yes
 
 .. include:: includes/supervisord.rst
 
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -38,7 +38,10 @@ We need a database:
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ mysql -e "CREATE DATABASE ${USER}_gitea"
+  [isabell@stardust ~]$ mysql --verbose --execute="CREATE DATABASE ${USER}_gitea"
+  --------------
+  CREATE DATABASE isabell_gitea
+  --------------
   [isabell@stardust ~]$
 
 We can use the uberspace or your own domain:
@@ -56,10 +59,10 @@ Check current version of Gitea at releases_ page.
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ VERSION=1.14.4
+  [isabell@stardust ~]$ VERSION=1.16.3
   [isabell@stardust ~]$ mkdir -p ~/gitea
   [isabell@stardust ~]$ wget -O ~/gitea/gitea https://github.com/go-gitea/gitea/releases/download/v${VERSION}/gitea-${VERSION}-linux-amd64
-  --2020-06-01 21:00:31--  https://github.com/go-gitea/gitea/releases/download/v1.14.4/gitea-1.14.4-linux-amd64
+  --2020-06-01 21:00:31--  https://github.com/go-gitea/gitea/releases/download/vX.Y.Z/gitea-X.Y.Z-linux-amd64
   Resolving github.com (github.com)... 140.82.118.3
   Connecting to github.com (github.com)|140.82.118.3|:443... connected.
   HTTP request sent, awaiting response... 302 Found
@@ -114,7 +117,8 @@ If the verification is fine, we get a ``gpg: Good signature from "Teabot <teabot
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ chmod u+x ~/gitea/gitea
+  [isabell@stardust ~]$ chmod u+x --verbose ~/gitea/gitea
+  mode of '/home/isabell/gitea/gitea' changed from 0664 (rw-rw-r--) to 0764 (rwxrw-r--)
   [isabell@stardust ~]$
 
 
@@ -142,6 +146,7 @@ Create a custom directory.
 The minimum set of ``~/gitea/custom/conf/app.ini`` is:
 
 .. code-block:: ini
+  :emphasize-lines: 3
 
   [server]
   HTTP_PORT = 9000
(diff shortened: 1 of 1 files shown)


source/guide_gitea.rst
```diff
@@ -56,10 +62,10 @@ Check current version of Gitea at releases_ page.
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ VERSION=1.14.4
+  [isabell@stardust ~]$ VERSION=1.16.3
   [isabell@stardust ~]$ mkdir -p ~/gitea
   [isabell@stardust ~]$ wget -O ~/gitea/gitea https://github.com/go-gitea/gitea/releases/download/v${VERSION}/gitea-${VERSION}-linux-amd64
-  --2020-06-01 21:00:31--  https://github.com/go-gitea/gitea/releases/download/v1.14.4/gitea-1.14.4-linux-amd64
+  --2020-06-01 21:00:31--  https://github.com/go-gitea/gitea/releases/download/vX.Y.Z/gitea-X.Y.Z-linux-amd64
```
noave, line 68: I guess its ok to shorten all the output above `Saving to:` here with `[...]`

noave:
I am still unsure about the verification part. We normally don't do this in other guides (maybe just because its not provided in the repos). I guess we should split it in a separate `Verifying (optional)` part after the downloading, what do you think?

source/guide_gitea.rst
```diff
@@ -114,7 +120,8 @@ If the verification is fine, we get a ``gpg: Good signature from "Teabot <teabot
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ chmod u+x ~/gitea/gitea
+  [isabell@stardust ~]$ chmod u+x --verbose ~/gitea/gitea
```
noave, line 123: same here, you can remove the `--verbose` to keept it simple

source/guide_gitea.rst
```diff
@@ -200,112 +191,44 @@ For more information about the possibilities and configuration options see the G
...
-
-.. code-block:: ini
-
-  [markup.asciidoc]
-  ENABLED = true
-  FILE_EXTENSIONS = .adoc,.asciidoc
-  RENDER_COMMAND = "asciidoctor -e -a leveloffset=-1 --out-file=- -"
-  IS_INPUT_FILE = false
+  [repository]
+  DEFAULT_BRANCH = main
```
noave, line 195: thanks for doing these updates, but still, I think this configuration part is too much confusing and we should reduce it somehow more. Its hard to explain what I have in mind so I will try to commit it by myself :-)

EV21:
* We may make a `.. note::` about that and make it much shorter without all the output. 
* Or we move that big part to the end of the guide and make a reference it in a note.

source/guide_gitea.rst
```diff
@@ -355,6 +309,198 @@ To update do:
 * Start the application ``supervisorctl start gitea``
 * Check if the application is running ``supervisorctl status gitea``
 
+You can also automate the update by using the ``gitea-update`` script.
+
+Create ``~/bin/gitea-update`` with the following content:
```
noave, line 314: there is still some `isabell` etc in here that needs to be replaced. I am unsure if this shouldn't be externalised in a specific repo. We have scripts like this in some guides and I find it useful and valuable, but we might need a better solution for it in the future. For now I think its ok to have it here and thanks for the approach.

source/guide_gitea.rst
```diff
@@ -340,13 +264,43 @@ Installed files and folders are:
 * ``~/etc/services.d/gitea.ini``
 * ``~/.gem/ruby/2.7.0/*/asciidoctor*``, if AsciiDoctor is installed.
```
noave, line 265: this should be moved downward now

noave (changes requested):

noave:
> * We may make a `.. note::` about that and make it much shorter without all the output.
that might work, may you try?

source/guide_gitea.rst
EV21: to keep it short!

EV21:
> > * We may make a `.. note::` about that and make it much shorter without all the output.
> 
> that might work, may you try?

Have a look at the current deploy preview. - As the commands (`formatted like this`) are to long so they don't fit into the `..note::` box and a code box inside the note box is also not possible (as of my knowledge), I just separate the three commands without output. - Do you have another idea for this?

source/guide_gitea.rst
```diff
@@ -38,7 +41,10 @@ We need a database:
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ mysql -e "CREATE DATABASE ${USER}_gitea"
+  [isabell@stardust ~]$ mysql --verbose --execute="CREATE DATABASE ${USER}_gitea"
```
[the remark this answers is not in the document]
noave: > I also do this to get visual feedback [...] My idea was that the user sees the transformation from $USER to isabell.
yes, I understand that and I also liked that but...

> It seems to me like you are comparing apples and oranges as KISS and short visual feedback or interaction usecases are different things.

... its more like that I would also like to keep the commands simple and readable. I think 90% of users will just copy and paste the, but keeping it short and reduced to the necessary will maybe enforce them to understand and reuse them  by their own :slightly_smiling_face: 

However, you are also correct, there is no perfect golden way.

> There are some other guides (Nextcloud, HedgeDoc, ...?) with the same thing

didn't know yet :smile:  Actually I would like to enforce more consistency within the guides, but we need to find the golden middle grounds for some style options first.

noave:
> Do you have another idea for this?

I just created a separated step now, will push when Github is up again for pushed :wink: 

image

noave:
ok, lets merge finally :rocket:

PR #1223, opened 2022-03-15
[mailman] configure max SMTP recipients limit

abaumg:
The Uberspace SMTP servers accept max 100 recipients per mail. This PR updates the Lab guide for Mailman2 and Mailman3 to reflect these limits.


noave:
thanks!

PR #1224, opened 2022-03-15
[mailman3] add migration from Mailman 2

abaumg:
Add migration instructions from Mailman 2 to Mailman 3

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -327,7 +327,7 @@ After we have adjusted our configuration file, we need to compile and configure 
 
 .. note:: In case you get an error message when like ``django.core.exceptions.ImproperlyConfigured: SQLite 3.8.3 or later is required (found 3.7.17).``, when running those `manage.py` commands, then you need a newer version of the sqlite library. Install ``pysqlite3-binary`` and create a symlink in order to override ``sqlite3``:
 
- .. code :: bash
+ ::
 
   [isabell@stardust ~]$ pip3.8 install --user pysqlite3-binary
   [...]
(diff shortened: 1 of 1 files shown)


source/guide_mailman-3.rst
```diff
@@ -462,20 +462,78 @@ Updates
...
+Otherwise, update the ``.qmail`` files to use ``qmail-lmtp`` instead of ``vdeliver``:
+
+::
+
+ [isabell@stardust ~]$ echo "|/home/isabell/bin/qmail-lmtp 8024 1 isabell.local.uberspace.de" | tee ~/.qmail-foo-list{,-admin,-bounces,-confirm,-join,-leave,-owner,-request,-subscribe,-unsubscribe}
+
+
+Remove Mailman 2
+----------------
+First, remove the supervisord configuration and stop the service:
```
noave, line 512: acutally, the guide was changed: https://github.com/Uberspace/lab/commit/aa0d26dde4f8e0fde55ccaeb0dea10e03ce59580 to use cronjobs instead of the service, that should at least be mentioned here

noave:
thank you for this, its a good thing to have this on board :slightly_smiling_face: But because the mailman3 guide is already quite huge and complex, and this is not directly connected to a fresh installation of mailman3, we should move this to a separate page and link it from within the mailman3 guide. 

This also opens up the possibility to become more verbose. I left some comments in the code but I would generally say, we can expose more single steps in more detail. Do you like to work on this further? If not, of course, that's ok too and I can do the rest as well.

luto:
Thanks for your guide! We would like to keep the list of open pull requests clean to improve the overview. Since this has been dormant for almost a year, I'm closing this now. Once you find time to look at the comments, feel free to submit it again.

PR #1226, opened 2022-03-19
Added forgotten word

jammon:


noave:
thanks, but please, for the future, _always_ use the `[guide_name]` prefix in commit messages

PR #1233, opened 2022-03-23
[umami] fix build error

thomas-a-johnson:
Fixing build SWC build error: https://nextjs.org/docs/messages/failed-loading-swc

### DIFF
--- a/source/guide_umami.rst
+++ b/source/guide_umami.rst
@@ -87,7 +87,7 @@ Now you can create the production build:
  [isabell@stardust umami]$ npm run build
  [isabell@stardust umami]$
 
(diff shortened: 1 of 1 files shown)


noave:
> If the build process fails with a "Failed to load SWC binary" you need a `.babelrc` file in your umami directory with the following content ``{"presets": ["next/babel"]}``

this pretty much sounds like the guide itself has to be updated, not like this only sometimes happens?

BTW, please follow the guideline for commit messages: https://github.com/Uberspace/lab/blob/main/STYLE.md#add-your-changes-to-the-uberspace-lab

PR #1247, opened 2022-04-16
Specify python version 3.8

korakinos:
At least one dependency of searx doesn't work with python version below 3.8 anymore, namely Flask, see here:
https://flask.palletsprojects.com/en/2.1.x/changes/
https://github.com/pallets/flask/pull/4335


noave:
thanks, please in the future use the commit message with a guide tag like `[searx] specify python version...`

korakinos:
Will do!

PR #1250, opened 2022-04-22
Update guide_znc.rst

no-one:
Use "warning" instead of "note" (as above in the certificate part).


noave:
We don't currently have a good general solution on how to handle this with these replacements in the lab, but it should at least be compliant within a guide, thanks.

PR #1251, opened 2022-04-23
Update guide_seafile.rst

fredl99:
Added version to `pip3 install Pillow` because `seahub.sh` will not start when a newer package is installed. 
Beware of blindly initiated updates! This way I was caught by this problem. Took me two days to figure it out.

### DIFF
--- a/source/guide_seafile.rst
+++ b/source/guide_seafile.rst
@@ -81,7 +81,7 @@ Install the required python libraries.
 
 ::
 
- [isabell@stardust seafile]$ pip3 install --upgrade pip Pillow pylibmc captcha jinja2 sqlalchemy==1.4.3 django-pylibmc django-simple-captcha python3-ldap mysqlclient --user
+ [isabell@stardust seafile]$ pip3 install --upgrade pip Pillow==8.2.0 pylibmc captcha jinja2 sqlalchemy==1.4.3 django-pylibmc django-simple-captcha python3-ldap mysqlclient --user
  [isabell@stardust seafile]$
 
 Create databases


noave:
thanks, do you have any insights if this problem will be fixed in seahub? Because if we lock a specific pillow version here, this might cause problems in the future

fredl99:
Not at all, I'm sorry. I'm neither a python pro nor do I have any insight on seafile.
I was already glad I could fix that issue by replacing some modules from a backup. That was the last idea I could try.

korakinos:
From my own experiments, this has less to do with the version of Pillow than with the version of Python. With `python3.8` and `pip3.8`, specifying the version of Pillow is not necessary.

To be a bit more precise, when calling seahub as `seahub.sh start-fastcgi` one gets an error message involving Pillow and "PyIndex_Check":

```
"ImportError: /home/korakin1/.local/lib/python3.6/site-packages/PIL/_imaging.cpython-36m-x86_64-linux-gnu.so: undefined symbol: PyIndex_Check"

```

Turns out, PyIndex_Check was only defined as a function in Python 3.8, which makes its use in Pillow fail with 3.6:

https://github.com/python/cpython/issues/77919
https://github.com/python/cpython/commit/5cbefa99198729a1d4e93d93f890c066039ee1d2

In fact, I was just meaning to create a pull request from my fork of the guide when I found this one:

https://github.com/korakinos/lab/tree/update-seafile

korakinos:
Here is the pul request from my fork:

https://github.com/Uberspace/lab/pull/1260

fredl99:
Thanks for the explanation @korakinos 

As I stated, I'm no python expert at all. I've followed the lab's tutorial and got a beautifully running seafile-server until some modules were updated. Without any clue the best I could do to revert them. That's how I came to my conclusion.

BTW: After your explanation I just checked the default python3 version at uberspace7: Python 3.6.8

As I understand your message, the better solution would be to run all commands with a newer version than the default.

korakinos:
> As I stated, I'm no python expert at all. 

Me neither! I also just tried to follow the tutorial and figured out a way to make it work.

> BTW: After your explanation I just checked the default python3 version at uberspace7: Python 3.6.8 As I understand your message, the better solution would be to run all commands with a newer version than the default.

Yes. 3.6 is not supported anymore.

https://peps.python.org/pep-0494/
https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions

> BTW2: I don't use `start-fastcgi`. With `start` the error message is much shorter. Something like "didn't start". Not much help.

Yeah, it's really not obvious that `start-fastcgi` has better error messages! If I hadn't found this trick on some forum (forgot where), I would have been pretty lost…

noave:
because the guide has been removed anyway, I will close this as well https://github.com/Uberspace/lab/commit/a2e32055a0892237db99aff20fdf94f6458e9cea

PR #1253, opened 2022-04-28
[ghost] switch to ghost-cli's `ghost update`

### LINKED ISSUE #1243: [ghost] Update Script is not working without `browser_download_url`
The following line in the [update script](https://lab.uberspace.de/guide_ghost.html#update-via-script) is not working anymore:
```
CURRENT_GHOST_DOWNLOAD=$(curl -s https://api.github.com/repos/TryGhost/Ghost/releases/latest | grep browser_download_url | cut -d '"' -f 4)
```

`browser_download_url` doesn't exist in the API response of https://api.github.com/repos/TryGhost/Ghost/releases/latest. It looks like GitHub changed their API response, and `zipball_url` might be a suitable replacement. Downloading this file reveals, though, that the repository is now nested inside another folder with a dynamically generated name.

I didn't have the time yet to fix the update script, but wanted to make you aware of the problem. Thank you!

SalocinHB:
closes #1243

### DIFF
--- a/source/guide_ghost.rst
+++ b/source/guide_ghost.rst
@@ -139,19 +139,15 @@ Configuration
 Change network interface
 ------------------------
 
-You need to adjust your ``~/ghost/config.production.json`` to use the right
-network interface. Find the following block and change host ``127.0.0.1`` to
-``0.0.0.0``:
+You need to change the network interface from ``127.0.0.1`` to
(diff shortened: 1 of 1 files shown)


SalocinHB:
Using node 16 instead of 12 because the latest Ghost version dropped Node 12 support.

PR #1260, opened 2022-05-06
[seafile] update guide

korakinos:
The current version of the Seafile guide doesn't work because of incompatibilities with python3.6 and the system requirement of `memcached` and `libmemcached-devel` not being satisfied on Uberspace.

### DIFF
--- a/source/guide_seafile.rst
+++ b/source/guide_seafile.rst
@@ -89,8 +89,10 @@ Install the required python libraries.
 
 ::
 
- [isabell@stardust seafile]$ pip3.8 install --upgrade pip Pillow pylibmc captcha jinja2 sqlalchemy==1.4.3 django-pylibmc django-simple-captcha python3-ldap mysqlclient --user
+ [isabell@stardust seafile]$ pip3.8 install --upgrade pip Pillow captcha jinja2 sqlalchemy==1.4.3 django-simple-captcha python3-ldap mysqlclient --user
  [isabell@stardust seafile]$
+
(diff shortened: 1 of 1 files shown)


pascoda:
I worked through the guide to check if it follows our guidelines, and if it works.

Unfortunately, starting the service fails due to missing `glibc 2.25`. this is also an issue e.g. for rocket.chat #1272 and it seems like there is no quick fix for it.

korakinos:
> I worked through the guide to check if it follows our guidelines, and if it works.

Thanks! Just to be clear, you mean with my changes or without?

> Unfortunately, starting the service fails due to missing `glibc 2.25`. this is also an issue e.g. for rocket.chat #1272 and it seems like there is no quick fix for it.

I am guessing you tested with Seafile 9.0.x? I am noticing that I installed 8.0.8 myself, which was still advertised on Seafile's main download page back in April. My install is still working, but I see that Seafile 8.0 is no longer supported since the end of 2021, so recommending it is not a good option for the guide:

https://cloud.seatable.io/dtable/external-links/a85d4221e41344c19566/

pascoda:
> Thanks! Just to be clear, you mean with my changes or without?

both, actually. When I noticed that your guide wasn't working, I also tried the existing one -- and as you figured, using the latest release, 9.0.6, for both attempts.

korakinos:
I see. The upstream documentation says explicitely about Seafile 9.0 that "CentOS 8 is no longer supported" (see link), so with Uberspace running on CentOS 7 I probably didn't even try installing it back then. And I completely missed that Seafile 8.0 is not supported anymore, sorry. I am closing this pull request (even though it works for Seafile 8.0) because it could mislead somebody into installing outdated software.

https://manual.seafile.com/deploy/using_mysql/#installing-prerequisites

Issue #1262, opened 2022-05-08
[typo3] Update TYPO3 CMS guide for version 11.5 LTS


tstahn:
TYPO3 version 11.5 was released some time ago, the guide should take that into account as well.

PR #1263, opened 2022-05-08
[typo3] update TYPO3 CMS guide for version 11.5 LTS

tstahn:
Resolves: #1262

### DIFF
--- a/source/guide_typo3-cms.rst
+++ b/source/guide_typo3-cms.rst
@@ -36,12 +36,12 @@ development of the platform.
 Prerequisites
 =============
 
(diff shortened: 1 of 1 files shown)


noave:
thanks, did you try to run this with an even more recent version of php like 8.1? Otherwise its already outdated again in november

tstahn:
Good catch. I did indeed test with all available PHP versions (and yet somehow failed to use the latest supported one in the docs) – I updated the PR accordingly.

pascoda:
PR has been tested, works neatly. thanks for the update!

PR #1264, opened 2022-05-22
workaround for ejabberd-22.05 to build with erlang 21

wertziop:
rebar and reabar3 provided by ejabberd-22.05 are build against erlang 22 and not runnable on older erlang versions. I just tried this quick fix by myself.

Maybe someone could review my proposal. My language skills are limited, both english and reST :)

(Maybe a better solution is to update erlang/OTP?)

### DIFF
--- a/source/guide_ejabberd.rst
+++ b/source/guide_ejabberd.rst
@@ -58,6 +58,21 @@ Run ``./configure --help`` to see all options.
   [isabell@stardust ejabberd-20.04]$ make
   [isabell@stardust ejabberd-20.04]$ make install
 
+.. note:: Since **ejabberd-22.05** ``make`` will output an error, caused by rebar which was build against a newer erlang/OTP version.
+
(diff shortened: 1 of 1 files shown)


noave:
If the guide is currently not working for the newest ejabberd version and this fixes it, you don't need to add it just as a note but may rework the whole guide for the newest version

coderkun:
ejabberd 22.10 works fine with Erlang 24. I can test to compile it with Erlang 25.

wertziop:
sorry, didn‘t found enough time to review my pull request.
please ignore it!

I followed you documentation again and everything is perfectly fine - if you switch erlang/OTP to a version above 22.
when i wrote the work around, i didn‘t know about switching language versions with standard uberspace tools.

sorry again!

hasmaneuroda:
I think it should be mentioned that you can switch language versions by using
`uberspace tools  version use erlang 25`
as it's crucial to use a version > 21, like wertziop already wrote.

PR #1287, opened 2022-06-28
[Nextcloud] Workaround for hanging contacts app

astehlik:
See https://github.com/nextcloud/server/issues/33048 for more background.

In a Nutshell:

* The contacts app makes many requests to load contact images like `/remote.php/dav/addressbooks/users/<user>/<Addressbook>/<contact-uid>.vcf?photo=`
* The `ProxyErrorOverride` config in Uberspace accounts hides the 404 error thrown by Nextcloud and opens the configured `ErrorDocument` from `.htaccess`
* The default error document redirects the user to the Nextcloud start page (Dashboard)
* This means a lot of requests are made at once to load the Nextcloud dashboard which can lead to a completely hung up Nextcloud instance until restarting the PHP service with `uberspace tools restart php`

### DIFF
--- a/source/guide_nextcloud.rst
+++ b/source/guide_nextcloud.rst
@@ -482,6 +482,20 @@ The update to Nextcloud 21.0.1 may fail with the following error message:
 
 To solve the issue, apply the ``apc.enable_cli=1`` step above to your installation.
 
+Contacts app hangs
+------------------
+
+When your instance is hanging when you open the contacts app you might be affected by `Nextcloud issue #33048 <https://github.com/nextcloud/server/issues/33048>`_
+
+As a workaround you can remove or override ``ErrorDocument`` entries in the ``.htaccess`` file:
+
+.. code-block:: bash
+
+ ErrorDocument 403 "Forbidden"
+ ErrorDocument 404 "Not found"
+
+You need to reapply these changes after you installed Nextcloud updates or after you executed the ``maintenance:update:htaccess`` command.
+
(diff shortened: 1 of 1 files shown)


noave:
I really would have liked that this bug is handled in the NC repo, but I guess that could take some time. Because we already have this section of "troubleshooting" its ok to add this workaround here. But I would ask you to remove the automatic update part and just add a hint, that this might have to be checked again after updating

entorb:
Thank you so much, that workaround fixed a headage of mine ;-)

astehlik:
Thank you for your feedback, @noave 

I now replaced this section:

---

You can add these lines to the `~/bin/nextcloud-update` script to make sure the workaround stays active after automated updates:

```bash
 php ~/html/occ maintenance:update:htaccess
 
 # Override default ErrorDocument directives as a workaround for
 # https://github.com/nextcloud/server/issues/33048
 {
   echo ""
   echo 'ErrorDocument 403 "Forbidden"'
   echo 'ErrorDocument 404 "Not found"'
 } >> ~/html/.htaccess
```

---

with a hint. I hope this is OK. Please let me know if I should change anything else.

F3000:
Is this a general bug from Nextcloud which can be fixed in the next updates ?

astehlik:
> Is this a general bug from Nextcloud which can be fixed in the next updates ?

@F3000 The issue filed with Nextcloud is pending for quite some time now :disappointed: 

See: https://github.com/nextcloud/server/issues/33048

PR #1295, opened 2022-08-04
Update guide_postgresql.rst: fix postgresql.ini

patrickbkr:
It's `startsecs` not `startsec`.


noave:
thanks, please use correct `[postgres]` prefix in commit message next time :-)

PR #1299, opened 2022-08-10
Specify where `--remove-prefix` is needed.

debuglevel:
It's a bit ambiguous whether this argument applies to `uberspace web ...` or the couchdb configuration.


SalocinHB:
I'd rather remove the whole sentence. To quote the style guide:

> Don't mention additional document roots. Keep it simple. *Don't use subfolders*. Always use the standard document root ~/html. Always assume the document root is empty.

PR #1303, opened 2022-08-12
babybuddy needs python3.7+ now

whnr:
Assuming that we'll use python 3.9 now.
Then we need to use uswgi with the same  version. Noted that in there.
mzsqlclient module is now always installed


nichtmax:
Thanks!

Issue #1304, opened 2022-08-14
Guide for filebrowser


on4r:
https://github.com/filebrowser/filebrowser

PR #1305, opened 2022-08-14
Add filebrowser guide

on4r:
Closes https://github.com/Uberspace/lab/issues/1304


source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+
+.. note:: You have to be in the same directory as the *filebrowser.db* file which you created before, in order to configurate it or else it will create a new one.
+.. note:: Check your running web backends (``uberspace web backend list``) and choose a free port between 1024 and 65535.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser config set --port 1337
+ [isabell@stardust filebrowser]$ filebrowser config set --address 0.0.0.0
+ [isabell@stardust filebrowser]$ filebrowser config set --baseurl /my_files
+ [isabell@stardust filebrowser]$ filebrowser config set --root /home/<username>/html/my_files
```
EV21, line 69: - We don't need to change the port, the default port `8080` is also working.
```
--port string                        port to listen on (default "8080")
```
- You may add a short info at some point:
 note: changes to the config via CLI can not be done while the server is running as the database access is blocked by the instance.
- Why do you use a subfolder for the `baseurl`? As per style guide you need to keep it simple.
 You even remove the that prefix with the web backend, so it actually makes no sense to me.

This is how I did it:

```bash
cd ~/bin
VERSION=2.22.4
curl --location https://github.com/filebrowser/filebrowser/releases/download/v$VERSION/linux-amd64-filebrowser.tar.gz | tar -xzf - filebrowser
mkdir ~/filebrowser ~/my_files && cd ~/filebrowser
filebrowser config init
filebrowser config set --address 0.0.0.0
filebrowser config set --root $HOME/my_files
filebrowser users add '<username>' '<password>' --perm.admin
uberspace web backend set / --http --port 8080
```

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser users add <username> <password> --perm.admin
+
+Create a new `uberspace web backend`_.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ uberspace web backend set /my_files --http --port 1337 --remove-prefix
```
EV21, line 81: In case you accept my prefix-free suggestion above, this is the way to go:

```suggestion
.. note::

 File Browser is running on port 8080.

.. include:: includes/web-backend.rst
```

EV21 (commented):
# Hi 👋,
thank you for your nice guide for the great app. 🚀
I left a few comments and suggestions.

[deploy preview 👀](https://deploy-preview-1305--uberlab.netlify.app/guide_filebrowser/)

---

Just as an addition because of the nice CLI we may add this to the guide⁉️:

## completion (optional)

```bash
filebrowser completion bash > ~/filebrowser/filebrowser_bash_completion
echo 'source ~/filebrowser/filebrowser_bash_completion' >> .bash_profile
source ~/.bash_profile
```

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+============
+
+Download and extract the `latest release`_
+-------------------
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ cd ~/bin
+ [isabell@stardust bin]$ wget <url_of_linux_amd64_filebrowser>
+ [isabell@stardust bin]$ tar -xzf linux-amd64-filebrowser.tar.gz
```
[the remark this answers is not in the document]
on4r: In order to know the latest version you already have to visit the release page. From there on it's only copy & paste. To add an extra step and define a variable and use an slightly more complicated command afterwards is, at least in my opinion, an inconvenience.  
What i do like however is the suggestion to just extract the binary. 

```suggestion
.. code-block:: console

 [isabell@stardust ~]$ cd ~/bin
 [isabell@stardust bin]$ wget <url_of_linux_amd64_filebrowser>
 [isabell@stardust bin]$ tar -xzf linux-amd64-filebrowser.tar.gz filebrowser
```

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+  .. image:: _static/images/filebrowser.png
+      :align: center
+
+##########
+File Browser
+##########
+
+.. tag_list::
+
+filebrowser_ provides a file managing interface within a specified directory and it can be used to upload, delete, preview, rename and edit your files.
```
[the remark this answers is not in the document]
on4r: 👍

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+ command=%(ENV_HOME)s/bin/filebrowser --database %(ENV_HOME)s/filebrowser/filebrowser.db
+ autostart=yes
+
+Now reread, start and check the service.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ supervisorctl reread
+ [isabell@stardust filebrowser]$ supervisorctl update
+ [isabell@stardust filebrowser]$ supervisorctl status
```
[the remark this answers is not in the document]
on4r: 👍

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+ [isabell@stardust filebrowser]$ filebrowser config set --port 1337
+ [isabell@stardust filebrowser]$ filebrowser config set --address 0.0.0.0
+ [isabell@stardust filebrowser]$ filebrowser config set --baseurl /my_files
+ [isabell@stardust filebrowser]$ filebrowser config set --root /home/<username>/html/my_files
+
+Create your admin account.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser users add <username> <password> --perm.admin
```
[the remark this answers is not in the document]
on4r: 👍

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,129 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ touch ~/etc/services.d/filebrowser.ini
+
+.. code-block:: ini
+
+ [program:filebrowser]
+ command=%(ENV_HOME)s/bin/filebrowser --database %(ENV_HOME)s/filebrowser/filebrowser.db
+ autostart=yes
```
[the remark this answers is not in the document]
on4r: 👍

source/guide_filebrowser.rst
on4r: - `8080` is only working if there are no other backends running which use this port. I think it's good practice to first check what's already running and used and then choose a port. 
- in my personal setup i have many different services and applications running at different urls, thats why i use `baseurl`. About the prefix thou... yeah, it doesn't make any sense. I guess we can keep it simple here and go with your suggestion.

EV21: Every uberspace (V7) user has its separate and conflict free port network. As I have checked it looks like the port 8080 is free to use at a fresh uberspace instance, so I don't see a problem. - You can't even reach another users network services without opening extra firewall ports with the uberspace command line tool. - Only these open ports (we don't need in this case) will be proxied to the host network and then are open to the galaxy wide internet. 😉
Have a look at https://manual.uberspace.de/background-network/

The lab guides and the uberspaces are ment to be used as single app containers, so for each service you are running you should run an extra uberspace (containers). So port conflicts with other of you own services should not be a thing in context of these lab guides.

source/guide_filebrowser.rst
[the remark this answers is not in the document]
EV21: If you like to use the separate commands, don't forget to remove the archive afterwards.

```
...
[isabell@stardust bin]$ rm linux-amd64-filebrowser.tar.gz
[isabell@stardust bin]$
```

source/guide_filebrowser.rst
on4r: I wasn't talking about conflicts with other users ports but with my own running applications. But thanks for the reminder. 
I read somewhere in your docs about putting each service onto its own uberspace account. While i don't think it's a practicable approach for the default user, i see why we should stay in context of the lab guides.

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser users add '<username>' '<password>' --perm.admin
+
+Create a new `uberspace web backend`_.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ uberspace web backend set / --http --port 8080
```
EV21, line 77:
```suggestion
File Browser runs on port ``8080`` by default.

.. include:: includes/web-backend.rst
```

The include function will insert the following snippet. It's a common practice to use these snippets and just mention the port that needs to be used.

```rst
To make the application accessible from the outside, configure a `web backend <webbackend_>`_:

::

  [isabell@stardust ~]$ uberspace web backend set / --http --port <port>
  Set backend for / to port <port>; please make sure something is listening!
  You can always check the status of your backend using "uberspace web backend list".
  [isabell@stardust ~]$

.. _webbackend: https://manual.uberspace.de/web-backends.html
```

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,122 @@
...
+
+ [isabell@stardust filebrowser]$ mkdir ~/my_files
+
+At this point you should check if everything is working as expected by starting manually.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser --database $HOME/filebrowser/filebrowser.db
+
+File Browser should now be running on port ``8080``.
```
EV21, line 91: At this point we already have set up the web backend, so I think the following suggestion is what you actually want to say? - Or maybe you can also show a short part from the console output.

```suggestion
File Browser should now be accessible via https://isabell.uber.space
```

EV21 (approved):
Great, it's almost done. 
The next review should be done by (I think) @SalocinHB
He may have a few more change requests before this guide is ready for publication.

EV21 (approved):
The Uberspace team (I am not one of them) seems to be busy, the other pull requests have also been waiting for care for weeks. So be patient, you are not forgotten.

CC: @SalocinHB @nichtmax @noave @kimdiallo ...

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+License
+=======
+
+File Browser is licensed under the `Apache License 2.0`_
+
+Installation
+============
+
+Download and extract the `latest release`_
+------------------------------------------
```
noave, line 40: just use simple text without headlining here

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ cd ~/bin
+ [isabell@stardust bin]$ curl --location <url_of_linux_amd64_filebrowser> | tar -xzf - filebrowser
+
+Initialize a new database
+-------------------------
```
noave, line 48: just use simple text without headlining here

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+Installation
+============
+
+Download and extract the `latest release`_
+------------------------------------------
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ cd ~/bin
+ [isabell@stardust bin]$ curl --location <url_of_linux_amd64_filebrowser> | tar -xzf - filebrowser
```
noave, line 45: say exactly which url should be used here, typically the url stays the same and one only needs to replace the version number

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+=============
+
+.. note:: You have to be in the same directory as the *filebrowser.db* file which you created before, in order to configurate it or else it will create a new one.
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser config set --address 0.0.0.0
+ [isabell@stardust filebrowser]$ filebrowser config set --root $HOME/my_files
+
+Create your admin account.
```
noave, line 67: state that username and password needs to be replaced

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser users add '<username>' '<password>' --perm.admin
+
+File Browser runs on port ``8080`` by default.
+
+.. include:: includes/web-backend.rst
+
+Create your file folder (the one you specified via the ``--root`` option).
```
noave, line 77: better put this part at the beginning of configuration and reuse the chosen path afterwards

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ mkdir ~/my_files
+
+At this point you should check if everything is working as expected by starting manually.
```
noave, line 83: make a section "testing" here and state that the command should be stopped after testing

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: console
+
+ [isabell@stardust filebrowser]$ filebrowser --database $HOME/filebrowser/filebrowser.db
+ 2022/10/03 19:31:02 Listening on [::]:8080
+
+File Browser should now be accessible via `https://isabell.uber.space`.
```
noave, line 90: use double ticks for inline code here

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
+.. highlight:: console
+
+.. author:: on4r <https://github.com/on4r>
+
+.. categorize your guide! refer to the current list of tags: https://lab.uberspace.de/tags
+.. tag:: web
+.. tag:: file-storage
+.. tag:: self-hosting
```
noave, line 8: add lang-go tag

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,121 @@
...
+ [isabell@stardust ~]$ cd ~/bin
+ [isabell@stardust bin]$ curl --location <url_of_linux_amd64_filebrowser> | tar -xzf - filebrowser
+
+Initialize a new database
+-------------------------
+
+.. code-block:: console
+
+ [isabell@stardust bin]$ cd ~
+ [isabell@stardust ~]$ mkdir filebrowser && cd filebrowser
```
noave, line 53: dont cd to home but use `~/filebrowser` here

noave:
nice simple tool, thanks, I could succesfully run it on uberspace 7.14. I just added some style change requests. Sorry for delayed reviewing and thanks to intial review @EV21

noave:
maybe there is a way to show the correct user quota here, but its not so important for the guide right now

image

source/guide_filebrowser.rst
on4r: It looks like we can now easily link to the latest release assets, see: [Linking to releases](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases) (at the end of the site)

on4r:
Thanks for the review @noave i updated the guide according to your annotations.

source/guide_filebrowser.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Installation
+============
+
+Download and extract the `latest release`_
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ cd ~/bin
+ [isabell@stardust bin]$ curl --location https://github.com/filebrowser/filebrowser/releases/latest/download/darwin-amd64-filebrowser.tar.gz | tar -xzf - filebrowser
```
EV21, line 45:
```suggestion
 [isabell@stardust bin]$ curl --location https://github.com/filebrowser/filebrowser/releases/latest/download/linux-amd64-filebrowser.tar.gz | tar -xzf - filebrowser
```

luto:
Thank you! 🎉

PR #1314, opened 2022-09-07
[Gitea] fix stop signal

EV21:
The HUP signal may result into forking the process into the backgroud, so it breaks the daemon when updating your instance.
Restarting or reloading are failing as the port is still in use by the forked instance.
https://github.com/alpinelinux/aports/commit/8a8c116a951f8c8a53be667e4697a76c97db93dc

Quick test:
```
bash -c "$(wget -q -O - https://sh.ev21.de/uberspace/gitea-installer.sh)" _ use 1.16.8
```
```
gitea-update
```

With this fix updates should run without any headaches.

[👀 deploy preview](https://deploy-preview-1314--uberlab.netlify.app/guide_gitea)

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -168,9 +168,6 @@ Create a config file ``~/gitea/custom/conf/app.ini`` with the content of the fol
   MAILER_TYPE = sendmail
   FROM        = isabell@uber.space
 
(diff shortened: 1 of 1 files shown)


nichtmax:
Thanks! 💞

PR #1323, opened 2022-09-26
[mattermost] add update instructions for v6.4 and newer

fabolhak:
see also https://docs.mattermost.com/install/self-managed-changelog.html#id84


noave:
sorry for reviewing this so late, could you put this in a CLI codeblock with `$ mysql -e "(statement)"` like in the other code examples?

PR #1326, opened 2022-10-03
[ghost] add update (v4...v5) instructions

putzwasser:


noave:
as we removed the ghost guide anyway I unfortunately will close this PR

Issue #1330, opened 2022-10-19
[FreshRSS] branch renamed; new update procedure and some improvements


rainerrose:
# new branch

Meanwhile the name of the branch has changed to `edge` . Some names and pathes have to be changed.

for example

    ln -s FreshRSS-master/p/ html

to

    ln -s FreshRSS-edge/p/ html

# update

The exclude of `do-install.txt` is no longer needed; see commit cc6c529562ef5751133d97e3fec067c0072f215b on
https://github.com/FreshRSS/FreshRSS/commits/edge/docs/en/admins/04_Updating.md

# logging / cronjob

    */10 * * * * php /var/www/virtual/$USER/FreshRSS-master/app/actualize_script.php > $HOME/logs/FreshRSS.log 2>&1

This command overwrites the logfile every run, should be ` >> $HOME/logs/FreshRSS.log 2>&1 `

Valid for version `FreshRSS 1.20.1-dev`

rainerrose:
Have a look to my PR https://github.com/Uberspace/lab/pull/1331 please

PR #1331, opened 2022-10-19
bug fix issue #1330 from fork

rainerrose:
* new name of branch fixed
* cronjob improvements
* update  improvements


noave:
thanks for this, unfortunately there was another commit with this fix now already done (sorry for late review). But anyway, I would say we should go with releases here from https://github.com/FreshRSS/FreshRSS/releases and not with rolling release, could you imagine to fix the guide based on this?

Issue #1340, opened 2022-11-02
would you like me to contribute an ntfy howto?


jpmens:
[ntfy](https://ntfy.sh) sends push notifications to your phone or desktop via PUT/POST. It's an open source  utility, and I've just set it up on a uberspace. I'd be willing to submit an _Uberlab guide_ if desired.

luto:
sure! :)

PR #1341, opened 2022-11-02
Guide for ntfy

jpmens:
This adds a new Lab guide for ntfy.

closes https://github.com/Uberspace/lab/issues/1340


FriedrichFroebel:
It seems like the `/home/jane` parts should be `/home/isabell` instead, as this is the user from the console prompts.

jpmens:
of course, thanks. Fixed and pushed.

PR #1342, opened 2022-11-02
[bookstack] Note of file migration in production.

paradonym:
Some updates seem to need a non-default answer after installing. So I added a note after the code block.


noave:
thanks for the info, however, the codeblock you refer to should be splitted anyway because there are three commands used in a single line. Could you image to take care for that and add your comment at the correct codeblock?

paradonym (commented):
whatever I said before. Github needs so many comments...

Anyway, this is what @noave probably wanted. Removed the sample output because I don't know which exact command was responsible for it.

luto:
Thank you! :)

PR #1436, opened 2022-11-21
[restic] use full path in backup script

SaschaMann:
When following the guide as it is, the cron logs showed the following error:

```
/home/<user>/restic_backup.sh: line 20: restic: command not found
```

This change made the script work in cron for me.

I don't know cron well enough to determine the proper way, so please correct me if there's a better solution.

### DIFF
--- a/source/guide_restic.rst
+++ b/source/guide_restic.rst
@@ -117,7 +117,7 @@ It is recommended to create a script to automate the backups. Please edit this t
   FILES='/var/www/virtual/isabell/html/ /home/isabell/myImportantDocument.md'
 
   # Run restic backup
-  restic -r b2:bucketname:path/to/repo backup $FILES
+  ~/bin/restic -r b2:bucketname:path/to/repo backup $FILES
 
   #set +x # Stop printing
 


SalocinHB:
> When following the guide as it is, the cron logs showed the following error:
> 
> ```
> /home/<user>/restic_backup.sh: line 20: restic: command not found

This should work as `/home/<user>/bin` is in the default PATH on Uberspaces. Did you modify your PATH variable (e. g. in `~/.bashrc`)?

EV21:
> This should work as /home/<user>/bin is in the default PATH on Uberspaces.

FYI:
Crontab (currently) does not care about this default, so it does not know the users `PATH` setup, so you need the full path in this case!

```
For
*/1 * * * * echo $PATH

the output is:
/usr/bin:/bin
```

I also ran into the same issue when setting up the daily `nextcloud-update` with cron because `restorecon` was not found without the full path `/usr/sbin/restorecon`.

EV21 (approved):

luto:
Thank you! 🎉

PR #1440, opened 2022-11-23
[radicale] start service before web server config

tvogel:
... also, put "Start Service" before "Configure web server" such that the output there, reporting a "listening: PID...", is actually correct.


luto:
Thank you! 🎉

PR #1447, opened 2022-11-27
[firefly-iii] Added note how to solve endles redirect issue for `/login`

stevleibelt:
Added a note in the section `Finishing installation`. I ran into this issue and debugged it to figure out the Solution `RewriteBase /`.

My two cents, I have a slightly different setup with:

* subdomain like `https://my.isabell.uber.space` instead of `https://isabell.uber.space`
* `my.isabell.de` as softlink to `data/de/isabell/my/public` instead of `firefly_iii/public html`


SalocinHB:
This is explained in the manual. Lab guides should always assume the default setup.

PR #1449, opened 2022-12-02
Code server update path structure

Tschoepler:
Updated tar names according to current release and added /bin/ to path of the binary as it was introduced.


noave:
thanks

Issue #1345, opened 2022-11-08
[pleroma] guide seems to be broken


pascoda:
we were informed, and could reproduce, that setting up pleroma currently does not work (neither with new, nor with older versions). 

@taddydevil - can you help?

taddydevil:
Do you have _some_ more informations for me? :) 
Error in compiling, dependencies or something else?

taddydevil:
**Oo.** 

Please try Erlang 22 @custompyramidfellow . On my working instance with Pleroma 2.4.0 I can compile prometheus_ex.

On my other U7 I can't compile idna, maybe a memory issue. But what I saw in your logs is you have no issues with idna, so this should be not a problem for you.

taddydevil:
Nice! 

But how?
I am pretty sure that Pleroma once worked with Erlang version 24 - Uberspace itself has adapted the guide to 24, they do not do that without testing.
However, there must be a reason why I continue to use 22 on my production instance.
I guess 22 is safe until the Uberspace team removes the libraries. Similar to PHP7 - and every damn Pleroma update must be checked for compatibility for that time. 

This issue should remain open until 24 is running again.

Never touch a running version :p

taddydevil:
@pascoda I can reproduce that pleroma 2.4.4 / the compiling of prometheus_ex works only with Erlang 22. The guide should be adapted to Erlang 22 and everything is fine - can you do that?

Background:

There is a active issue in prometheus_ex repo already with a fix:
https://github.com/deadtrickster/prometheus.ex/issues/48

> Elixir 1.14 changed Kernel.Utils.defdelegate/2 into Kernel.Utils.defdelegate_each/2

There are two ways to "fix" the compiling.

* Change erlang to 22 temporarily until fix https://github.com/deadtrickster/prometheus.ex/pull/47 is merged and copied to pleroma git
**my recommendation because pleroma developing is slooooow**
* Change mix.exs. Search for `prometheus_ex`, exlude the pleroma source and add the original source with fixed branch `{:prometheus_ex, git: "https://github.com/lanodan/prometheus.ex", branch: "fix/elixir-1.14", override: true},
      {:prometheus_plugs, "~> 1.1"},`
**Tested and works without any issues. But local edited configs from a replicated repo are not cool imho. This will blew up the lab guide.** 
```
      ...
      {:prometheus, "~> 4.6"},
#      {:prometheus_ex,
[...]

PR #1465, opened 2022-12-30
[plerma] remove broken disclaimer, update to 2.5.0

taddydevil:
closes https://github.com/Uberspace/lab/issues/1345 (there are no issues anymore, fixed with 2.4.5)

Update Erlang to 25. Tested with Pleroma 2.5.0 with Erlang 25 on my own instance.


luto:
nice, thank you!

PR #1468, opened 2023-01-03
add --locale= to guide_postgresql.rst

machinaeXphilip:
adds german locale to db cluster init to match behaviour of fresh uberspace instances.


machinaeXphilip:
closes #1469

SalocinHB:
New Uberspaces do not have a German locale set. Most likely, your SSH client set this locale using [SendEnv](https://man7.org/linux/man-pages/man5/ssh_config.5.html)

PR #1473, opened 2023-01-12
[vaultwarden] Improve updateability

rikroe:
This PR moves both vaultwarden's `/data` directory as well as the `.env` file out of the `/output` dictionary, which gets overwritten with every update.

Therefore, one only has to stop vaultwarden, extract the latest update and is able to start it again.


rikroe:
cc @pxlfrk @knhash

knhash (approved):
Looks good to me.

luto:
nice, thanks :)

PR #1475, opened 2023-01-16
[Forgejo] new guide

redwerkz:
https://forgejo.org/


source/guide_forgejo.rst
```diff
@@ -0,0 +1,519 @@
...
+  #!/usr/bin/env bash
+
+  APP_NAME=Forgejo
+  FORGEJO_LOCATION=$HOME/forgejo/forgejo
+  TMP_LOCATION=$HOME/tmp
+  GPG_KEY_FINGERPRINT=7C9E68152594688862D62AF62D9AE806EC1592E2
+
+  ORG=forgejo # Organisation or GitHub user
+  REPO=forgejo
+  GITHUB_API_URL=https://codeberg.org/$ORG/$REPO/releases/latest
```
EV21, line 294: I suggest to remove the update script for now. Maybe we can add an adjusted version later when this guide is published with the basic manual steps.

Since there is no `latest` tagging system for versions on Forgejo/Gitea like we have on GitHub this does not work.
This latest link redirects to the last non-pre-release, so it could also be a maintenance release for a lower major version. 
You have to get the releases via `https://codeberg.org/api/v1/repos/forgejo/forgejo/releases` (a `latest` endpoint does not exist) and check for the version tags and also the `prerelease: false` state.
This is much more complicated so the `get_latest_version` function has to be changed a lot with check loops etc.

EV21 (approved):

luto:
Thank you for the guide and thanks @EV21 for the review! I just tested it and it works great! Sorry about the delay :(

PR #1476, opened 2023-01-16
Update guide_directus.rst

MrMacStripe:
The recompilaion of the argon2 module does not work in the way provided.

It's necessary to install and recompile the module inside the directus node_modules folder.

Therefore the installation needs to happen before the recompilation.


luto:
We went with #1470 because it was submitted earlier. If this is still relevant or better, please ping us again. Thanks!

PR #1477, opened 2023-01-18
Fix Cryptopad documentation

ewolff:
I tried to install Cryptopad as documented and failed. I actually believe the documented installation cannot work. The originally documented way makes the instance think it runs on localhost:3000 - the `httpUnsafeOrigin` which makes it fail. The way described here works.


luto:
Thank you! :)

PR #1478, opened 2023-01-20
Add missing dollar char to crontab line for nextcloud

okoetter:


EV21 (commented):
No, this is an arithmetic expansion `$(( SOME_VARIABLE + SOME_FUNCTION + 5 ))` works fine.

Using the dollar sign for functions may cause wrong results. 

Have a look at this:
https://www.shellcheck.net/wiki/SC2004

PR #1479, opened 2023-01-20
Mask percent sign in crontab for nextcloud

okoetter:


EV21:
Can you give a description for your PR that explains why this is needed? I don't get it.
Again, this is an arithmetic expansion - escaping the operators has no effect in this context. It is also less readable.
The syntax for this makes sure to handle the parameters correctly inside the brackets unlike `expr`-parameters

This PR is partially a duplicate of #1478 because of the `$`-sign (bad practice for arithmetic expansions)

source/guide_nextcloud.rst
```diff
@@ -249,7 +249,7 @@ Add the following cronjob to your :manual:`crontab <daemons-cron>`:
 
 ::
 
- */5  *  *  *  * sleep $(( 1 + RANDOM % 60 )) ; php -f $HOME/html/cron.php > $HOME/logs/nextcloud-cron.log 2>&1
+ */5  *  *  *  * sleep $(( 1 + $RANDOM \% 60 )) ; php -f $HOME/html/cron.php > $HOME/logs/nextcloud-cron.log 2>&1
 
 .. note:: The actual cronjob is preceded by a random sleep of maximum one minute to prevent load peaks every 5 minutes due to simultaneous execution of all cronjobs.
```
EV21, line 254:
```suggestion
 */5  *  *  *  * sleep $(( 1 + RANDOM \% 60 )) ; php -f $HOME/html/cron.php > $HOME/logs/nextcloud-cron.log 2>&1

.. note::
  | The actual cronjob is preceded by a random sleep of maximum one minute to prevent load peaks every 5 minutes due to simultaneous execution of all cronjobs.
  | The ``\%`` is required as crontab replaces ``%`` with new-line characters.
```

EV21 (changes requested):

source/guide_nextcloud.rst
okoetter: Yeah... sorry for the unneeded dollar. At first it didn't work for me so I thought it was the missing dollar while it indeed was the missing backslash before the percent character.

EV21 (approved):

SalocinHB:
Fixed in https://github.com/Uberspace/lab/commit/6445a3dc9389c1ecf67ba07b83ce8687a7dc9b75. 

@okoetter sorry I didn't see your PR earlier.

PR #1482, opened 2023-01-24
[mattermost] Add automatic update script

chief-nerd:
I wrote myself an update script for mattermost and wanted to share it with everyone.

Open:
- Get the latest version of mattermost via a feed, so the version input is not needed.

Tried:
- Reading latest version from website (very cumbersome, but might be possible)
- Read latest version from Github (needs Auth)
- No directory listing via https://releases.mattermost.com/

### DIFF
--- a/source/guide_mattermost.rst
+++ b/source/guide_mattermost.rst
@@ -177,72 +177,74 @@ Automatic Update
 Use the script attached to update Mattermost to the current version. Run the script above the mattermost-root directory.
 
 .. code-block:: shell
-#!/bin/sh
-printf "\n===\nWelcome to updating Mattermost... let us begin!\n===\n"
+	#!/bin/sh
+	printf "\n===\nWelcome to updating Mattermost... let us begin!\n===\n"
 
-printf "\n== Preflight Checks ==\n"
-if [ ! -d "./mattermost" ]; then
-  printf "Error: ./mattermost directory does not exist. Are you running in the correct place?. Exit\n"
-  exit 1
-else 
-	printf "Directory ./mattermost found.\n"
-fi
+	printf "\n== Preflight Checks ==\n"
+	if [ ! -d "./mattermost" ]; then
+	printf "Error: ./mattermost directory does not exist. Are you running in the correct place?. Exit\n"
+	exit 1
+	else 
+		printf "Directory ./mattermost found.\n"
+	fi
 
-database=$(cat ./mattermost/config/config.json | jq '.SqlSettings.DataSource' | grep -o '/.*?' | tail -c +2 | head -c -2)
-if [ -z "$database" ] 
-then
-	printf "\nError: Could not extract the database name from ./mattermost/config/config.json. Maybe the script runs in the wrong directory? Exit 1.\n"
-	exit 2
-else 
-	printf "Using database: $database\n"
-fi
+	database=$(cat ./mattermost/config/config.json | jq '.SqlSettings.DataSource' | grep -o '/.*?' | tail -c +2 | head -c -2)
+	if [ -z "$database" ] 
+	then
+		printf "\nError: Could not extract the database name from ./mattermost/config/config.json. Maybe the script runs in the wrong directory? Exit 1.\n"
+		exit 2
+	else 
+		printf "Using database: $database\n"
+	fi
 
-version=$1
-if [ -z "$version" ] 
-then
-	printf "No manual version given. Check from official website ...\n"
+	version=$1
+	if [ -z "$version" ] 
+	then
+		printf "No manual version given. Check from official website ...\n"
 
-	downloadLink=$(curl -s https://mattermost.com/deploy/ | grep -o '<code class="wrap break-all">.*</code>' | sed 's/\(<code class="wrap break-all">\|<\/code>\)//g')
+		version=$(curl -s https://api.github.com/repos/mattermost/mattermost-server/releases/latest | jq '.tag_name' | tail -c +3 | head -c -2)
 
-	version=$(grep -o 'com/.*/matter' <<< $downloadLink | tail -c +5 | head -c -8)
-	printf "Newest Version detected: v$version\n"
-fi
+		printf "Newest Version detected: v$version\n"
+	fi
 
-printf "\n== Prepare for takeoff ==\n"
-printf "Downloading Version v$version...\n"
-mattermostfile="mattermost-$version-linux-amd64.tar.gz"
-wget https://releases.mattermost.com/$1/$mattermostfile
+	printf "\n== Prepare for takeoff ==\n"
 
-printf "\nExtracting Version ..."
-tar -xf $mattermostfile --transform='s,^[^/]\+,\0-upgrade,'
+	# TODO check if update is needed
 
-printf "\nCreate Backup of filesystem..."
-backupdir="mattermost-back-$(date +'%F-%H-%M')/"
-cp -ra mattermost/ $backupdir
+	printf "Downloading Version v$version...\n"
+	mattermostfile="mattermost-$version-linux-amd64.tar.gz"
+	wget https://releases.mattermost.com/$1/$mattermostfile
 
-printf "\nCreate Backup of database: '$database' ..."
-mysqldump --databases $database > $backupdir/$database.dump.sql
+	printf "\nExtracting Version ..."
+	tar -xf $mattermostfile --transform='s,^[^/]\+,\0-upgrade,'
 
-printf "\n== Take off ==\n"
-printf "\nStop Service - ..."
-supervisorctl stop mattermost
+	printf "\nCreate Backup of filesystem..."
+	backupdir="mattermost-back-$(date +'%F-%H-%M')/"
+	cp -ra mattermost/ $backupdir
 
-printf "\nDelete all old files ..."
-find mattermost/ mattermost/client/ -mindepth 1 -maxdepth 1 \! \( -type d \( -path mattermost/client -o -path mattermost/client/plugins -o -path mattermost/config -o -path mattermost/logs -o -path mattermost/plugins -o -path mattermost/data \) -prune \) | sort | xargs rm -r
+	printf "\nCreate Backup of database: '$database' ..."
+	mysqldump --databases $database > $backupdir/$database.dump.sql
 
-printf "\nApply new files ..."
-cp -an mattermost-upgrade/. mattermost/
+	printf "\n== Take off ==\n"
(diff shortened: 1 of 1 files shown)


chief-nerd:
Added automatic version detection based on their website.

EV21:
Can you please indent your code by at least one space so it gets rendered correctly as a code box.
Have a look at https://deploy-preview-1482--uberlab.netlify.app/guide_mattermost/ 👀

```rst
.. code-block:: shell

  #!/bin/sh
  echo "This is part of the code block"
this is NOT part of the code block
```

---

> * Get the latest version of mattermost via a feed, so the version input is not needed.
> ...
> * Read latest version from Github (needs Auth)

Are you sure? - The current rate limit for anonymous GitHub-API requests is 60 requests per hour (per IP)
This will only be a problem when many people on the same Uberspace host start automating these updates with daily cron jobs.

You can just call
https://api.github.com/repos/mattermost/mattermost-server/releases/latest

You can also have a look for other update scripts in the lab. These are some I wrote:
 - https://lab.uberspace.de/guide_gitea/#automated-updating-by-custom-script
 - https://lab.uberspace.de/guide_hedgedoc/#update-script <- This one has a "are your sure?"-barrier to prevent automatic updates
 - https://github.com/EV21/uberspace-helper/blob/main/grocy-update.sh < a very short one (not in the lab)
 - https://github.com/EV21/uberspace-helper/blob/main/uptime-kuma-update.sh (not in the lab)

Using the website with screen scraping techniques may fail as the website structure changes from time to time.

chief-nerd:
Great points @EV21, thanks. Updated the intent + changed the code

chief-nerd:
Found a way to get the local version through the CLI. This make it useable in a cronjob now. :)

source/guide_mattermost.rst
```diff
@@ -160,12 +171,93 @@ When upgrading to Mattermost 6.4 or newer you need to change the collation of th
...
   [isabell@stardust ~]$
 
+Automatic Update
+-----------------
+
+Use the script attached to update Mattermost to the current version. Run the script above the mattermost-root directory.
+
+.. code-block:: shell
+	#!/bin/sh
+	printf "\n===\nWelcome to updating Mattermost... let us begin!\n===\n"
```
EV21, line 181: You also need a blank line before the content to get this rendered.

```suggestion
.. code-block:: shell

	#!/bin/sh
	printf "\n===\nWelcome to updating Mattermost... let us begin!\n===\n"
```

chief-nerd:
@EV21 this should be now good to go. Let me know if there are more things to consider.

source/guide_mattermost.rst
```diff
@@ -160,12 +171,94 @@ When upgrading to Mattermost 6.4 or newer you need to change the collation of th
...
+-----------------
+
+Use the script attached to update Mattermost to the current version. Run the script above the mattermost-root directory.
+
+.. code-block:: shell
+
+	#!/bin/sh
+	printf "\n===\nWelcome to updating Mattermost... let us begin!\n===\n"
+
+	printf "\n== Preflight Checks ==\n"
```
EV21, line 184: To increase the readability I would not use all this formatting stuff in all the info messages.
All these backslashes and other cosmetic characters make the code less readable.

If you really want this you can use a helper function for such things. You can even work with colors ... (different topic)
```bash
# This is what you are doing
function info_message { echo "\n== $* ==\n" }

# I think this is enough
function info_message { echo "INFO: $*" }

# usage
info_message "Stage 2 seperated"
```

source/guide_mattermost.rst
```diff
@@ -160,12 +171,94 @@ When upgrading to Mattermost 6.4 or newer you need to change the collation of th
...
+	exit 1
+	else 
+		printf "Directory ./mattermost found.\n"
+	fi
+
+	database=$(cat ./mattermost/config/config.json | jq '.SqlSettings.DataSource' | grep -o '/.*?' | tail -c +2 | head -c -2)
+	if [ -z "$database" ] 
+	then
+		printf "\nError: Could not extract the database name from ./mattermost/config/config.json. Maybe the script runs in the wrong directory? Exit 1.\n"
+		exit 2
```
EV21, line 196: It says `Exit 1` but it does `exit 2`.
You can just say something like `Could not ... - Abort update procedure` and if you want `with exit code 2`

And just to make it more simple you can extract all the formatting stuff to a helper function.
This example also redirects the error messages so users are able to filter the important error messages and ignore the info messages.
```bash
function error_message { echo "Error: $*" >>/dev/stderr }

# And then just use it like this
error_message "Houston, we have a problem - Abort update procedure with exit code 42"
```

source/guide_mattermost.rst
```diff
@@ -160,12 +171,94 @@ When upgrading to Mattermost 6.4 or newer you need to change the collation of th
...
+	else 
+		printf "Using database: $database\n"
+	fi
+
+	version=$1
+	if [ -z "$version" ] 
+	then
+		printf "No manual version given. Check from official website ...\n"
+
+		version=$(curl -s https://api.github.com/repos/mattermost/mattermost-server/releases/latest | jq '.tag_name' | tail -c +3 | head -c -2)
```
EV21, line 206: `head` and `tail` is magic you don't need this here.

```suggestion
		tag_name=$(curl -s https://api.github.com/repos/mattermost/mattermost-server/releases/latest | jq --raw-output '.tag_name')
		version=${TAG_NAME:1} # This will remove the first character `v` from the version tag `v1.2.3`
```

source/guide_mattermost.rst
```diff
@@ -160,12 +171,94 @@ When upgrading to Mattermost 6.4 or newer you need to change the collation of th
...
+	then
+		printf "\n=== You are up-to-date. No update needed. ===\n"
+		exit 0
+	fi
+
+	printf "\n== Prepare for takeoff ==\n"
+
+	printf "Downloading Version v$version...\n"
+	mattermostfile="mattermost-$version-linux-amd64.tar.gz"
+	wget https://releases.mattermost.com/$1/$mattermostfile
```
EV21, line 222: I think you missed that one
```suggestion
	wget https://releases.mattermost.com/${version}/$mattermostfile
```

EV21 (commented):
As I don't have any experience with mattermost I can just give you a quick review for the cosmetics 💄and basic functionalities - I did not execute your script.
The final review will be done by a Uberspace team member. 

As your script has some checks and error handling it starts to get more complex and less readable.
A best practice for less headaches is to only work with functions (with descriptive names) and don't have code at the top level (divide and conquer).

This is just a suggestion. You don't need to do this, especially when you are not on that level of experience for bash.

example:
```bash
#!/bin/bash
function update_procedure
{
  # backup, download, etc.
}
function database_checks { ... }
function update_available { ... }

function main
{
  database_checks # This may abort the script
  if update_available "${@}"
  then update_procedure
  else echo "We are already in space 🚀"
  fi
}

# This is at the end of the script
main "${@}"
exit $?
```

source/guide_mattermost.rst
```diff
@@ -160,12 +171,95 @@ When upgrading to Mattermost 6.4 or newer you need to change the collation of th
...
+		printf "Using database: $database\n"
+	fi
+
+	version=$1
+	if [ -z "$version" ] 
+	then
+		printf "No manual version given. Check from official website ...\n"
+
+		tag_name=$(curl -s https://api.github.com/repos/mattermost/mattermost-server/releases/latest | jq --raw-output '.tag_name')
+		version=${TAG_NAME:1} # This will remove the first character `v` from the version tag `v1.2.3`
```
chief-nerd, line 207:
```suggestion
		version=${tag_name:1} # This will remove the first character `v` from the version tag `v1.2.3`
```

luto:
Thank you! and thanks @EV21 for the review! 🎉

errotu:
The detection of the correct version could be improved. Currently, the script only checks if the version used at the moment equals the latest one from Github. However, it might be the case that bugfixes for older versions are published on Github which do not affect more current versions. 

As an example: I am currently on v7.10 (published two weeks ago), but the script wants to downgrade me to v7.9.3 (published three days ago).

EV21:
I think it is a really bad practice to mark maintenance releases of lower major versions as `latest` on GitHub. - v7.10.0 should be marked as `latest` on GitHub, same as they do at https://mattermost.com/download/
Maybe create an issue so they may fix their release tagging mechanism or workflow.
If they keep their bad practice the update script will need more intelligence to pick the right version via the API.

It looks like the script also takes a version argument, so something like `mattermost-update 7.10.0` may work for you until it is fixed - either the tagging or the script does some magic. 

Like you suggested the version check could be improved with a little function like I am using in my scripts that prevents downgrades.

something like this:
```bash
# version_lower_than A B
# returns whether A < B
function is_version_lower_than
{
  test "$(echo "$@" |                 # get all version arguments
    tr " " "\n" |                     # replace `space` with `new line`
    sed '/alpha/d; /beta/d; /rc/d' |  # remove pre-release versions (version-sort interprets suffixes as patch versions)
    sort --version-sort --reverse |   # latest version will be sorted to line 1
    head --lines=1)" != "$1"          # filter line 1 and compare it to A
}

function is_update_available
{
  set_local_version
  set_latest_version
  if is_version_lower_than "$LOCAL_VERSION" "$LATEST_VERSION"
  then return 0
  else return 1
  fi
}

function main
{
  if is_update_available
  then
    echo "There is a new version available."
    echo "Doing update from $LOCAL_VERSION to $LATEST_VERSION"
    do_update_procedure
  else
    echo "Your $APP_NAME is already up to date."
    echo "You are running $APP_NAME $LOCAL_VERSION"
  fi
}

main "$@"
exit $?
```

errotu:
Great, thanks a lot!

errotu:
For some reason the script stopped working for me. It seems to recognize "null" as most current version and tries to download and install: https://releases.mattermost.com/ull/mattermost-ull-linux-amd64.tar.gz accordingly (which fails for obvious reasons). 

However, the current tagging and the API endpoint looks good to me: https://api.github.com/repositories/37448451/releases/latest

EV21:
GitHub redirects to another location. I have not seen this behavior before on that endpoint and for other repositories that call works just like before. So in this case we need to change the API call so it follows that redirect

```diff
- 	tag_name=$(curl -s https://api.github.com/repos/mattermost/mattermost-server/releases/latest | jq --raw-output '.tag_name')
+ 	tag_name=$(curl --silent --location https://api.github.com/repos/mattermost/mattermost-server/releases/latest |
+        jq --raw-output '.tag_name')
```

Issue #1437, opened 2022-11-21
Ansible based Uberspace provisioning


tpraxl:
Some time ago, I created an [ISC](https://choosealicense.com/licenses/isc/) licensed collection of [Ansible](https://www.ansible.com/) roles for repeated Uberspace provisionings.

I use them to setup mail domains, accounts, auto responders, web domains, the php version and more with Ansible.
Some of the roles are biased towards special needs, but all are configurable:

* Collection of modular Ansible roles: https://gitlab.com/macrominds/provision/uberspace
* Group Readme with linked real life examples: https://gitlab.com/macrominds/provision/uberspace/readme

I was wondering if there is any interest in this kind of tooling.
If so, I would be glad to share the projects (although I am not sure if the collection even matches the goals of https://lab.uberspace.de).

With this issue, I would like to gather insight on the potential of interest.

Long story short: I would appreciate your feedback.

tpraxl:
Closing this issue due to a lack of interest.

PR #1493, opened 2023-02-18
[ansible] Add Ansible article with references and tags

friesenkiwi:
Hi. I have developed an Ansible collection that contains roles both for administration of an uberspace itself (including tools, domains, webbackends, price etc.) and certain software to install into it:
* https://github.com/friesenkiwi/ansible-collection-uberspace
* https://galaxy.ansible.com/friesenkiwi/uberspace

The roles for the software are based on lab articles.

My main driver is that I'd like to enable other people to just execute one command and have a up and running production-ready software from scratch, including uberspace registration and domain DNS config (via INWX). THis already works for Pleroma and Prosody

This PR includes an article about Ansible itself, as well as a tag and a note for each article for which a known Ansible role exists:

* PostgreSQL
* Coturn
* Prosody
* Pleroma
* Babybuddy

I'd be very happy to adjust this PR and it's contents as well as the collection/roles according to comments. Also of course happy if you'd like to adapt the collection officially or if anybody would like to contribute to it e.g. by modeling some other lab article as a new role.


luto:
Aw, this is so nice, but we can't include it :( There are a couple of projects adding ansible modules to uberspace, for example [the one macrominds made](https://gitlab.com/macrominds/provision/uberspace). However, we're unable to provide support for 3rd-party ansible modules and we provide support for everything in the lab. So we can't include the modules in the lab. Sorry :(

friesenkiwi:
@luto All right, I understand the problem that support from you might be expected by users if the roles/modules are presented too prominently in the lab and you cannot provide it.

On the other would you think these community contributions could also be valuable to uberspace users in a way? Because if so, maybe we could find some kind of middle ground, like e.g. just one page "Unsupported 3rd party community contributions" with explicit support waiver where only links to external resources are listed or so?

For example I myself wasn't aware of @macrominds' contribution (so thanks for linking!) although I had looked around before.

Cheers,

tpraxl:
@friesenkiwi here is the [issue for the macrominds roles](https://github.com/Uberspace/lab/issues/1437), providing the [group readme link](https://gitlab.com/macrominds/provision/uberspace/readme), FYI. I closed the issue due to a lack of interest.

PR #1495, opened 2023-02-21
[freshrss] rewrite guide to use CLI instead of web installer

SalocinHB:


EV21:
I also have a few ideas, maybe you can adopt some of them. 🙂

What do you think about variables? So users can do more `copy & paste` and don't need to build their install command.

```console
[isabell@stardust FreshRSS-latest]$ FRESHRSS_ADMIN_USERNAME='yourFreshRssUsername'
[isabell@stardust FreshRSS-latest]$ FRESHRSS_ADMIN_PASSWORD='yourFreshRssPassword'
[isabell@stardust FreshRSS-latest]$ MYSQL_PASSWORD='yourMySqlPassword'
[isabell@stardust FreshRSS-latest]$ ./cli/do-install.php --default_user ${FRESHRSS_ADMIN_USERNAME} --base_url https://isabell.uber.space --title FreshRSS --api_enabled --db-type mysql --db-host localhost:3306 --db-user ${USER} --db-password ${MYSQL_PASSWORD} --db-base ${USER}_freshrss
...
[isabell@stardust FreshRSS-latest]$ ./cli/create-user.php --user ${FRESHRSS_ADMIN_USERNAME} --password $FRESHRSS_ADMIN_PASSWORD
```
---

What do you think about using
`--base_url https://$USER.uber.space` 
instead of
`--base_url https://isabell.uber.space`
and just mention a note about it?
or
use a variable `DOMAIN=https://isabell.uber.space` like the other suggestion above

---
I think some default values can be skipped - see https://github.com/FreshRSS/FreshRSS/blob/edge/config.default.php
It looks like it should work without `--title FreshRSS` and `--db-host localhost:3306`
```php
'title' => 'FreshRSS',
'db' => [
		# Database server
		'host' => 'localhost',
],
```

---

Do not use double quotes for passwords like `"<yourmysqlpassword>"` - use single quotes `'<yourmysqlpassword>'` so it is handled as a constant string and signs like `$` or `\` will be taken as they are.

SalocinHB:
Yeah, I thought about using variables when I put together that ugly one-liner. I might switch to that later. If it's merged as-is, feel free to change it in another PR.

For options such as `--title`, I skipped the ones that are explicitly listed as optional in the [FreshRSS CLI installation guide](https://github.com/FreshRSS/FreshRSS/blob/edge/cli/README.md#system) but otherwise stuck to that guide.

luto:
Thanks! 🎉

@EV21 I fixed the password quotes. Would you like to submit your proposed changes as an additional PR? :)

PR #1502, opened 2023-03-17
[MongoDB] Extended the guide

kastnerp:
Added instructions to access MongoDB externally without an SSH tunnel

### DIFF
--- a/source/guide_mongodb.rst
+++ b/source/guide_mongodb.rst
@@ -164,12 +164,15 @@ Access without SSH tunnel
 You can access your MongoDB instance externally by opening a port and modifying the service config file.
 
 .. code-block:: none
- uberspace port add
 
-Modify your ``~/etc/services.d/mongodb.ini`` file like so. Replace the port number with the port that was opened by the previous command. 
+ [isabell@stardust ~]$ uberspace port add
+ Port 43120 will be open for TCP and UDP traffic in a few minutes.
+ [isabell@stardust ~]$ 
+
+Modify your ``~/etc/services.d/mongodb.ini`` file. Replace the port number with the port that was opened by the previous command. 
 
 .. code-block:: ini
- :emphasize-lines: 4,5
+ :emphasize-lines: 4,5,8,9
 
  [program:mongodb]
  command=mongod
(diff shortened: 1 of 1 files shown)


nichtmax (changes requested):
Thanks for your addition! From a security perspective you should only open ports with encryption. [MongoDB supports TLS](https://www.mongodb.com/docs/manual/tutorial/configure-ssl/), [we provide the certificates](https://manual.uberspace.de/web-https/#certificate-access). As we have a teaching assignment here could you please update your PR with TLS configuration?

source/guide_mongodb.rst
```diff
@@ -157,6 +157,58 @@ Now you can just run ``mongo`` to connect to your MongoDB instance:
...
+   --auth
+   --unixSocketPrefix %(ENV_HOME)s/mongodb
+   --tlsMode requireTLS 
+   --sslPEMKeyFile %(ENV_HOME)s/mongodb/%(ENV_USER)s.uber.space.pem
+ autostart=yes
+ autorestart=yes
+ startsecs=30
+
+
+Create the `.pem` file with the certificate chain:
```
SalocinHB, line 191: The following steps need to be repeated whenever the Let's Encrypt certificate changes (typically every 60 days).

source/guide_mongodb.rst
```diff
@@ -157,6 +157,58 @@ Now you can just run ``mongo`` to connect to your MongoDB instance:
...
+ [isabell@stardust ~]$ 
+
+Update and restart the service.
+
+.. code-block:: none
+
+ [isabell@stardust ~]$ supervisorctl stop mongodb
+ supervisorctl reread 
+ supervisorctl update mongodb 
+ supervisorctl start mongodb
```
SalocinHB, line 205: This should be a single `supervisorctl restart mongodb`. If the service's `.ini` hasn't changed, `reread` and `update` won't do anything and `start` would only tell you that it's already running.

source/guide_mongodb.rst
SalocinHB, line 191: The following steps need to be repeated whenever the Let's Encrypt certificate changes (typically every 60 days).

source/guide_mongodb.rst
```diff
@@ -157,6 +157,58 @@ Now you can just run ``mongo`` to connect to your MongoDB instance:
...
+
+Update and restart the service.
+
+.. code-block:: none
+
+ [isabell@stardust ~]$ supervisorctl stop mongodb
+ supervisorctl reread 
+ supervisorctl update mongodb 
+ supervisorctl start mongodb 
+ [isabell@stardust ~]$
```
SalocinHB, line 206: If only the certificate changed but the `.ini` remains the same, you also need a `restart`  here.

SalocinHB (changes requested):
Please include a way to detect renewed certificates, that should trigger a replace of the `.pem` file and restart the service.

source/guide_mongodb.rst
kastnerp: Sure, how would you like me to do that? Any suggestions?

SalocinHB: Periodically re-create the .pem file and restart the service. A `@monthly` cronjob springs to mind.

ankit-gautam23 (commented):
Creating a cron job could be better option though.

noave:
hey @kastnerp , we will take over the next days and complete the last issue (looks like its only the cronjob) here to complete the merge. If you wanna finish it yourself just give us a short notice :-)

kastnerp:
Busy with work currently, please go ahead with adding the cronjob. Thank you! 🙂

nichtmax:
@kastnerp please get in touch via hallo@uberspace.de to get your mug ☺️

PR #1513, opened 2023-04-05
Create howto_upgrade_node_apps.rst

kimdiallo:
First suggestion for a general upgrade guide for nodejs apps.


noave:
thanks, could you add an detail about "check logs at /logs/supervisord.log if you run into problems"

noave (commented):
[howto_nodejs] fix inline code

PR #1516, opened 2023-04-09
[privatebin] enhance guide

franok:
Enhance guide by instructions for read-only access for non-authenticated users using Basic Auth restriction for POST requests in .htaccess file.

I successfully implemented [a "read-only" restriction](https://github.com/PrivateBin/PrivateBin/wiki/FAQ#how-can-i-link-to-a-read-only-mode-where-users-cant-submit-pastes) on my instance of PrivateBin and would like to share my knowledge in the lab guide.

### DIFF
--- a/source/guide_privatebin.rst
+++ b/source/guide_privatebin.rst
@@ -161,7 +161,7 @@ Of course also adjust the file if you already use a robots.txt.
 Making your PrivateBin instance read-only
 -----------------------------------------
 
-If you want to limit write access to your PrivateBin instance, i.e. specify who can create content, you can configure the ``.htaccess`` file accordingly.
+If you want to limit write access to your PrivateBin instance, i.e. specify who can paste data, you can configure the ``.htaccess`` file accordingly.
 Choose a username that should have write access and provide it to the ``htpasswd`` command:
 
 .. code-block:: console
--- a/source/guide_privatebin.rst
+++ b/source/guide_privatebin.rst
@@ -174,6 +174,7 @@ Choose a username that should have write access and provide it to the ``htpasswd
  [isabell@stardust html]$
  
 Further users can be added by omitting the ``-c`` flag:
+
 .. code-block:: console
  
  [isabell@stardust html]$ htpasswd .htpasswd another-user
--- a/source/guide_privatebin.rst
+++ b/source/guide_privatebin.rst
@@ -158,7 +158,7 @@ If you followed this guide, it is already at the right place in your :manual:`Do
 However, if you installed PrivateBin into a subdirectory, you have to move ``robots.txt`` back into the DocumentRoot.
 Of course also adjust the file if you already use a robots.txt.
 
-Making your PrivateBin instance read-only
+Making your PrivateBin Instance read-only
 -----------------------------------------
 
 If you want to limit write access to your PrivateBin instance, i.e. specify who can paste data, you can configure the ``.htaccess`` file accordingly.
@@ -228,7 +228,7 @@ The .htaccess file should look like this:
  php_value max_file_uploads 100
  </IfModule>
 
-The PrivateBin site is still visible to the public
+The PrivateBin site is still visible to the public. 
 When a user tries to publish content, a Basic-Auth popup will ask for username and password.
 The generated links are accessible to everyone.
 
@@ -238,7 +238,17 @@ Updates
 
 .. note:: Check the update feed_ regularly to stay informed about the latest version.
 
-Updating is quite easy. Just repeat all steps of the :lab_anchor:`Installation chapter <guide_privatebin.html#installation>`.
+Backup your config.
+
+.. code-block:: console
+
+ [isabell@stardust ~]$ cd ~/html
+ [isabell@stardust html]$ cp -p .htaccess .htaccess.backup
+ [isabell@stardust html]$ cp -p .htpasswd .htpasswd.backup
+ [isabell@stardust html]$ cp -rp ~/privatebin-data/ ~/privatebin-data-backup
+ [isabell@stardust html]$ 
+
+Then repeat the steps of the :lab_anchor:`Installation chapter <guide_privatebin.html#installation>`.
 Your configuration file won't get overwritten.
 
 Check the Release-Notes if the configuration changed between ``cfg/conf.sample.php`` and your ``conf.php``.
--- a/source/guide_privatebin.rst
+++ b/source/guide_privatebin.rst
@@ -1,7 +1,7 @@
 .. highlight:: console
 
 .. author:: Nepomacs <https://github.com/Nepomacs/>
-.. author:: <https://franok.de>
+.. author:: franok <https://franok.de>
 
 .. categorize your guide! refer to the current list of tags: https://lab.uberspace.de/tags
 .. tag:: privacy
@@ -113,7 +113,7 @@ If not already there, go to the ``html`` directory before running ``mv``.
 Changing index.php
 ------------------
 
-Now edit ``~/html/index.php``  to inform PrivateBin about to the new location of the folders.
+Now edit ``~/html/index.php``  to inform PrivateBin about the new location of the folders.
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


source/guide_privatebin.rst
```diff
@@ -97,27 +99,27 @@ PrivateBin provides a .htaccess file, which blocks some known robots and link-sc
...
-It is recommended_ to move the configuration, data files, templates and PHP libraries outside of your document root.
-To do that, create a folder  ``privatebin-data`` in ``/var/www/virtual/isabell/`` and move the folders to the new location (remember to replace ``isabell`` with your own username!).
+It is recommended_ to move the configuration, data files, templates and PHP libraries outside of your document root. This is useful to secure your installation.
+To do that, create a folder  ``privatebin-data`` in ``/home/isabell/`` and move the folders to the new location (remember to replace ``isabell`` with your own username!).
 If not already there, go to the ``html`` directory before running ``mv``.
 
 .. code-block:: console
 
  [isabell@stardust ~]$ cd ~/html
- [isabell@stardust html]$ mkdir /var/www/virtual/isabell/privatebin-data
```
luto, line 107: is there a reason for moving this directory?

source/guide_privatebin.rst
```diff
@@ -59,29 +60,30 @@ Installation
...
 
-Untar the archive and then delete it. Replace the version in the file name with the one you downloaded.
+Untar the archive and then delete it.
 
 .. code-block:: console
  :emphasize-lines: 1,2
 
- [isabell@stardust html]$ tar -xzf 1.5.0.tar.gz --strip-components=1
- [isabell@stardust html]$ rm 1.5.0.tar.gz
+ [isabell@stardust html]$ tar -xzf PrivateBin-1.5.1.tar.gz --strip-components=1
```
luto, line 85: Please use `$PBIN_VERSION` here and in the following lines.

source/guide_privatebin.rst
```diff
@@ -156,13 +158,92 @@ If you followed this guide, it is already at the right place in your :manual:`Do
...
+    Require valid-user
+ </LimitExcept>
+ 
+ <IfModule mod_php7.c>
+ php_value max_execution_time 30
+ php_value post_max_size 10M
+ php_value upload_max_size 10M
+ php_value upload_max_filesize 10M
+ php_value max_file_uploads 100
+ </IfModule>
```
luto, line 224: We don't use mod_php, you, can skip this block.

source/guide_privatebin.rst
franok: `/var/www/virtual/$USER/<folder>` is the DocumentRoot, i.e. the place where directories are located that are directly made accessible via the apache webserver (e.g. `html` or any other folder matching a subdomain, e.g. `privatebin.example.com`).

PrivateBin suggests to locate the data in a different directory than the public content, for access security reasons.
Uberspace manual also mentions  [1]:

> If one of the DocumentRoots gets compromised (e.g. because of a CVE), all other files within all other DocumentRoots can be compromised as well.

In all other guides, data is usually separated from DocumentRoot and located under `/home/$USER`. Thus, I wanted to follow this practice.

[1] https://manual.uberspace.de/web-documentroot/

franok:
@luto  thanks for the review, I replied to one of your comments and addressed all other requested changes

source/guide_privatebin.rst
SalocinHB: > In all other guides, data is usually separated from DocumentRoot and located under /home/$USER. Thus, I wanted to follow this practice.

Good idea, but shouldn't it be something like `~/privatebin-data` then?

source/guide_privatebin.rst
```diff
@@ -97,27 +99,27 @@ PrivateBin provides a .htaccess file, which blocks some known robots and link-sc
...
+It is recommended_ to move the configuration, data files, templates and PHP libraries outside of your document root. This is useful to secure your installation.
+To do that, create a folder  ``privatebin-data`` in ``/home/isabell/`` and move the folders to the new location (remember to replace ``isabell`` with your own username!).
 If not already there, go to the ``html`` directory before running ``mv``.
 
 .. code-block:: console
 
  [isabell@stardust ~]$ cd ~/html
- [isabell@stardust html]$ mkdir /var/www/virtual/isabell/privatebin-data
- [isabell@stardust html]$ mv -t /var/www/virtual/isabell/privatebin-data cfg/ lib/ tpl/ vendor/
+ [isabell@stardust html]$ mkdir /home/$USER/privatebin-data
```
SalocinHB, line 109: you can use `$HOME` or `~` instead of `/home/$USER`

franok: done

source/guide_privatebin.rst
noave: the choice of words in the manual is not very precise, I opened an Issue there to improve it. Actually, it is enough to store the data above the document roots in /var/www/username/ because only domain named folders and the html folders are interpreted as document roots. This is the way in all other guides. If there is a compromised PHP app, this will have full access to your uberspace anyway, that is why we propose to use a dedicated uberspace account for each app

noave:
hey, like suggested in the comment I would propose to do it like this with moving the app folders:

```
[noaguide@perseus noaguide]$ pwd
/var/www/virtual/noaguide
[noaguide@perseus noaguide]$ mkdir privatebin
[noaguide@perseus noaguide]$ mv -t privatebin html/cfg/ html/lib/ html/tpl/ html/vendor/
```
so you have a folder structure like here
```
[noaguide@perseus noaguide]$ tree -L 2
.
├── html
│   ├── bin
│   ├── browserconfig.xml
│   ├── CHANGELOG.md
│   ├── CREDITS.md
│   ├── css
│   ├── i18n
│   ├── img
│   ├── index.php
│   ├── INSTALL.md
│   ├── js
│   ├── LICENSE.md
│   ├── manifest.json
│   ├── nocontent.html
│   ├── Procfile
│   ├── README.md
│   ├── robots.txt
│   └── SECURITY.md
└── privatebin
    ├── cfg
    ├── lib
    ├── tpl
    └── vendor
```

this would rather be consistent with the other guides. But i must admit, I am a bit surprised, that privatebin does not have a _real_ clean public folder like it is usual in other apps.

But anyway, I was able to install it this way

source/guide_privatebin.rst
SalocinHB: No, Apache can still access those directories, even without PHP. Granted, it would need something like an ill-configured `.htaccess`, but the possibility is there – which is why Nextcloud keeps it's data directory in $HOME.

franok:
> hey, like suggested in the comment I would propose to do it like this with moving the app folders:
> [...]
> this would rather be consistent with the other guides. But i must admit, I am a bit surprised, that privatebin does not have a _real_ clean public folder like it is usual in other apps.

The latest comment by @SalocinHB suggests something different:
https://github.com/Uberspace/lab/pull/1516#discussion_r1172365440
> which is why Nextcloud keeps it's data directory in $HOME.


> 
> But anyway, I was able to install it this way

I updated this guide according to my setup on my own uberspace.
If you were able to install it with my instructions, I do not really see a reason, why it coudln't be merged this way.

franok:
As I do not see any functional Todo for this PR, I set it back from "Draft" state, to "Ready for Review".
Please reconsider, if this update can be merged in current state.

noave:
> The latest comment by @SalocinHB suggests something different:
nextcloud only stores the static data files in the home folder, which is a good and clean approach. But the all the code files reside in /var/www. Most frameworks nowadays usually have a specific public folder (like symfony etc.).

However, here you named the folder privatebin-data which is not really correct why I chose just privatebin where I moved the codefiles and changed the path like you suggested. Then with the first note one creates, the app creates a folder `data` within there. I thought about using a symlink from `/var/www/virtual/username/privatebin/data` to `~/privatebin-data`. Then code, public folder, data would have been separated. But the apps also creates php code files within the data folder :smile: so its all mixed up again anyways.

So, since there is this little tiny security afford of storing the data files in `~/home` I would for now follow your approach using the ~/privatebin folder (but with this naming, not privatebin-data). Even if this means storing php code files in ~/home which is not the usual way here. 

The cleanest approach would be using a mysql database, since we have everything on board on a Uberspace I guess there is no reason to depend on storing the data in files at all. But for now I will add this only as an [issue](https://github.com/Uberspace/lab/issues/1519).

noave:
looks like you did not allow pushing to your branch, so I will first merge here and rename privatebin-data to privatebin afterwards

noave:
thanks in any case :)

franok:
> So, since there is this little tiny security afford of storing the data files in `~/home` I would for now follow your approach using the ~/privatebin folder (but with this naming, not privatebin-data). Even if this means storing php code files in ~/home which is not the usual way here.

Sure, I'm also fine with `~/privatebin`.


> The cleanest approach would be using a mysql database, since we have everything on board on a Uberspace I guess there is no reason to depend on storing the data in files at all. But for now I will add this only as an [issue](https://github.com/Uberspace/lab/issues/1519).

Yes, I guess it can be tackled in a separate PR to enhance the guide with instructions for using MySQL.


> looks like you did not allow pushing to your branch, so I will first merge here and rename privatebin-data to privatebin afterwards

Ah, sorry, did not remember I restricted edits by maintainers.

Thank you for following up on the naming and thank you for the review & merging of the PR 🙂👍

PR #1521, opened 2023-04-22
forgejo: convert markdown link to RST

chloe-the-catgirl:
The current compiled output is broken on https://lab.uberspace.de/guide_forgejo/, this small fix should solve the issue.


noave:
thanks

PR #1522, opened 2023-04-28
[Gitea] update the update script

EV21:
## config

- hide version number, so this info is only visable at `/admin` or via CLI
  grafik

## update script changes

- I refactored a few things,
- removed some checks (they were actually added for debug reasons) as the issues should be fixed by the `fix_stop_signal` function,
- fixed a bug in the version comparator so it is not a tautology
- ignore versions with pre-release code words,
  this is maybe important for Forgejo, but it looks like they now publish their pre-releases in another [repository](https://codeberg.org/forgejo-experimental/forgejo/releases),
   so maybe a slightly changed version of this script (API requests) can be used for Forgejo,
- handle expired PGP keys

## [deploy preview 👀](https://deploy-preview-1522--uberlab.netlify.app/guide_gitea/)

### DIFF
--- a/source/guide_gitea.rst
+++ b/source/guide_gitea.rst
@@ -295,9 +295,9 @@ You can also automate the update by using a custom script that automatically exe
   #!/usr/bin/env bash
 
   APP_NAME=Gitea
-  GITEA_LOCATION=$HOME/gitea/gitea
+  GITEA_BINARY=$HOME/gitea/gitea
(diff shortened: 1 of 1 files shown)


noave:
thanks, the update script is quite complex and it would be nicer to have it in a separate repository (maybe with a responsible maintainer). But for now we don't have a process for that and this way it keeps in sync with the guide itself.

Issue #1432, opened 2022-11-17
[invoiceplan] improve guide


noave:
## installation

- provide link to latest version and advise to replace it
- only the beta 1.6. supports php > 7.4, check how stable it is and how installation works

## configuration

- put creating DB, copying config, moving htaccess in an extra steps
- use `isabell_invoiceplan` ad db name
- check if the config could be just done in the config file (not in web installer)

PR #1533, opened 2023-05-18
[invoiceplane] improve guide and update for PHP 8.1 (#1432)

toddy15:


nichtmax:
This does not work at all, when I open the setup page I get _a lot_ of errors like

```A PHP Error was encountered

Severity: 8192

Message: Creation of dynamic property CI_URI::$config is deprecated

Filename: core/URI.php

Line Number: 102

Backtrace:

File: /var/www/virtual/moritz/html/index.php
Line: 329
Function: require_once

A PHP Error was encountered

Severity: 8192

Message: Creation of dynamic property MY_Router::$uri is deprecated

Filename: core/Router.php

Line Number: 128

Backtrace:

File: /var/www/virtual/moritz/html/index.php
Line: 329
Function: require_once
```

toddy15:
Hi @nichtmax,

thanks for the review. That's weird -- I think that I've tested everything with PHP 8.2 and it worked. The problem is that PHP 8.2 emits deprecation warnings about dynamic properties. Did you change that recently in the uberspace setup?

Anyway, the application works well with PHP 8.1, so I've added a patch to use the default PHP 8.1 on uberspace. I've just created a new uberspace and did run all instructions by copying them from the manual. It's a working installation. If you want, I'll give you the URL, so you can have a look. If you don't need this, I'll remove that uberspace instance.

Regards,
Tobias

PR #1545, opened 2023-06-05
Update guide_pretix.rst to make it work with latest Python version

Konafets:


source/guide_pretix.rst
```diff
@@ -61,13 +61,20 @@ Install pretix using the python package manager. Be sure to replace the pseudo b
 
 ::
 
- [isabell@stardust ~]$ pip3.9 install pretix==4.16.0 gunicorn --user
+ [isabell@stardust ~]$ pip3.11 install pretix gunicorn --user
```
luto, line 64: Is there a reason for not specifying a version here? Having `==4.20.0` ensures the guide works as-is.

Konafets: My idea was to have the guide change less often but thinking about again it makes more sense to pin the guide to a specific version.

luto:
Thanks! :)

PR #1550, opened 2023-06-08
[Nextcloud] add `occ upgrade` command for app updates

EV21:
App updates may require additional steps to be done by the `occ upgrade` command.

Usually you would notice this by being redirected to an upgrade page in your browser.

So this should fix it:
app-update


luto:
sweet, thanks! :)

PR #1551, opened 2023-06-12
[OSTicket] Changed a broken command

n41lit:
Deleted a point on line 62, the git command throws an error if the external path is followed by a point

### DIFF
--- a/source/guide_osticket.rst
+++ b/source/guide_osticket.rst
@@ -59,7 +59,7 @@ To install osTicket we clone the current version using Git. ``cd`` to your :manu
  :emphasize-lines: 2
 
  [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
- [isabell@stardust html]$ git clone https://github.com/osTicket/osTicket.git .
+ [isabell@stardust html]$ git clone https://github.com/osTicket/osTicket.git
  Cloning into '.'...
  (...)
  [isabell@stardust ~]$


source/guide_osticket.rst
```diff
@@ -59,7 +59,7 @@ To install osTicket we clone the current version using Git. ``cd`` to your :manu
  :emphasize-lines: 2
 
  [isabell@stardust ~]$ cd /var/www/virtual/$USER/html/
- [isabell@stardust html]$ git clone https://github.com/osTicket/osTicket.git .
+ [isabell@stardust html]$ git clone https://github.com/osTicket/osTicket.git
  Cloning into '.'...
  (...)
  [isabell@stardust ~]$
```
EV21, line 65: Can you be more specific about what the error looks like?

The dot just prevents the clone command from creating a subfolder.
Your fix would result into moving everything into `~/html/osTicket` - so the rest of the guide would be broken and also the app would be located in a subpath `isabell.uber.space/osTicket`
```suggestion
 [isabell@stardust html]$ rm ~/html/nocontent.html
 [isabell@stardust html]$ git clone https://github.com/osTicket/osTicket.git ~/html
 Cloning into '/home/isabell/html'...
 (...)
 [isabell@stardust html]$
```

SalocinHB:
The dot `.` is correct. Did you skip the space ` ` between the URL and the dot?

PR #1552, opened 2023-06-12
FastAPI: minor improvements

entorb:
/my-backend as default route
Python 3.11


nichtmax:
Thanks for your addition! Unfortunately we can't merge your PR because of our [style guide](https://github.com/Uberspace/lab/blob/main/STYLE.md): 
- Don't mention additional document roots. Keep it simple. Don't use subfolders. Always use the standard document root ~/html. Always assume the document root is empty.

PR #1556, opened 2023-06-15
Update guide_postgresql.rst

Konafets:
Add export LC_ALL and LC_TYPE to prevent error 

> initdb --pwfile ~/pgpass.temp --auth=scram-sha-256 -E UTF8 -D ~/opt/postgresql/data/
> The files belonging to this database system will be owned by user "nas".
> This user must also own the server process.
>
> initdb: error: invalid locale settings; check LANG and LC_* environment variables

### DIFF
--- a/source/guide_postgresql.rst
+++ b/source/guide_postgresql.rst
@@ -120,6 +120,8 @@ We now use the temporary password file ``~/pgpass.temp`` (containing only your p
 .. code-block:: console
     :emphasize-lines: 1
 
+    [isabell@stardust ~]$ export LC_CTYPE="de_DE.UTF-8"
+    [isabell@stardust ~]$ export LC_ALL="de_DE.UTF-8"
     [isabell@stardust ~]$ initdb --pwfile ~/pgpass.temp --auth=scram-sha-256 -E UTF8 -D ~/opt/postgresql/data/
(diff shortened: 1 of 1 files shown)


source/guide_postgresql.rst
```diff
@@ -120,6 +120,8 @@ We now use the temporary password file ``~/pgpass.temp`` (containing only your p
 .. code-block:: console
     :emphasize-lines: 1
 
+    [isabell@stardust ~]$ export LC_CTYPE="de_DE.UTF-8"
```
SalocinHB, line 123: I'm not sure de_DE is the best choice here. The lab is completely in English, exporting a de local would cause some applications to print German instead of English info and error messages.

SalocinHB:
This only leads to errors if your SSH client is sending an environment (`SendEnv`) which includes a locale not available on the server.

Konafets:
How can I check if my SSH client does this?

SalocinHB:
> How can I check if my SSH client does this?

Check your client's config. On unix-like (Linux, macOS) systems, it's likely `/etc/ssh/ssh_config`

noave:
I marked the PR as draft please change it back to ready if this is completed from your perspective

noave:
I will close this now, if there is any more need for this you may reopen the PR

Issue #1553, opened 2023-06-14
[mailman3] guide is broken


kimdiallo:
@rikroe  After following all the steps, the mailman3 service cannot be started because the system openssl version is too low.

```bash
 File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/xmlconfig.py", line 393, in endElementNS
    self.context.end()
  File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/config.py", line 701, in end
    self.stack.pop().finish()
  File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/config.py", line 868, in finish
    args = toargs(context, *self.argdata)
  File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/config.py", line 1698, in toargs
    args[str(name)] = field.fromUnicode(s)
  File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/fields.py", line 272, in fromUnicode
    v = vt.fromUnicode(s)
  File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/fields.py", line 174, in fromUnicode
    value = self.context.resolve(name)
  File "/home/isabell/.local/lib/python3.8/site-packages/zope/configuration/config.py", line 223, in resolve
    __import__(mname)
  File "/home/isabell/.local/lib/python3.8/site-packages/mailman/model/autorespond.py", line 28, in <module>
    from mailman.utilities.datetime import today
  File "/home/isabell/.local/lib/python3.8/site-packages/mailman/utilities/datetime.py", line 27, in <module>
    from mailman.testing import layers
  File "/home/isabell/.local/lib/python3.8/site-packages/mailman/testing/layers.py", line 43, in <module>
    from mailman.testing.helpers import (
  File "/home/isabell/.local/lib/python3.8/site-packages/mailman/testing/helpers.py", line 47, in <module>
    from requests import request
  File "/home/isabell/.local/lib/python3.8/site-packages/requests/__init__.py", line 43, in <module>
    import urllib3
  File "/home/isabell/.local/lib/python3.8/site-packages/urllib3/__init__.py", line 42, in <module>
    "urllib3 v2.0 only supports OpenSSL 1.1.1+, currently "
[...]

PR #1558, opened 2023-06-17
[Mailman3] Update to latest versions

rikroe:
This PR updates the Mailman 3 guide:
- use `python3.11`
- pin `urllib3<2`
- use `mailman-web` to install all required web components
- due to `py3.11`, the `pysqlite3-binary` workaround is not needed anymore

(Hopefully) fixes https://github.com/Uberspace/lab/issues/1553

### DIFF
--- a/source/guide_mailman-3.rst
+++ b/source/guide_mailman-3.rst
@@ -76,7 +76,7 @@ Install Mailman 3 and its dependencies via pip.
 
 ::
 
- [isabell@stardust ~]$ pip3.8 install --user mailman hyperkitty postorius mailman-hyperkitty whoosh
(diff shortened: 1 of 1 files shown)


nichtmax:
Phew... This is a very complex guide and I can't get Mailman running on a test account. Regardless, we will only continue to support .qmail until the end of Uberspace 7 and I am not aware of any way to run Mailman on our end without .qmail. 

We have decided as a team that we will deprecate and no longer support guides that require .qmail. It is simply unfair to our users to provide guides that will no longer work in the future.

Please contact hallo@uberspace.de, you will still get a thank you in form of a mug (or a t-shirt if you like) from us.

rikroe:
No worries about that! Will you mark the guide as deprecated and/or delete it?

PR #1561, opened 2023-07-02
Update guide_redmine.rst

fahrplaner:
remove highlighting, no c&p possible


SalocinHB:
The problem with copy & paste here isn't the (unnecessary)  highlighting but the `code-block: console`.

PR #1568, opened 2023-07-22
Update guide_vaultwarden.rst

ppac:
data is stored in `~/vaultwarden/data/` instead of `~/vaultwarden/output/data`

### DIFF
--- a/source/guide_vaultwarden.rst
+++ b/source/guide_vaultwarden.rst
@@ -201,6 +201,14 @@ Alternatively, you can do the backup with a timestamp and it can be useful if yo
 
  [isabell@stardust ~]$ sqlite3 ~/vaultwarden/data/db.sqlite3 ".backup '$HOME/vaultwarden/data/db-backup/$(date +%Y-%m-%d).sqlite3'"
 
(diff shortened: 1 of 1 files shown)


noave:
I cannot really follow your approach here, as far as I see, the command `./docker-image-extract vaultwarden/server:alpine` will install the vaultwarden code into the output folder. Is that something you want to change or you just want to put the backup somewhere else?

ppac:
Yes, Vaultwarden itself is installed into the `vaultwarden/output`-folder. The database is stored in the `vaultwarden/data`-folder as you can see here:

`[isabell@stardust ~]$ export DATA_FOLDER=$HOME/vaultwarden/data`

So, from my point of view, there is a need for an adjustment of the paths (which I have done). I've double-checked my own Vaultwarden-instance and there is no data stored in the `vaultwarden/output`-folder.

escuco:
I can confirm that the backup won't work with the actual commands and this PR from @ppac makes sense. I hadn't noticed this before because I didn't used the backup in this way before.
(tested on my own installation)

kimdiallo:
works, thank you.

PR #1571, opened 2023-07-30
[baikal] mentioned baikal_updates script

fl3a:
Added one sentence with link to above script which i wrote for periodic checks on baikal updates.


luto:
thanks :)

PR #1576, opened 2023-08-18
Update guide_mosquitto.rst startsecs for daemon

MarvinGS:
Support added this line to my config, so i guess it might be a good default!


luto:
Thank you!

PR #1590, opened 2023-09-21
improve beehive guide

noave:


noave:
thanks to https://github.com/Uberspace/lab/pull/1586

PR #1613, opened 2023-10-04
Update guide_grafana-loki.rst

tillmannheigel:
The labels stage takes the level and component entries from the previous regex stage and promotes them to a label. For example, level=error may be a label added by this stage.


nichtmax:
Hey @tillmannheigel thanks for your PR! To be honest we're not experts  in Grafana Loki and I'm not sure what your commit fixes. Help me out: Is the guide broken without it? Is this just an improvement? Is this based on taste or best practice and don't you usually configure this to your own needs?

nichtmax:
I got feedback from my colleagues who setup and maintain our Grafana:
> that even contradicts the best practice that you shouldn't put everything into labels.

We won't merge your PR as this seems to be more of a personal preference thing.

Issue #1274, opened 2022-06-14
[fittrackee] Add a guide for FitTrackee


escuco:
[Fittrackee](https://samr1.github.io/FitTrackee/index.html) is a web application to track and store your outdoor and sport activities. It seems to be a good alternative to Runalyze, which has become proprietary.

I tried to install FitTrackee to a Uberspace different times, but my knowledge (PiPy, pip, virtualenv, ...) isn't enough to finish it.

Maybe I will try again at a later time, but If someone want to give it a try.

escuco:
So far I've got the web backend working now and I'm tricking around with an error in the user registration process and mail function.

When I got everything working, I'll write a documentation for FitTrackee.

qimmiq:
@escuco would be happy to test and help :)

escuco:
Even if Fittrackee is working fine on my Uberspace and I'm using the application every week, I'm not able to write a guide for the lab at the moment - sorry.
So if someone wants to take over, go for it. If there are any questions, I am available.

escuco:
If you want to have a look, here's my installation:
https://fittrackee.escuco.de/
(open for registrations)

PR #1617, opened 2023-10-11
Create Guide for Fittrackee

astridx:
I installed [Fittrackee](https://samr1.github.io/FitTrackee/en/index.html) and noted my steps in detail here:
https://blog.astrid-guenther.de/weblog/2023-09-23-fittrackee-uberspace-installation/

For the Lab I have created this shortened version. I hope that as you expect.

--
See https://github.com/SamR1/FitTrackee/issues/187
See https://github.com/Uberspace/lab/issues/1274


source/guide_fittrackee.rst
```diff
@@ -0,0 +1,194 @@
...
+  * :manual:`Python <lang-python>` in the version >3.8.1. 
+  * :manual:`PostgreSQL <guide_postgresql>` in the version 11+. 
+
+Installation
+============
+
+Create folder
+-------------
+
+We do not install Fittrackee in the :manual:`DocumentRoot <web-documentroot>`. We install it in a separate folder. First we create this folder.
```
nichtmax, line 60: No, we don't install it in a separate folder, please check out the [STYLEGUIDE](https://github.com/Uberspace/lab/blob/main/STYLE.md):

source/guide_fittrackee.rst
```diff
@@ -0,0 +1,194 @@
...
+
+::
+
+ [isabell@stardust ~]$ cd /home/isabell
```
nichtmax, line 64: Please use `cd ~` instead. `~` is home and users don't have to change `isabell` to their user name.

source/guide_fittrackee.rst
```diff
@@ -0,0 +1,194 @@
...
+
+
+Install Fittrackee
+------------------
+
+``cd`` into your empty Fittrackee folder and set up virtual environment for Python 3.9:
+
+::
+
+ [isabell@stardust ~]$ cd /home/isabell/fittrackee
```
nichtmax, line 76: same here: Please use `cd ~/fittrackee` instead. `~` is home and users don't have to change `isabell` to their user name.

nichtmax (changes requested):
Thank you very much for your guide! 

I did not try to install the guide, please adapt it to our [STYLEGUIDE](https://github.com/Uberspace/lab/blob/main/STYLE.md) and remove the user name `isabell` wherever possible. 

We will try out the guide when you get back to us with the requested changes ☺️

astridx:
@nichtmax Thanks for the review. Please check if it is ok now. It should work. I went through the steps myself when correcting and this is the result in a test instance: https://ub.uber.space/login

source/guide_fittrackee.rst
```diff
@@ -0,0 +1,186 @@
...
+Special note: For an installation in our system it is important to customize the variables ``HOST`` and ``DATABASE_URL`` as follows:
+
+::
+
+ export HOST=0.0.0.0
+ ...
+ export DATABASE_URL=postgresql://fittrackee:fittrackee@localhost:5432/fittrackee
+ ...
+
+For more information about the available ports, see :manual:`firewall ports <basics-ports>`.
```
nichtmax, line 141: When I get your guide right you don't need to open any ports in the firewall, the database connection is local and your app can be reached with a web backend. Maybe just leave out this part? The manual article you link to is about opening ports in the firewall, which we don't need to to.

source/guide_fittrackee.rst
```diff
@@ -0,0 +1,186 @@
...
+::
+
+ (fittrackee_venv) [isabell@stardust fittrackee]$ ftcli db upgrade
+
+Start the application
+
+::
+
+ (fittrackee_venv) [isabell@stardust fittrackee]$ fittrackee
+
```
nichtmax, line 173: What happens when we reboot the server? Can you please provide a supervisord service that monitores and keeps your service running? Check out [the manual](https://manual.uberspace.de/daemons-supervisord/). We reboot our servers on a regular basis and sometimes they crash and have to be rebooted to get everything back up running. In your version of the guide every user needs to login after every reboot and start `fittrackee` by hand.

nichtmax (changes requested):
Thank you for the changed, that looks better. I got two things we need you to change ☺️

astridx:
@nichtmax Thanks for the feedback. Please verify if it is ok for you now.

nichtmax:

PR #1621, opened 2023-10-14
[flask] Update guide, resolves #602

### LINKED ISSUE #602: Update Flask Guide with best practice for production
What do you think about updating the UWSGI section / configuration of the flask guide with best practices? For example, with the provided config, python threads do not work (missing `enable-threads` option). Also any errors/typos in the config are ignore (missing `strict` option).
Maybe the example configuration can be updated with some of the stuff explained here: https://www.techatbloomberg.com/blog/configuring-uwsgi-production-deployment/

cmacht:
Updated the guide to use Python 3.11 (instead of default 3.6, which has reached its EOL and is no longer supported by Flask).

Added the new flask logo and made some minor changes to improve readability and the uwsgi configuration as suggested ages ago by #602


nichtmax:
Thank you very much!

PR #1627, opened 2023-10-24
[gitea] Update optional ssh config, fix #1616, resolve #1578

### LINKED ISSUE #1578: [gitea] Emphasize breaking parameters for optional ssh setup
I updated gitea to 1.20.3 and added new SSH keys. I used the [optional ssh setup](https://lab.uberspace.de/guide_gitea/#gitea-ssh-setup-optional) for using the same key for gitea and shell access but failed to notice that I need to delete a few parameters. Compare:

`command="if [ -t 0 ]; then bash; elif [[ ${SSH_ORIGINAL_COMMAND} =~ ^(scp|rsync|mysqldump).* ]]; then eval ${SSH_ORIGINAL_COMMAND}; else /home/isabell/gitea/gitea serv key-1 --config='/home/isabell/gitea/custom/conf/app.ini'; fi",no-port-forwarding,no-X11-forwarding,no-agent-forwarding ssh-ed25519 ... user@host`

vs.

`command="if [ -t 0 ]; then bash; elif [[ ${SSH_ORIGINAL_COMMAND} =~ ^(scp|rsync|mysqldump).* ]]; then eval ${SSH_ORIGINAL_COMMAND}; else /home/isabell/gitea/gitea serv key-1 --config='/home/isabell/gitea/custom/conf/app.ini'; fi",no-port-forwarding,no-X11-forwarding,no-agent-forwarding,no-pty,no-user-rc,restrict ssh-ed25519 ... user@host`

The latter will not work, apparently because of `no-pty,no-user-rc,restrict`. The guide already does not include these, but as I was typing instead of copy&paste, I did miss it initially. Possibly, these were also added in the new version?

Please close this issue, if this is a non-problem (as I said, guide is correct). Otherwise, I would find it helpful to mention the need to omit these parameters and could provide a pull request.

### LINKED ISSUE #1616: [gitea] scp not working due to regex used in the command in authorized_keys
A while ago I did set up a `gitea` instance on my uberspace and modified the `authorized_keys` file with following entry, as shown and described in the `gitea` guide:

`command="if [ -t 0 ]; then bash; elif [[ ${SSH_ORIGINAL_COMMAND} =~ ^(scp|rsync|mysqldump).* ]]; then eval ${SSH_ORIGINAL_COMMAND}; else /home/isabell/gitea/gitea serv key-1 --config='/home/isabell/gitea/custom/conf/app.ini'; fi",no-port-forwarding,no-X11-forwarding,no-agent-forwarding ssh-...`

In general this worked fine for me, with the exception that I could not copy files via `scp` to/from my uberspace. I now did analyze that problem by logging the SSH_ORIGINAL_COMMAND to a file.

It turned out, that in case of using `scp` for me the command is: `/usr/libexec/openssh/sftp-server`.
Therefore the regular expression `^(scp|rsync|mysqldump).*` cannot match. My fix for now is to use `^.*(scp|sftp|rsync|mysqldump).*`.

I think this should be fixed in the guide, but I am not sure yet what would be the best regex to use here. While my approach currently works, I added a `.*` at the beginning of the command as well, which makes this match quite easily. On the other hand I also didn't want to add the whole path within the regex.

cmacht:
Recent changes made it necessary to update the guide:

- gitea has recently added additional ssh options (which got me confused in #1578 )
- `scp` has switched from its own protocol to `sftp` (which prompted #1616 )


kimdiallo (commented):
First of all, thank you very much for your efforts! 

In itself a nice PR nothing wrong and everything works as you describe it. But...

There is the possibility to use the SSH server built into gitea on a specially opened port and either point to a separate key file or use the database as an authorized_key store instead. 

This way you can use one key for both without having to go the way of specially written Auth_Commands. If something changes again on the part of Gitea, no further customizations are needed and nobody you give access to Gitea comes into contact with the SSH account at any point. This not only seems simpler to me, but in the end also more stable, more cleanly separated and therefore more secure. 

Or have I left something out, what do you think? 

All that would be needed is an open port:

 $ uberspace port add

and these lines in the [server] part of app.ini

START_SSH_SERVER = true
SSH_DOMAIN = isabell.uber.space
DISABLE_SSH = false
SSH_PORT = <port>
SSH_LISTEN_HOST = 0.0.0.0
SSH_LISTEN_PORT = %(SSH_PORT)s

cmacht:
Thank you Kim for your polite remarks and helpful insight. I just tested it successfully and this is obviously the superior solution. I'll change the guide accordingly.

kimdiallo:
Everything is running as it should - many thanks for the great cooperation @cmacht !

PR #1632, opened 2023-11-14
[passbolt] Update guide due to changes and updates to be able to install successfully 

AminZoubaa:
Due to updates in the latest version off Passbolt there where updates needed to install and run it successfuly

### DIFF
--- a/source/guide_passbolt.rst
+++ b/source/guide_passbolt.rst
@@ -37,12 +37,24 @@ Passbolt is released under the `AGPL-3.0 license`_.
 Prerequisites
 =============
 
-We’re using :manual:`PHP <lang-php>` in the stable version 7.2. Since new Uberspaces are currently setup with PHP 7.1 by default you need to set this version manually:
+Every Uberspace can choose their :manual:`PHP <lang-php>` version manually.
+We need PHP version 8.1 or above to use Passbolt.
 
+Check your current version with 
 ::
 
- [isabell@stardust ~]$ uberspace tools version use php 7.2
- Selected PHP version 7.2
+  [isabell@stardust ~]$ uberspace tools version show php
+  Using 'PHP' version: '7.1'
+
+If you use an older version, change your php version to the latest available version.
+::
+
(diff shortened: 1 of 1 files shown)


source/guide_passbolt.rst
```diff
@@ -108,20 +120,114 @@ Install the dependencies:
 ::
 
  [isabell@stardust ~]$ cd ~/html/
- [isabell@stardust html]$ wget --output-document=composer.phar https://getcomposer.org/composer-1.phar
+ [isabell@stardust html]$ wget --output-document=composer.phar https://getcomposer.org/composer.phar
  [isabell@stardust html]$ php composer.phar install --no-dev
  [isabell@stardust html]$ rm composer.phar
- [isabell@stardust html]$ cp config/passbolt.default.php config/passbolt.php
+
```
nichtmax, line 126: Is there no `config/passbolt.default.php` anymore in the package? Do you need to create the config completely from scratch with this new version?

nichtmax (changes requested):
Thanks for your guide, I got one small question

source/guide_passbolt.rst
AminZoubaa: No, there is still a `config/passbold.default.php` provided.
I dumped it, because u still need to edit the config manually, and it's spammed with comments.
See: https://github.com/passbolt/passbolt_api/blob/master/config/passbolt.default.php

To get started and have an easier config, I added a minimalistic version with config vars in the header. That would simplify the configuration and search/replace process.

If someone needs more or extended configurations, I will assume that he will look into passbolt docs himself.

nichtmax:

source/guide_passbolt.rst
nichtmax: We want our users to understand what they're doing and not just copy files without thinking. For us it makes more sense to use the provided file (which can change with newer versions) and change thee needed settings.

foxytocin:
#1725 updates the guide for the latest passbolt version 4.5.2. // tested with Uberspace 7.15.10 and PHP 8.1

PR #1633, opened 2023-11-17
Update guide_ejabberd.rst: added module location

wertziop:
I just checked my installation and found the missing module location. This is important during updates, because sometimes new files won't override the existing ones.


nichtmax:
Thank you!

PR #1634, opened 2023-11-21
Create guide_zola.rst

Kaisa-Marysia:
Create guide how to build and install zola on U7.

### DIFF
--- a/source/guide_zola.rst
+++ b/source/guide_zola.rst
@@ -1,0 +1,216 @@
+.. highlight:: console
+
+.. author:: Kaisa-Marysia <https://www.kasiandras-dreams.de>
+
+.. tag:: rust
+.. tag:: blog
+.. tag:: cms
+
+####
+Zola
+####
+
+.. tag_list::
+
+Zola is an in rust written static site generator similar to Hugo and use the Tera template engine.
+
+----
+
+License
+=======
+
+Zola is released under the `MIT License <https://github.com/getzola/zola/blob/master/LICENSE>`_.
+
+Sources
+=======
+
+  * Documentation_
+  * Github_
+  * Zola_
+
+Prerequisites
+=============
+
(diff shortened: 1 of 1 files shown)


nichtmax:
closes #1634

source/guide_zola.rst
```diff
@@ -0,0 +1,216 @@
...
+Modify the configurations
+------------------------
+
+We need to change the `Link-time optimization (LTO) <https://nnethercote.github.io/perf-book/build-configuration.html#link-time-optimization>`_ from true to false, else the building process will fail because of a SIGKILL.
+Also, we need to set some cargo configs, which allows allocating more memory for the building process.
+
+::
+
+  [isabell@stardust zola]$ sed -i 's/lto = true/lto = false/g' Cargo.toml
+  [isabell@stardust zola]$ echo -e -n "[target.nightly-x86_64-unknown-linux-gnu]\nrustflags = [\"-C\", \"target-feature=+atomics,+bulk-memory,+mutable-globals\", \"-C\", \"link-arg=--max-memory=8589934592\"]" > ~/.cargo/config
```
nichtmax, line 68: This does not work at this point for me, there is no ~/.cargo/config. Also please check the [Styleguide](https://github.com/Uberspace/lab/blob/main/STYLE.md) and describe what to edit where in the config files. Your approach might be convenient but we want our users to understand what's going on and this is not self explaining :)

nichtmax (changes requested):
Thanks for your guide, looks pretty solid to me! One small thing though ☺️

source/guide_zola.rst
Kaisa-Marysia: I modified the guide and removed the inplace changes with like in the style guide

kimdiallo:
first of all, thank you very much for your efforts

Unfortunately, I can't get Zola to run. Despite the described measures to keep everything lean, the "cargo install" process is always killed by the OOM killer. 

Presumably the size and thus also the RAM consumption has increased with the last versions and what was still just running under 0.17.x is now simply too big.

Kaisa-Marysia:
You are right.
Also, my latest tests runs into OOMs/SIGKILLs. Looks like I had just luck at my first successful builds.
I agree with this closing of the PR. Thx for validating.

PR #1635, opened 2023-11-23
check for new nodejs version compatibility #1608

GodMod:
Guide works with Node.js 18.


nichtmax:
Thank you for your PRs! Please get in touch via hallo@uberspace.de for your goodie ☺️

PR #1647, opened 2023-11-29
[docusaurus] Update guide for Node 18 and version 3.0.0

### LINKED ISSUE #1596: [docusaurus] check for new nodejs version compatibility
Nodejs 12 / 14 / 16 are EOL and wont be available on Uberspace any longer.

Check if this guide works with Nodejs18 or Nodejs20 and update it accordingly. Otherwise, if you cannot fix it with the new versions mark the guide as broken like it is described here: https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md 

taddydevil:
So basically it works with Node 18 BUT this is a really outdated version.

In this guide Docusaurus 1.14.4 is described and there is already version 3.0.0 released.

I updated the guide for version 3.0.0 which works fine on my instance with the website template, I hope this is ok ;)

closes #1596

### DIFF
--- a/source/guide_docusaurus.rst
+++ b/source/guide_docusaurus.rst
@@ -56,6 +56,7 @@ We create the application directory and install the latest version. The website 
  [isabell@stardust ~]$ cd docusaurus
  [isabell@stardust docusaurus]$ npm init docusaurus@latest my-website classic
  [...]
+ [isabell@stardust docusaurus]$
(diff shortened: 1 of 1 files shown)


nichtmax:
Hey @taddydevil! Thank you very much! The update to version 3.0.0 is *very* ok!

Issue #1629, opened 2023-10-27
[mailman3] Update Python version


kimdiallo:
With the latest Mailman 3 update to version 3.3.9, support for Python 3.8 was dropped. 

https://docs.mailman3.org/projects/mailman/en/latest/src/mailman/docs/NEWS.html

PR #1652, opened 2023-12-03
Update guide_mailman-3.rst to Python 3.9

elRadish:
Mailman 3.3.9 is not compatible with Python3.8 anymore, I updated the docs while installing on 3.9


nichtmax:
closes #1629

nichtmax:
Thank you!

PR #1676, opened 2023-12-09
[OwnTracks] Created guide

schneidr:
I've written up my OwnTracks installation as a lab guide. It's a little extensive, and is not really for beginners as it requires people to compile their own binaries, but it's not that difficult.


nichtmax:

SalocinHB:
I don't think we can accept the guide in this form, but if you let us know which header files are missing we'll check if we can provide them.

schneidr:
That would be awesome, it would make the process so much easier. `libuuid-devel` and `lua-devel` are already installed, that leaves:

- mosquitto-devel
- libcurl-devel
- libsodium-devel
- libconfig-devel
- lmdb-devel

SalocinHB:
> That would be awesome, it would make the process so much easier. `libuuid-devel` and `lua-devel` are already installed, that leaves:
> 
> * mosquitto-devel
> * libcurl-devel
> * libsodium-devel
> * libconfig-devel
> * lmdb-devel

yum says libcurl-devel is obsoleted by the curl RPM which is already installed. We'll provide the rest with the next U7 release.

SalocinHB:
@schneidr we have rolled out the binaries, would you like to try and rewrite your guide accordingly?

kimdiallo:
Flawless, thank you very much for your effort, @schneidr

nichtmax:
@schneidr Did you get a goodie? If not please get in touch via hallo@uberspace.de, I'm sorry, I somehow lost track ☺️

Issue #1680, opened 2023-12-10
[code-server] Requires glibc 2.28 


taddydevil:
Code-Server requires with version 4.17 glibc 2.28 which is not shipped in Uberspace / CentOS 7.
The workaround compile node with glibc 2.18 what the uberspace team already did doesn't work too. 

@tobimori

MaluNoPeleke:
Any updates about this?

Gr1nse:
Any update about this?

SalocinHB:
It's not possible with U7.

runxel:
> It's not possible with U7.

That's a real shame.
Just found out this morning. 

For everybody else who figured this out: [Version 1.98](https://code.visualstudio.com/updates/v1_98) (2025 February release) is the last version that works. At the top the is the version linked for download so you can roll back your VSCode – and hope that U8 won't take another year ;)
Regarding that: Arch is **not** listed by VSCode as supported. Will it be possible for us to use VSCode's remote feature on U8? @SalocinHB

PR #1681, opened 2023-12-10
[code-server] Mark guide as broken

### LINKED ISSUE #1594: [code-server] check for new nodejs version compatibility
Nodejs 12 / 14 / 16 are EOL and wont be available on Uberspace any longer.

Check if this guide works with Nodejs18 or Nodejs20 and update it accordingly. Otherwise, if you cannot fix it with the new versions mark the guide as broken like it is described here: https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md 

taddydevil:
See #1680 for any further information. 

Closes #1594


nichtmax:

Issue #1612, opened 2023-10-04
[meta] add missing programming language tags to guides


noave:
Some guides do not have any programming language tags like:

```.. tag:: lang-nodejs```

please add the tag(s) accordingly to the used programming languages in the tool. The folloing guides are missing the tags:

- [ ] source/guide_couchdb.rst
- [ ] source/guide_ejabberd.rst
- [ ] source/guide_eturnal.rst
- [ ] source/guide_focalboard.rst
- [ ] source/guide_gancio.rst
- [ ] source/guide_gotosocial.rst
- [ ] source/guide_influxdb2.rst
- [ ] source/guide_jellyfin.rst
- [ ] source/guide_keycloak.rst
- [ ] source/guide_libreoffice.rst
- [ ] source/guide_navidrome.rst
- [ ] source/guide_offen.rst
- [ ] source/guide_prosody.rst
- [ ] source/guide_resilio.rst
- [ ] source/guide_salt.rst
- [ ] source/guide_shellinabox.rst
- [ ] source/guide_tarsnap.rst
- [ ] source/guide_teamcity.rst
- [ ] source/guide_teamspeak.rst
- [ ] source/guide_vaultwarden.rst
- [ ] source/guide_wkd.rst

PR #1686, opened 2023-12-12
[meta] added missing language tags

schneidr:
I updated allmost all guides in the list where the tags where missing (some were already added).

I was unable to find the language of Resilio Sync, Teamspeak, as they are proprietary and only provide binaries.

The Web Key Directory (WKD) doesn't really have a language, as it's only a defined directory structure for PGP files.

I was unable to commit the guides for couchdb and gancio because they are missung the `Tested on Uberspace` part and I'm currently unable to do the tests myself.

The tags would be:

CouchDB

```
.. tag:: Erlang
```

gancio:

```
.. tag:: lang-php
.. tag:: lang-nodejs
```

#1612

### DIFF
--- a/source/guide_couchdb.rst
+++ b/source/guide_couchdb.rst
@@ -4,6 +4,7 @@
 
 .. tag:: self-hosting
 .. tag:: database
+.. tag:: Erlang
 
 .. sidebar:: About
 
@@ -144,5 +145,7 @@ Web backend
 .. include:: includes/web-backend.rst
 
 ----
+
+Tested with CouchDB 3.3.3 and Uberspace 7.15.4.0
 
 .. author_list::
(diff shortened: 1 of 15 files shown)


nichtmax:
Hey @schneidr, thank you, that's awesome!

All guides here are tested be us, couchdb and gancio are tested as well, there's just the section missing then.

schneidr:
> All guides here are tested be us, couchdb and gancio are tested as well, there's just the section missing then.

I can of course just add the sections, but I did not want to make up some version numbers :/

nichtmax:
> > All guides here are tested be us, couchdb and gancio are tested as well, there's just the section missing then.
> 
> I can of course just add the sections, but I did not want to make up some version numbers :/

Ah now I get the problem. 

Ganico v1.6.14 / U7.15.4.0

the couchdb binaries are from us and we got testing for it in our release process. Every binary that makes in into production will work. The guide for the configuration is tested with U7.10.0

schneidr:
With all guides updated this closes #1612.

nichtmax:
Thank you! Check your mails ☺️

PR #1688, opened 2023-12-12
[baikal] Add config/baikal.yaml to backup

digicommons:
A Baïkal setup may break when not backing up `baikal.yaml` due to database encryption not matching. The official [upgrading guide](https://sabre.io/baikal/upgrade/) recommends:

> Keep the whole Specific and config directories intact. Baïkal needs all files, not just the database.


nichtmax:
Thank you!

PR #1691, opened 2023-12-15
[linkace] Update guide to PHP 8.2

### LINKED ISSUE #1657: [linkace] Check for new PHP version compatibility
PHP 8.0 is EOL and wont soon be available on Uberspace any longer.

Check if this guide works with PHP 8.2 (or at least 8.1) and update it accordingly. Otherwise, if you cannot fix it with the new versions mark the guide as broken like it is described here: https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md

maltekrupa:
Works as expected.

Closes #1657

### DIFF
--- a/source/guide_linkace.rst
+++ b/source/guide_linkace.rst
@@ -29,13 +29,13 @@ LinkAce
 Prerequisites
 =============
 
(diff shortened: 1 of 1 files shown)


nichtmax:

PR #1702, opened 2023-12-28
Fix(NOISSUE): Remove mentioning "nginx" from flask guide

FalkoFlessner:
The flask guide mentions "Configure nginx". Since the user doesn't interact with nginx directly (I'm not even sure if Uberspace uses nginx, IIRC it's Apache), this was confusing to me.

I propose a change to "Uberspace web backend", as this phrasing is used in the Gitea documentation (which I found super helpful).

### DIFF
--- a/source/guide_flask.rst
+++ b/source/guide_flask.rst
@@ -202,7 +202,7 @@ Now let's start the service:
 
 
 
-Setup nginx
+Uberspace web backend
 -----------
 
 .. note::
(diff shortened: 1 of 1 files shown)


source/guide_flask.rst
```diff
@@ -202,7 +202,7 @@ Now let's start the service:
 
 
 
-Setup nginx
+Uberspace web backend
 -----------
```
FriedrichFroebel, line 206: The heading marker line should usually match the length of the actual heading.

FalkoFlessner: Yep, my bad. (Not used to `.rst` files, sorry)

Fixed.

PR #1711, opened 2024-01-18
Add guide for tor onion service

expiringplatform:
Hello everyone, 
I wrote a little guide how to install tor and setup an onion service in your uberspace.
I hope it might be helpful. :)
I'm very new to the used .rst format, so probably I made some errors in the style/formatting. I'm looking forward to any feedback! :)

### DIFF
--- a/source/guide_onionservice.rst
+++ b/source/guide_onionservice.rst
@@ -135,8 +135,8 @@ Users connecting to the onion service will look to the service that listens on t
 
 Read the tor FAQ_ and the tor browser manual_ to understand the security it can provide, the limitations and how to use it correctly:
 
-_FAQ: https://support.torproject.org/faq/
-_manual: https://tb-manual.torproject.org/
+.. _FAQ: https://support.torproject.org/faq/
+.. _manual: https://tb-manual.torproject.org/
 
 The folder ``~/onionservice`` contains the cryptographic keys of the onion service, which are critical for the security. Make sure to set restrictive permissions.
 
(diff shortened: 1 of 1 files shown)


nichtmax:
Please add 
```
ExitPolicy reject *:*
```
to the config, we don't want any exit nodes.

nichtmax:
I can't start the binary, I get the following error: 
```
./tor/tor: error while loading shared libraries: libevent-2.1.so.7: cannot open shared object file: No such file or directory
```

nichtmax (changes requested):
I can't get Tor so start, do you have a hint maybe?

expiringplatform:
> I can't start the binary, I get the following error:
> 
> ```
> ./tor/tor: error while loading shared libraries: libevent-2.1.so.7: cannot open shared object file: No such file or directory
> ```

Hi, thank you for the feedback and sorry for the late reply. 

@nichtmax  This error is expected. Tor dependends on some libraries that are shipped by torproject together with the tor binary. The start the tor binary the LD_LIBRARY_PATH has to be set accordingly. You can try:

`LD_LIBRARY_PATH=~/tor ~/tor/tor --torrc-file tor/torrc`

This command in also contained in the bottom of the guide in the Debugging section. I see that it might be confusing to people who follow the guide and then recognize that the binary can not be started. Do you think it would be useful to mention this command somewhere more prominent, like right before the "Setup the daemon" section. 

@SalocinHB Thanks for adding the exit policy to the config. I omitted it because I without checking assumed that this would be the implicit default when this option is not set in the config file.

nichtmax:
I apologize that it has taken so long. I had an exceptional personal situation and have hardly been able to get anything done in the last few weeks. 

I added a note about `LD_LIBRARY_PATH`.

expiringplatform:
Thanks for adding the explanation, that makes it way more clear what is happening.
No problem at all, don't worry. I just hope you are well. Great that this got merged, thanks! 🥳

PR #1717, opened 2024-01-24
New guide for PartDB-server (resolves #1514)

### LINKED ISSUE #1514: [Part-DB-server] New Guide for Part-DB-Server or mixin with PartDB Guide
[PartDB-Server](https://github.com/Part-DB/Part-DB-server) is the new symfony based version of PartDB, which has already a guide in the [Uberlab](https://lab.uberspace.de/guide_part-db/).

The new version offers several new features so it might be a good idea to add a second guide or update the current one.

You can also upgrade from an PartDB (legacy) installation - which i did on my uberspace.
Therefore you can basically follow the [upgrade guide](https://docs.part-db.de/upgrade_legacy.html)

Currently i'm not able to work out a full text guide, maybe i'll do it in the future, but i will already note the (few) caveats here, if your are nosy.

### TL;DR

You can basically run the Part-DB-server by their installation instruction.

You only have to replace the `yarn install` command with `node --max_old_space_size=800 ./node_modules/webpack/bin/webpack.js `

### Step-by-Step in condensed version for Uberspace

* Download and unpack
 `curl -L https://github.com/Part-DB/Part-DB-server/archive/refs/tags/v1.2.0.tar.gz | tar -xzf - --strip-components=1`
* Copy configuration file
 `cp .env .env.local`
* Edit configuration file
  * Change  `DATABASE_URL` to
 `mysql://isabell:password@127.0.0.1:3306/isabell_partdb?serverVersion=10.6.12`
  The server version can be gathered by calling `mysql --version` (Left the current value here for better understanding).
  * Change `DEFAULT_LANG` as needed
  * Change `DEFAULT_URI` to `https://isabell.uber.space/`
  * optional set SMTP Server (`MAILER_DSN`)
   `smtp://isabell:password@host.uberspace.de:587`
  * Change `APP_SECRET` to a new value, e.g. the output of `hexdump -vn16 -e'4/4 "%08x" 1 "\n"' /dev/urandom`
  **I'm not sure if this is necessary**
* Run composer
`composer install -o --no-dev`
* Run yarn
`yarn install`
* Run build
`node --max_old_space_size=800 ./node_modules/webpack/bin/webpack.js`
Just running `yarn build` would not work because an uberspace has a fixed memory limit. The current memory value of `800` worked for me, but may differ, depending on the amount of apps running on your uberspace.
* Run `php bin/console doctrine:migrations:migrate`
* Change your domain symlink to `public/`
 This is also known from other symfony based apps
* Edit `public/.htaccess` file and change 
 `FollowSymLinks` to `SymLinksIfOwnerMatch`
 **There are two occurrences in the file, keep in mind you change at least the one that is not commented out**
* Recommendation: Limit access to installation by using .htaccess/authentication

sunflower-seed:


SalocinHB:
I've checked your guide, it seems to work – even without the specification of the MySQL server version. Unless there is a specific reason to include it, consider removing that section to make the guide a little more accessible.

SalocinHB (approved):

sunflower-seed:
I removed the MySQL version as it seems to work with current Symfony/Doctrine and Uberspace MySQL setup

PR #1719, opened 2024-01-26
[oTree] Create guide (resolve #1718)

### LINKED ISSUE #1718: [oTree] Create guide
[oTree](https://www.otree.org/) is an open-source platform for web-based interactive tasks, such as lab experiments and multiplayer games, written in Python.
The oTree documentation recommends deployment via Heroku for simplicity, however this may no longer be the best option
for some users since the removal of free plans from the platform.

mirkoreul:
Hope this is useful and matches your style guidelines, happy to make adjustments if not.


SalocinHB:
The guide doesn't work in its current form because the Python venv isn't activated. Using a starter bash script to do this should work, but there might also be a more elegant solution.

SalocinHB:
@mirkoreul thanks for your guide. Please contact hallo@uberspace.de to receive a voucher for your goodie.

PR #1725, opened 2024-02-17
Update guide_passbolt.rst

foxytocin:
Updates guide to work with the latest passbolt version 4.5.2.. Testet with Uberspace 7.15.10.

- fixes GPG-Key generation (headless-mode)
- corrects parameters for config/passbolt.php
  - allows full security
  - allows sending emails using uberspace


kimdiallo:
Everything works as described, thank you very much  @foxytocin

PR #1726, opened 2024-02-24
Update guide_linkstack.rst

BabyFnord:
Some minor but vital changes here: The database port needs to be localhost according to the Uberspace MySQL page, hence, the prior port number returned an error. LinkStack supports PHP 8.3, which is now included in this manual. Maybe more importantly, I have included to edit .htaccess in the likely use case of access through a subdomain. Lastly, the ambiguous registration section now recommends to set a secure admin registration right away, where it was formerly recommended to do so in a(n unnecessary) second step.

### DIFF
--- a/source/guide_linkstack.rst
+++ b/source/guide_linkstack.rst
@@ -1,6 +1,7 @@
 .. highlight:: console
 
 .. author:: Jan Klomp <https://jan.klomp.de>
+.. author:: BabyFn0rd <https://github.com/BabyFnord/>
 
 .. categorize your guide! refer to the current list of tags: https://lab.uberspace.de/tags
 .. tag:: lang-php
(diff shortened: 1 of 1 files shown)


nichtmax:
Thank you for your PR! One small thing: Our [GUIDE](https://github.com/Uberspace/lab/blob/main/STYLE.md) states:

BabyFnord:
> One small thing: Our [GUIDE](https://github.com/Uberspace/lab/blob/main/STYLE.md) states:
> 
> > Don't mention additional document roots. Keep it simple. Don't use subfolders. Always use the standard document root ~/html. Always assume the document root is empty.

Thanks for the hint. I would like to add to that, though, that while adhering to the Guide ensures consistency across the manuals and as such is hardly debatable, the supposedly most likely use case is a subdomain. Unless someone intends to dedicate one uberspace and single domain to LinkStack (which I doubt most of us would). 

BTW, thanks for providing these excellent manuals! 💯

nichtmax:
> Unless someone intends to dedicate one uberspace and single domain exclusively to LinkStack (which I doubt most of us would).

That's what we want you to do. When you make a new account with us now you'll get an "umbrella account" and can add as many Asteroids as you want and set the price accordingly - as long as you pay at least 5€ for all Asteroids.

PR #1732, opened 2024-03-10
Update guide_ntfy.rst - new release link

Matthias84:
The scema is a bit different now and 2.x should be suggested?


SalocinHB:
covered by #1729

PR #1734, opened 2024-03-10
Update Pretix

chriskbach:
Since pretix 2023 PostgreSQL is required as a database. I rewrote and updated the guide


SalocinHB:
@chriskbach Thank you for the update. To claim your goodie, please contact hallo@uberspace.de.

PR #1737, opened 2024-03-17
Use version 3.5.* of pre-commit to avoid InvalidManifestError

### LINKED ISSUE #1733: `InvalidManifestError` when doing a git commit
I checked out the repo as usual and followed the setup step `make setup`. If I then execute `git commit -m "my commit message"`. I receive following error from git:

```
An error has occurred: InvalidManifestError:
==> File /Users/marvin/.cache/pre-commit/repoyg3plrqf/.pre-commit-hooks.yaml
==> At Hook(id='black')
==> At key: stages
==> At index 0
=====> Expected one of commit, commit-msg, manual, merge-commit, post-checkout, post-commit, post-merge, prepare-commit-msg, push but got: 'pre-commit'
```

I'm running MacOS 14.2.1 with git version 2.43.0.

Is it something that I did wrong or just an outdated script?

MetroMarv:
Upgrade pre-commit from version 2.13.* to 3.5.* to avoid the following error:

An error has occurred: InvalidManifestError:
==> File /Users/<user>/.cache/pre-commit/repoyg3plrqf/.pre-commit-hooks.yaml ==> At Hook(id='black')
==> At key: stages
==> At index 0
=====> Expected one of commit, commit-msg, manual, merge-commit, post-checkout, post-commit, post-merge, prepare-commit-msg, push but got: 'pre-commit'

Fixes #1733

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 -e ./authorship
 feedgen==0.9.0
-pre-commit==3.6.*
+pre-commit==3.5.*
 Sphinx==4.4.0
 sphinx-autobuild==2021.3.*
 sphinx-notfound-page==0.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 -e ./authorship
 feedgen==0.9.0
-pre-commit==2.13.*
+pre-commit==3.6.*
 Sphinx==4.4.0
 sphinx-autobuild==2021.3.*
 sphinx-notfound-page==0.3


MetroMarv:
I also tried to use the latest version of `pre-commit` (which is 3.6.*) but then the initialization of the Netlify checks fails.

FriedrichFroebel:
The failure is rather easy to explain. In the logs we have

> 7:05:49 PM: Python version set to 3.8

while https://github.com/pre-commit/pre-commit/blob/main/CHANGELOG.md#360---2023-12-09 states that 3.6.0 dropped support for Python 3.8 before the actual EOL.

MetroMarv:
That makes sense. Thanks for the clarification.

nichtmax:
Thank you!

Issue #1746, opened 2024-04-11
`make setup` fails due to error in pre-commit setup (InvalidManifestError)


bliepp:
Hello there,

on my machine I get an error when running `make setup`. Some investigation showed that `setup-venv` finished successfully while `setup-pre-commit` fails with an `InvalidManifestError`. Full error message:

```
[... venv setup and package inatallation ...]
Successfully installed [... alll the packages ...]
.venv/bin/pre-commit install --overwrite --install-hooks
pre-commit installed at .git/hooks/pre-commit
An error has occurred: InvalidManifestError: 
==> File /home/bliepp/.cache/pre-commit/repo9cqtpufl/.pre-commit-hooks.yaml
==> At Hook(id='black')
==> At key: stages
==> At index 0
=====> Expected one of commit, commit-msg, manual, merge-commit, post-checkout, post-commit, post-merge, prepare-commit-msg, push but got: 'pre-commit'
Check the log at /home/bliepp/.cache/pre-commit/pre-commit.log
make: *** [Makefile:28: setup-pre-commit] Fehler 1
```

I'm using Manjaro Linux and tried with Python3.8 and 3.11.

Any ideas?

Issue #1745, opened 2024-04-11
[pocketbase] add guide


bliepp:
Pocketbase (https://pocketbase.io/) is a selfhosted open source BaaS similar to Supabase or Firebase with a focus on minimalism for small to medium sized backend needs. I'll start working on this.

PR #1747, opened 2024-04-11
Add guide for Pocketbase

bliepp:
Added a guide for Pocketbase (see #1745). I'm sorry, but I had to commit using `--no-verify` due to the error described in #1746.


source/guide_pocketbase.rst
```diff
@@ -0,0 +1,162 @@
...
+
+
+Setup Supervisord
+=================
+
+First create a folder where all data is stored:
+
+.. code-block:: console
+
+    [isabell@stardus ~]$ mkdir /home/isabell/pocketbase
```
nichtmax, line 84: `mkdir ~/pocketbase` does the trick, no need for "isabell" here.

source/guide_pocketbase.rst
```diff
@@ -0,0 +1,162 @@
...
+    [isabell@stardus ~]$
+
+
+.. warning::
+    To modify this specific running instance of Pocketbase (e.g. updating, user management),
+    all ``pocketbase`` commands must be executed in the folder you just created!
+
+
+Now, create a configuration file ``/home/isabell/etc/services.d/pocketbase.ini``:
```
nichtmax, line 93: same here, please use `~/etc/...`

source/guide_pocketbase.rst
```diff
@@ -0,0 +1,162 @@
...
+
+    [program:pocketbase]
+    directory=%(ENV_HOME)s/pocketbase
+    command=pocketbase serve --http :<port>
+    autostart=yes
+    autorestart=yes
+    startsecs=60
+
+.. include:: includes/supervisord.rst
+
```
nichtmax, line 105: please add a note here with the port pocketbase uses like `pocketbase is running on port xxxx.`, see other guides for examples

source/guide_pocketbase.rst
```diff
@@ -0,0 +1,162 @@
...
+Updates
+=======
+
+.. note:: Check the update documentation_ or releases_ page regularly to stay informed about the newest version.
+
+To update the current executable you have two choices:
+
+* Repeat the steps in the *Installation* section above.
+* Run the built-in updater using ``pocketbase update`` inside ``/home/isabell/pocketbase``
```
nichtmax, line 129: replace `/home/isabell/` with `~/`

nichtmax (changes requested):
Thank you very much for your guide! Just a few small things ☺️

source/guide_pocketbase.rst
bliepp: Do you mean I that I should mention default port Pocketbase is running on? The thing is, Pocketbase has no default public port as it runs on 127.0.0.1:8090 by default and hence is not accessible to the public. You have to set an individual port explicitly with the `--http :<port>` flag to bind it via 0.0.0.0:<port>. So how should I do that? Saying `pocketbase is running on port 8090` seems kind of misleading, does it?

source/guide_pocketbase.rst
bliepp: Should I replace `/home/isabell` everywhere or only where you wrote the review comments?

nichtmax:
Thank you! Please get in touch via hallo@uberspace.de for your goodie!

PR #1760, opened 2024-05-07
Update guide_keycloak.rst

christianstrauch:
Updated for Keycloak 24.0.3

### DIFF
--- a/source/guide_keycloak.rst
+++ b/source/guide_keycloak.rst
@@ -1,6 +1,6 @@
 .. highlight:: console
 
-.. author:: Julius Künzel
+.. author:: Christian Strauch
 
 .. tag:: web
 .. tag:: sso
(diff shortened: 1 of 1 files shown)


source/guide_keycloak.rst
```diff
@@ -1,6 +1,6 @@
 .. highlight:: console
 
-.. author:: Julius Künzel
+.. author:: Christian Strauch
```
SalocinHB, line 3: is there any reason why you're removing the credit of the original guide's author here?

christianstrauch: Oh, absolutely not - apologies. Let's add him back in.

source/guide_keycloak.rst
```diff
@@ -143,144 +160,38 @@ Setup Backend
...
-Once you made sure had success and everything works as expected you can delete the previous version:
-
-::
-
- [isabell@stardust ~]$ rm -r keycloak-12.0.0
- [isabell@stardust ~]$
-
-
-
-.. _Keycloak: https://www.keycloak.org/downloads
```
SalocinHB, line 275: removing these lines breaks all external links

source/guide_keycloak.rst
```diff
@@ -143,144 +160,38 @@ Setup Backend
...
- [isabell@stardust ~]$
-
-And kill the process. Replace ``PID`` by the PID identified in the previous step.
-
-::
-
- [isabell@stardust ~]$ kill -SIGKILL PID
- [isabell@stardust ~]$
-
-Run the upgrade script
```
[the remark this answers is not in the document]
SalocinHB: Could you please provide some feedback to these questions? Otherwise, we'll close this PR next month.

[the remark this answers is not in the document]
christianstrauch: I had removed everything that, in its former form, does not apply to the newest Keycloak version any more. We need to completely rewrite the update guide. So far, my intention was to ensure the installation steps are properly reflected as they had to change quite dramatically, and that everything that would break any new installation is removed.

source/guide_keycloak.rst
SalocinHB: @christianstrauch since you didn't react to this comment, I'm assuming there was no reason to remove these lines. I've re-added them to resolve the broken links.

nichtmax:
https://github.com/Uberspace/lab/pull/1859

2martens:
Even with that configuration it does not seem to work well. Some endpoints require HTTPs in the 26 version but on Uberspace only HTTP is passed through.

Would be great to be able to pass the HTTPS traffic through to the backend to have it terminate it. The certificates are available after all.

PR #1766, opened 2024-05-20
Update2 guide_wiki-js.rst

punk-t:
Update guide_wiki-js.rst - Update script enhancements
  - variable supervisor service name
  - Version number conversion
  - GitHub URL update


nichtmax:
Hey @punk-t, thanks for your PR! To be honest I don't see the advantage for our users in your approach, that's a question of personal preference in my opinion and makes things more complicated for the average user.

PR #1770, opened 2024-06-03
Update guide_forgejo.rst. Fixes #1735

### LINKED ISSUE #1735: [forgejo] update info about ssh management
it looks like forgejo can and does already manage the ssh keys in `~/.ssh` without adding a symlink there. This might be new in an updated version, however the info in the guide should be updated because you might be able to 
break your ssh access if you add the same key without the fix﻿

toddy15:
I've tested this with Forgejo 7.0.3 and updated the guide. It's no longer necessary to link the .ssh directory, see #1735.

Moreover, I've moved the section about SSH key setup from the _optional_ configuration to the _needed_ configuration, because I think there is a real risk to lock oneself out of Uberspace. This can be fixed, of course, but I think that this is a major pain point for newcomers.

### DIFF
--- a/source/guide_forgejo.rst
+++ b/source/guide_forgejo.rst
@@ -22,7 +22,7 @@ Forgejo
 .. tag_list::
 
 Forgejo_ is a self-hosted Git service with a functionality similar to GitHub, GitLab and BitBucket.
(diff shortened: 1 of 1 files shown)


nichtmax:
Sorry for the long wait and thanks for your PR!

PR #1773, opened 2024-06-08
Update guide_resilio.rst to 2.8.1

rikroe:
Resilio Sync recently pushed an update (so they are not dead, yay!).

Download URLs have changed, so these needed adjustments.


nichtmax:
Thanks!

PR #1777, opened 2024-07-05
Update php url in guide_moodle.rst

mickeyhub:
The PHP link wasn't working as it was pointing to php-lang, but the actual URL is lang-php.


nichtmax:
Thank you!

PR #1778, opened 2024-07-05
Update guide_moodle.rst by describing a few bash commands

mickeyhub:
I've added the console steps to take for adding the input vars max to 5000.


nichtmax:
Please see our [Style guide](https://github.com/Uberspace/lab/blob/main/STYLE.md): 
> If there are files to edit, don't do stuff like cat > ~/.npmrc <<__EOF__, just tell the user to edit the file. Don't mention an editor like vi or nano.

mickeyhub:
@nichtmax my bad. Thank you for providing me with this feedback :)

PR #1780, opened 2024-07-05
Moodle guide seems to be broken

mickeyhub:
The guide for Moodle does not work anymore. This is due to not being able to use PHP 8.0.


luto:
#1784

PR #1784, opened 2024-07-10
[moodle] Update guide to use Moodle 4.4.1 and PHP 8.3

### LINKED ISSUE #1659: [moodle] Check for new PHP version compatibility
PHP 8.0 is EOL and wont soon be available on Uberspace any longer.

Check if this guide works with PHP 8.2 (or at least 8.1) and update it accordingly. Otherwise, if you cannot fix it with the new versions mark the guide as broken like it is described here: https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md

### LINKED ISSUE #1779: [moodle] guide is broken due to lack of correct PHP support
The guide for Moodle does not work anymore. This is due to not being able to use PHP 8.0, as suggested in the guide. We added a banner to the guide to signal the broken state.

If you have any insights, please comment on this issue. Pull requests fixing the guide are also highly appreciated. If the guide does not get fixed within two months, we will remove it from the lab, as documented in the [broken guides policy](https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md).

Not able to tag the author @direnc 

## Details:

I just tried to take all the [Moodle Lab steps](https://lab.uberspace.de/guide_moodle/).
In those steps, there isn't a "use PHP 8.0", step.
Trying this myself `uberspace tools version use php 8.0` did not work, as I received the reply `PHP version 8.0 does not exist.`

Not changing the PHP version, gives several errors in the first **initialize** step, opening the domain:
- Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /var/www/virtual/mick/html/lib/phpminimumversionlib.php on line 56
- Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /var/www/virtual/mick/html/lib/phpminimumversionlib.php on line 56
- Deprecated: Return type of moodle_recordset::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/virtual/mick/html/lib/dml/moodle_recordset.php on line 61
- Fatal error: Method lang_string::__set_state() must take exactly 1 argument in /var/www/virtual/mick/html/lib/moodlelib.php on line 10380

I tried running it, after setting my PHP version to 8.1, but I still received the following errors:
- Deprecated: Return type of moodle_recordset::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/virtual/mick/html/lib/dml/moodle_recordset.php on line 61
- Fatal error: Method lang_string::__set_state() must take exactly 1 argument in /var/www/virtual/mick/html/lib/moodlelib.php on line 10380

ihmels:
Fixes #1779 and fixes #1659, replaces #1780.

### DIFF
--- a/source/guide_moodle.rst
+++ b/source/guide_moodle.rst
@@ -12,44 +12,44 @@
 .. tag:: self-hosting
 
 
-.. sidebar:: About
+.. sidebar:: Logo
 
   .. image:: _static/images/moodle.svg
       :align: center
 
-##########
+######
 Moodle
-##########
+######
 
 .. tag_list::
 
-Moodle_ Moodle is an acronym for "Modular Object-Oriented Dynamic Learning Environment." It is a free and open-source learning management system (LMS) written in PHP and distributed under the GNU General Public License. More information on Wikipedia_ or the official Moodle_ page.
+Moodle_ Moodle is an acronym for "Modular Object-Oriented Dynamic Learning Environment". It is a free and open-source learning management system (LMS) written in PHP and distributed under the GNU General Public License. More information on Wikipedia_ or the official Moodle_ page.
 
 ----
 
 .. note:: For this guide you should be familiar with the basic concepts of
 
+  * :manual:`PHP <lang-php>`
   * :manual:`MySQL <database-mysql>`
   * :manual:`cron <daemons-cron>`
-  * :manual:`php <lang-php>`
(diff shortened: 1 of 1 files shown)


nichtmax:
Thanks for the update!

PR #1788, opened 2024-08-01
Add section "Migrate from gitea to forgejo"

stevleibelt:
* Change note about soft-fork to hard fork
* Add section `Migrate from gitea 1.22 to forgejo 8.0.0`

### DIFF
--- a/source/guide_forgejo.rst
+++ b/source/guide_forgejo.rst
@@ -22,7 +22,7 @@ Forgejo
 .. tag_list::
 
 Forgejo_ is a self-hosted Git service with a functionality similar to GitHub, GitLab and BitBucket.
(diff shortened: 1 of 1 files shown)


nichtmax:
Our guides are there to help you install apps and tools, not migrate from one to another. I'm not sure about your section to be honest. What's the advantage from forgejo over gitea from your perspective?

stevleibelt:
@nichtmax thanks for your response.

I was happy with the state of gitea as it was. Meaning it was feature complete for me.

The only reason for my change is the change in the gitea organisation. They changed to "open core" as written [here](https://lwn.net/Articles/963095/) and on the [official forgejo](https://forgejo.org/compare-to-gitea/) page.

Time will tell if bad things are happening on gitea and at the end of the day, your argument of "guides are there to install software" is fine for me to reject my pull request.

Thanks for your work so far

PR #1800, opened 2024-09-04
Add guide for screego

### LINKED ISSUE #1799: [screego] add guide
[screego](https://screego.net/) is an open-source software that provides screen sharing via WebRTC.

no-one:
Fixes #1799


no-one:
A question for the person who will review this pull request later: I have not currently entered a port range for the TURN server. There is the following example in the config:
``` EditorConfig
# Limit the ports that TURN will use for data relaying.
# Format: min:max
# Example:
#   50000:55000
SCREEGO_TURN_PORT_RANGE=
```
In most cases this will probably not be so relevant (because STUN can be used instead of TURN) but if this is not the case: As you can see, a fairly large range is specified. However, ports can only be opened individually. What do you suggest?

source/guide_screego.rst
```diff
@@ -0,0 +1,148 @@
...
+
+Configuration
+=============
+
+Configure Screego
+-----------------
+
+.. note:: You can use the following command to generate your own ``SCREEGO_SECRET``: ``tr -dc A-Za-z0-9 < /dev/urandom | head -c 40; echo``
+
+.. warning:: Replace ``isabell`` with your username, ``X940rRflEAWtVQvtPmCKgsj1WoJ8IoNZ2tEaDSxZ`` with your secret and ``40130``/``40131`` with your ports!
```
nichtmax, line 77: This looks like a real secret, people will just copy-paste it. How about `YourSecret`?

nichtmax: Check out https://lab.uberspace.de/guide_gitea/ for instance, there's a secret as well.

nichtmax (changes requested):
Thanks for your guide! I left you some comments, I'll test everything when these are resolved.

source/guide_screego.rst
no-one: You're right, I've changed it. However, I used “snake case” as in the Gitea guide.

source/guide_screego.rst
```diff
@@ -0,0 +1,157 @@
...
+
+If it's not in state RUNNING, check your configuration.
+
+Open website
+------------
+.. warning:: Replace ``isabell`` with your username and ``40130`` with your port!
+
+If Screego is running, you can find the website here:
+
+  * https://isabell.uber.space:40130/
```
nichtmax, line 128: Is there a reason you're not configuring [a backend](https://manual.uberspace.de/web-backends/)? Their docs say:
> If TLS should be enabled for HTTP requests. Screego requires TLS,
> you either have to enable this setting or serve TLS via a reverse proxy.

So I would just disable TLS here, you then only have to open one port and serve the webinterface via backend. This solves another problem: On a fresh asteroid there is no certificate until the first request, screego fails to start in this case.

nichtmax (changes requested):
One small thing left ☺️

source/guide_screego.rst
no-one: You're right, I've changed it.

nichtmax:
Thanks! Please get in touch via hallo@uberspace.de for your goodie!

PR #1804, opened 2024-09-07
add update method to FreshRSS

gedankenstuecke:
FreshRSS (now?) also allows updates via the web interface, which will automatically get the latest updates, without the need to manually download & replace the files. I've just done this on my space and it worked w/o problems. 

I've added that bit to the Upgrades section as an alternative.


nichtmax:
Thank you!

PR #1811, opened 2024-09-26
Add guide for Freelynx

escuco:
Feedlynx collects weblinks and generates a RSS feed with these links for your prefered RSS reader to read them later. Every time you transfer a link to *Feedlynx*, it fetches the URL to determine a title and description. With this information, *Feedlynx* generates a new entry to the feed.


nichtmax:
Thank you very much! I removed one line which I think was an accident, otherwise that's a very nice guide! Please get in touch via hallo@uberspace.de for your goodie!

PR #1815, opened 2024-10-09
[ssh-agent]: add guide on how to use on servers

kmoschcau:
This adds a guide on how to use `ssh-agent` on the uberspace servers.

I also added a background explanation on why this is necessary and different compared to most guides one finds on the net. Though I'm not sure myself if the background explanation should stay in.


kimdiallo:
@kmoschcau Thank you for the effort, we have sent you an email about your well-deserved goodie.

PR #1740, opened 2024-03-17
Add specific dependency versions to `requirements.txt`

MetroMarv:
Without the explicit declaration of this dependencies pip install incompatible versions for Sphinx 4.4.

Added the Enchant library as a prerequisite and that Python version must be less or equal to 3.11. As 3.12 doesn't support the `imp` library anymore ([0]), which is used by Sphinx 4.4.

Fixes #1739

[0]: https://docs.python.org/3.11/library/imp.html

PR #1817, opened 2024-10-14
Update Python and PyPI packages, fix errors and warnings

toddy15:
Hi Uberspace team,

I have Python 3.12 installed on my computer, so I wasn't able anymore to build the lab locally, because the `imp` library is not supported (see https://github.com/Uberspace/lab/pull/1740#issue-2190844007).

As the Sphinx version (4.4) in use is rather old, I've looked into updating Sphinx to the current version (8.0) and also updated the dependencies to their latest versions.

While at it, I fixed all reported errors and warnings (which are reported with Sphinx 4.4 as well, so those are not due to the update).

I'm able to build the lab locally now, without any warnings or errors. The generated HTML output looks good to me, I didn't spot any obvious errors or wrong formats in the output. However, I didn't verify *all* HTML files, just picked some examples.

From what I can tell, the minimum supported Python version seems to be 3.10 now, so if you want to loosen the requirement on Python 3.12 a bit, I'm happy to make that adjustment.

Regards,
Tobias


nichtmax:
The Netlify Build image only has Python 3.8 (thats why the build fails), we'll haven to throw some time at this.

nichtmax:
@luto any idea besides installing our own Python with every build?

toddy15:
Hm, I guess the move away from Netlify won't happen anytime soon. I've therefore created another PR (#1823), which updates the tooling up to the latest Sphinx version supporting Python 3.8. Moreover, all of the fixes for errors and warnings from this PR have been included in the new PR.

toddy15:
After #1823 has been merged, this PR can be closed. There are only the first two commits left, which currently are not working with Netlify.

Regards,
Tobias

PR #1820, opened 2024-10-23
fix actual guide, include bcrypt downgrade

### LINKED ISSUE #1802: [actual] guide broken
actual is broken with versions v24.7.0 and higher due to incompatible dependency on bcrypt v5.1.1

wysiib:
Resolves #1802

Actual in its current version depends on a too new version of bcrypt not compatible with Uberspace v7. Can still be used by downgrading the bcrypt dependency.

### DIFF
--- a/source/guide_actual.rst
+++ b/source/guide_actual.rst
@@ -17,13 +17,6 @@ Actual
 
 .. tag_list::
 
-.. error::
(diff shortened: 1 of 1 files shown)


nichtmax:
Thank you!

PR #1827, opened 2024-10-31
Add guide for Readeck

### LINKED ISSUE #1826: [Readeck] installation guide
if anyone could add an installation guide of [Readeck](https://readeck.org/en/docs/), an alternative to Wallabag and Omnivore, that would be fantastic!

no-one:
fixes #1826


kimdiallo:
@no-one First off, many thanks for your effort with the guide. 

I’ve been through it twice now, and initially, everything seems to work as described. But when I try to launch Readeck and finish the installation via ‘/onboarding’, absolutely nothing happens. I even attached strace to the process—really, nothing happens.

Any ideas on this? Did you perhaps not test it on a fresh Asteroid, or did you take any extra steps during your installation that you left out of the guide to keep things simple?

no-one:
Hey @kimdiallo,
please take a look at ``readeck@triangulum``. I've just reset this Uberspace (which only exists for this guide) and followed the guide. Everything looks good so far. In the home directory is the file "login_data" if you need access to Readeck. In the ``.bash_history`` you can see what I have done. As far as I understood, @crendl had no problems either (see #1826). Can you try again with this Uberspace? If that doesn't work either we'll have to figure out what we're doing differently.

SalocinHB:
I couldn't get it to work following the guide, either, nor was I able to log in to the instance at readeck.uber.space. The problem is that readeck by default expects to be served via HTTP. This leads to error messages in the browser console because my (and apparently @kim's as well) browser refuses to load HTTP components for a site that was accessed using HTTTPS. 

```
[Warning] [blocked] The page at https://readeck.uber.space/login?r=%2F requested insecure content from http://readeck.uber.space/logger/csp-report. This content was blocked and must be served over HTTPS. (content-mesh.js, line 6)

[Warning] Parsing application manifest https://readeck.uber.space/manifest.webmanifest: The start_url's origin of "http://readeck.uber.space" is different from the document's origin of "https://readeck.uber.space".
```

Adding `use_x_forwarded_proto = true` to the config solves the issues.

SalocinHB:
@no-one thank you for your guide. Please contact hallo@uberspace.de to receive your thank-you-goodie.

no-one:
@SalocinHB You're right, Chromium-based browsers don't work without it but Firefox does. Thank you!

Issue #1606, opened 2023-10-04
[umami] check for new nodejs version compatibility


noave:
Nodejs 12 / 14 / 16 are EOL and wont be available on Uberspace any longer.

Check if this guide works with Nodejs18 or Nodejs20 and update it accordingly. Otherwise, if you cannot fix it with the new versions mark the guide as broken like it is described here: https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md 

_(This issue has been created automatically, the guide might already be up to date, in this case you can just comment and close the issue here)_

josephbadow:
I read your [Tröt](https://chaos.social/@ubernauten@uberspace.social/111460487142843520) and decided to give umami a try.

### Node 18
I didn't encounter any real issues. Running `yarn build` resulted in an error, but this was already mentioned in your guide.

```yarn build

[...]

 ✓ Compiled successfully
   Skipping validation of types
   Skipping linting
uncaughtException Error: write EPIPE
    at target._send (node:internal/child_process:872:20)
    at target.send (node:internal/child_process:745:19)
    at ChildProcessWorker.initialize (/home/jwb/umami/node_modules/next/dist/compiled/jest-worker/index.js:1:11409)
    at new ChildProcessWorker (/home/jwb/umami/node_modules/next/dist/compiled/jest-worker/index.js:1:10741)
    at WorkerPool.createWorker (/home/jwb/umami/node_modules/next/dist/compiled/jest-worker/index.js:1:6314)
    at new BaseWorkerPool (/home/jwb/umami/node_modules/next/dist/compiled/jest-worker/index.js:1:8173)
    at new WorkerPool (/home/jwb/umami/node_modules/next/dist/compiled/jest-worker/index.js:1:6110)
    at new Worker (/home/jwb/umami/node_modules/next/dist/compiled/jest-worker/index.js:1:23092)
    at createWorker (/home/jwb/umami/node_modules/next/dist/lib/worker.js:30:28)
    at new Worker (/home/jwb/umami/node_modules/next/dist/lib/worker.js:67:9)
    at createStaticWorker (/home/jwb/umami/node_modules/next/dist/build/index.js:732:24)
    at /home/jwb/umami/node_modules/next/dist/build/index.js:797:56
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Span.traceAsyncFn (/home/jwb/umami/node_modules/next/dist/trace/trace.js:105:20)
[...]

PR #1829, opened 2024-11-03
Update guide to work with current version 2.13.2

hannesherold:
Currently there seems to be a migration issue that leads to a fatal error during the build process. Reference: https://github.com/umami-software/umami/issues/2645#issuecomment-2038626975


hannesherold:
Possibly related: #1606

kimdiallo (approved):
works exactly as described - thank you very much!

PR #1832, opened 2024-11-12
[writefreely] pin version to 15.0

nichtmax:


hikerus:
Please add the version information also under the chapter "Update".

PR #1847, opened 2024-12-09
Create guide_streamlit.rst

entorb:
Based on Django guide, setup using non-root document root, as this way my use case.

### DIFF
--- a/source/guide_streamlit.rst
+++ b/source/guide_streamlit.rst
@@ -1,0 +1,185 @@
+.. highlight:: console
+
+.. author:: Torben <https://entorb.net/contact.php?origin=UberspaceLab>
+
+.. tag:: lang-python
+.. tag:: streamlit
+.. tag:: web
+.. tag:: audience-developers
+
+######
+Streamlit
+######
+
+.. tag_list::
+
+Streamlit_ is a high-level Python Web framework focussing on data visualization. This guide is based on the :lab:`Django <guide_django>` guide.
+
+----
+
+.. note:: For this guide, you should be familiar with the basic concepts of
+
+  * :manual:`Python <lang-python>` and its package manager pip
+  * :manual:`Supervisord <daemons-supervisord>` to set up your service
+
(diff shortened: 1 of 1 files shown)


entorb:
adjusted

source/guide_streamlit.rst
```diff
@@ -0,0 +1,178 @@
...
+.. code-block:: ini
+  :emphasize-lines: 2,3
+
+  [program:mysiteproject]
+  directory=%(ENV_HOME)s/mysiteproject
+  command=streamlit run app.py
+
+After creating the configuration, tell :manual:`supervisord <daemons-supervisord>` to refresh its configuration and start the service:
+
+::
```
nichtmax, line 123: Please use
```
.. include:: includes/supervisord.rst
```

source/guide_streamlit.rst
```diff
@@ -0,0 +1,178 @@
...
+.. tag:: web
+.. tag:: audience-developers
+
+######
+Streamlit
+######
+
+.. tag_list::
+
+Streamlit_ is a high-level Python Web framework focussing on data visualization. This guide is based on the :lab:`Django <guide_django>` guide.
```
nichtmax, line 16: I don't think it's based on the Django Guide. I gave you the Django Guide as an example for another Python based guide, no need to mention it here.

entorb: I used that file as a starting point, so quite some of the text in this guide are copied from there, so I find it important to quote the source.

source/guide_streamlit.rst
entorb: added

entorb:
I added your proposals. If more modifications are needed, feel free to directly apply them. I think that would be much faster than ping pong ;-)

nichtmax:
Please get in touch via hallo@uberspace.de for your goodie.

entorb:
Thanks for the finishing!

Issue #1234, opened 2022-03-25
Prosody install guide needs improvement for 0.12.0


hasmaneuroda:
After trying to upgrade to the latest version of prosody (0.12.0) the server is starting but it is not possible to connect to it.

prosodyctl about:
```

**************************
Prosody was unable to find lua-unbound
This package can be obtained in the following ways:

  Debian/Ubuntu | sudo apt install lua-unbound
       luarocks | luarocks install luaunbound
         Source | https://www.zash.se/luaunbound.html

Old DNS resolver library will be used
More help can be found on our website, at https://prosody.im/doc/depends
**************************

general             warn        Lua 5.1 has several issues and support is being phased out, consider upgrading
certmanager         error       SSL/TLS: Error initialising for client_https port 0: unknown elliptic curve in "X25519:prime256v1:secp384r1"
Prosody 0.12.0

# Prosody directories
Data directory:     /home/hasmaneuroda/var/lib/prosody
Config directory:   /home/hasmaneuroda/etc/prosody
Source directory:   /home/hasmaneuroda/lib/prosody
Plugin directories:
  /home/hasmaneuroda/var/lib/prosody/custom_plugins - not a directory!
  /home/hasmaneuroda/var/lib/prosody/prosody-modules - prosody-modules rev: a8e9949a6ad2
  /home/hasmaneuroda/lib/prosody/modules/


# Operating system
Linux 3.10.0-1160.53.1.el7.x86_64

# Lua environment
Lua version:                    Lua 5.1

Lua module search paths:
  /home/hasmaneuroda/lib/prosody/?.lua
  /home/hasmaneuroda/.luarocks/share/lua/5.1/?.lua
  /home/hasmaneuroda/.luarocks/share/lua/5.1/?/init.lua
  /usr/share/lua/5.1/?.lua
  /usr/share/lua/5.1/?/init.lua
  /usr/lib64/lua/5.1/?.lua
  /usr/lib64/lua/5.1/?/init.lua
  /home/hasmaneuroda/var/lib/prosody/custom_plugins/share/lua/5.1/?.lua
  /home/hasmaneuroda/var/lib/prosody/custom_plugins/share/lua/5.1/?/init.lua

Lua C module search paths:
  /home/hasmaneuroda/lib/prosody/?.so
  /home/hasmaneuroda/.luarocks/lib64/lua/5.1/?.so
  /usr/lib64/lua/5.1/?.so
  /usr/lib64/lua/5.1/loadall.so

[...]

PR #1853, opened 2025-01-05
Update guide_prosody.rst

hasmaneuroda:
Clarified how to handle the bug described in 
https://github.com/Uberspace/lab/issues/1234#issuecomment-2571435772

Removed coturn from Voice/Video Calls section as it doesn't work since ages and the coturn guide is removed from uberlab.  Replaced it with eturnal.

Added me and don-philipe as authors because don-philipe gave an easy solution to handling the bug and me, because I tested, wrote down and added the eturnal thing.


nichtmax:
Thank you very much!

PR #1854, opened 2025-01-08
Add guide for Concrete CMS

thomas-temp-git:


thomas-temp-git:
Included a guide for Admidio as well

SalocinHB:
> Included a guide for Admidio as well
please move that to a separate pull request.

thomas-temp-git (commented):
Removed Admidio guide

source/guide_concrete_cms.rst
```diff
@@ -0,0 +1,98 @@
...
+ [isabell@stardust html]$ unzip 9.3.8.zip
+ [isabell@stardust html]$ mv concrete-cms-9.3.8/* ./
+ [isabell@stardust html]$ rmdir concrete-cms-9.3.8
+ [isabell@stardust html]$ rm 9.3.8.zip
+ [isabell@stardust html]$
+
+Configuration
+=============
+
+Point your browser to your domain (e.g. isabell.uber.space) to set up and configure your Concrete CMS installation.
```
SalocinHB, line 66: Some step must be missing here. In the browser it says
> Third party libraries not installed. Make sure that composer has required libraries in the concrete/ directory.

thomas-temp-git: My bad, I updated the download link.

SalocinHB: You changed the download URL and also the version in that URL, but kept the version the same in the next steps. `unzip` won't find a 9.3.8.zip if you downloaded 9.3.9. Might be a good idea to give `wget` a target file name with the `-O` option.

```
$ wget https://github.com/concretecms/concretecms/releases/download/9.3.9/concrete-cms-9.3.9.zip -O concrete.zip
$ unzip concrete.zip
[…]
```

thomas-temp-git: It wouldn't work with the target file; unzipping would also include the version name and already contain a directory called "concrete." I just fixed the version numbers to keep it simple. ;) Using a variable is another option if you prefer.

PR #1857, opened 2025-01-20
Update guide_readeck.rst

pheraph:
log_level parameter changed


no-one (approved):
As the original author of the guide I can confirm that this change is necessary.

PR #1858, opened 2025-01-22
[keycloak] broken with new versions

nichtmax:


SalocinHB:
The guide for Keycloak does not work anymore. We added a banner to the guide to signal the broken state.

If you have any insights, please comment on this issue. Pull requests fixing the guide are also highly appreciated. If the guide does not get fixed within two months, we will remove it from the lab, as documented in the [broken guides policy](https://github.com/Uberspace/lab/blob/main/BROKEN_GUIDES.md).

@jlskuz FYI.

PR #1866, opened 2025-02-05
Add guide for blocky

thomas-temp-git:


nichtmax:
Thanks for your guide! There are some problems with the rendering, please check [the preview](https://deploy-preview-1866--uberlab.netlify.app/guide_blocky/).

SalocinHB:
Your default config leaves the debug web interface wide open for everyone to use. I don't think that's a good idea.

thomas-temp-git:
I didn't find an option in the configuration to disable the debug web interface. The API interface would be exposed too. A possible workaround would be:
`uberspace web backend set /dns-query --http --port 4000`
This would only allow DoH queries.

SalocinHB:
@thomas-temp-git I believe you already know the drill: Please contact hallo@uberspace.de to get your free goodie.

pictosun:
As far as I understand now blocky is an open DNS server for everyone (not only you) under the specified IP - don't know if Uberspace does like it? Isn't there any chance to make a bit more secure as it is possible with AdGuard Home and so on?

nichtmax:
> As far as I understand now blocky is an open DNS server for everyone (not only you) under the specified IP - don't know if Uberspace does like it? Isn't there any chance to make a bit more secure as it is possible with AdGuard Home and so on?

No problem as long as it's DoH queries.

PR #1868, opened 2025-02-09
[firefly-iii] update updates info

unite42:
Updated updates info to match the [official guide](https://docs.firefly-iii.org/how-to/firefly-iii/upgrade/self-managed/), because the "create-project" method is no longer recommended.


nichtmax:
Thank you!

PR #1874, opened 2025-02-28
Add guide for DatedMail

Weishaupt:
I've created a new PowerShell module which restores the functionality of expiring mail addresses comparable to `qdated` which I previously used since U6.

This time the functionality is based on sieve filters and should be also be compliant to the upcoming U8, as qmail is not involved in any way.

### DIFF
--- a/source/guide_datedmail.rst
+++ b/source/guide_datedmail.rst
@@ -34,7 +34,7 @@ All relevant legal information can be found here
 Prerequisites
 =============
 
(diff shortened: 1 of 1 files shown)


nichtmax:
Sorry for the long wait, this is the first time for all of us with powershell. I tested your guide, works like a charm. Please get in contact via hallo@uberspace.de for your goodie.

PR #1875, opened 2025-03-01
Update guide_firefly-iii.rst

matthias-mader:
change installation method to recommended method as per project installation guide; update post-installation commands


nichtmax:
Thank you!

Issue #1886, opened 2025-03-29
[automatic-deployment] Enhance guide with notes about rsync exclusions


mechanicalgirl:
Per the [contributor page](https://github.com/Uberspace/lab/blob/main/CONTRIBUTING.md), I am signaling my intent to add some detail about rsync exclusions to the `automatic-deployment` guide. :) 

PR to follow

PR #1890, opened 2025-04-09
[automatic-deployment] Enhance guide with notes about rsync exclusions

mechanicalgirl:
Per https://github.com/Uberspace/lab/issues/1886

This PR suggests a small addition to the Automatic Deployment guide, the inclusion of an example of the `--exclude` rsync switch. I think this should be a recommended part of the workflow to prevent large, unneeded files such as the .git object from being included in the deploy.


nichtmax:
Makes sense! Thank you!

Issue #1899, opened 2025-05-11
Laravel Installation Guide


FelixWienss:
Hello,

I was working on a new installation guide for Laravel since the old issue regarding Laravel never got into uberlab. Happy to hear your feedback!

![Image](https://github.com/user-attachments/assets/8645fce7-1053-4bb1-aafd-59ff78eed82d)

.. highlight:: console

.. author:: Felix Wienss <https://github.com/FelixWienss>

.. tag:: lang-php
.. tag:: web

.. sidebar:: Logo

  .. image:: _static/images/laravel-logo.png

          :align: center

#######
Laravel
#######

.. tag_list::

Laravel is a web application framework with expressive, elegant syntax. We’ve already laid the foundation — freeing you to create without sweating the small things.

----

.. note:: For this guide you should be familiar with the basic concepts of

  * :manual:`PHP <lang-php>`
  * :manual:`Domains <web-domains>`
  * :manual:`MySQL <database-mysql>`
  * :manual:`DocumentRoot <web-documentroot>`

License
=======

All relevant legal information can be found here

 * https://laravel.com/docs/12.x/license

Prerequisites
=============

We're using :manual:`PHP <lang-php>` version 8.3:

::

 [isabell@stardust ~]$ uberspace tools version show php
 Using 'PHP' version: '8.3'
 [isabell@stardust ~]$

Your web domain needs to be setup:

.. include:: includes/web-domain-list.rst

.. include:: includes/my-print-defaults.rst

Installation
============

``cd`` to your user dir, then get Laravel installer via Composer. Finally create a new project using the installer and chose your `starter kit <https://laravel.com/docs/12.x/starter-kits>`.

::

 [isabell@stardust ~]$ cd /var/www/virtual/$USER/
 [isabell@stardust html]$ composer global require "laravel/installer"
 [isabell@stardust html]$ laravel new my_project
 Application ready! Build something amazing.
 [isabell@stardust ~]$
 

Change the DocumentRoot
=======================

After the installation has finished, remove your unused :manual:`DocumentRoot <web-documentroot>` and create a new symbolic link to the ``my_project/public`` directory.

::

 [isabell@stardust ~]$ cd /var/www/virtual/$USER/
 [isabell@stardust isabell]$ rm -f html/nocontent.html; rmdir html
 [isabell@stardust isabell]$ ln -s /var/www/virtual/$USER/my_project/public html
 [isabell@stardust isabell]$

Config
======

Set Database up
###############

[...]

PR #1900, opened 2025-05-11
[Laravel] add guide for Laravel

FelixWienss:
This PR adds a guide for Laravel. Closes [#1899](https://github.com/Uberspace/lab/issues/1899)

If there are any changes needed just comment and I'm happy to work on it!


kimdiallo (changes requested):
First of all, please excuse the long time you have been waiting for this review and thank you very much for your effort up to here.

In principle, everything works as you have described it, but the first step of the installation seems a bit incomplete to me. Although you refer to laravel's documentation of the starter kits, the whole selection process is not mentioned in the guide and could be confusing, especially for less experienced users. 

I think going through a complete example would be the more accessible option here.
And the MySQL part is there twice. :)

PR #1903, opened 2025-05-23
Update guide_pretix.rst

albwe:
The static files copy command just copied the folder static.dist in the static folder, not the directories in this folder. I changed that with editing the path


SalocinHB:
`cp -r` is recursive, so it copies all files and folders within a copied folder. Did you accidentally omit the `-r` parameter?

PR #1919, opened 2025-07-27
Remove copypaste leftover

swsnr:
Remove a block about suggesting pip upgrade isso which looks as if it was copied from the isso page.


nichtmax:
Thank you!

PR #1929, opened 2025-08-17
Add several updates to the Dendrite guide

coderkun:
Update the upstream URL of Dendrite, which was moved to https://element-hq.github.io/dendrite/.

Fix commands to use the correct folder `~/dendrite` instead of `dendrite`.

Suggest database configuration to lower the amount of connections to reduce the resource consumption of PostgreSQL.

### DIFF
--- a/source/guide_dendrite.rst
+++ b/source/guide_dendrite.rst
@@ -196,6 +196,16 @@ your PostgreSQL setup:
 Replace ``key`` with your PostgreSQL database key (see the corresponding
 installation step).
 
+Additionally, it is recommended to lower the amount of database connections to
+reduce the resource consumption of PostgreSQL:
+
+.. code-block:: yaml
+ :emphasize-lines: 3
+
+    database:
+      connection_string: postgresql://dendrite:key@localhost/dendrite?sslmode=disable
+      max_open_conns: 10
+
 Logging
 -------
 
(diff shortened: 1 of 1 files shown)


source/guide_dendrite.rst
```diff
@@ -49,22 +49,22 @@ Check out a stable version (replace ``v0.13.5`` with the latest version):
 .. code-block:: console
 
   [isabell@stardust ~]$ cd ~/dendrite/
-  [isabell@stardust dendrite]$ git checkout v0.13.5
+  [isabell@stardust ~/dendrite]$ git checkout v0.13.5
```
SalocinHB, line 52: nope. Default bash config will only show the current folder name, not a full path here

source/guide_dendrite.rst
```diff
@@ -238,8 +250,8 @@ commandline:
 
 .. code-block:: console
 
-  [isabell@stardust dendrite]$ cd ~/dendrite/
-  [isabell@stardust dendrite]$ ./bin/create-account -config dendrite.yaml -username myuser
+  [isabell@stardust ~/dendrite]$ ./bin/create-account -config dendrite.yaml -username myuser
```
SalocinHB, line 253: keeping the `cd` doesn't hurt, does it?

source/guide_dendrite.rst
```diff
@@ -262,17 +275,18 @@ To update the code, fetch all git updates and checkout the latest version
 
 .. code-block:: console
 
-  [isabell@stardust ~]$ cd ~/dendrite/
-  [isabell@stardust dendrite]$ git fetch -p
-  [isabell@stardust dendrite]$ git checkout v0.13.5
+  [isabell@stardust ~/dendrite]$ git fetch -p
```
SalocinHB, line 278: keeping the `cd` doesn't hurt, does it?

source/guide_dendrite.rst
coderkun: Right, I removed this change.

source/guide_dendrite.rst
coderkun: Right, I added it back—and also added it to a few blocks to make it more consistent.

PR #1931, opened 2025-08-19
Update guide_forgejo.rst - mysql host

pschulz:
The hostname is necessary in order to make the database initialization work. Tested with version 12.0.1


thisven:
This adaption is also necessary when upgrading from version < 12.0.0.

PR #1935, opened 2025-08-30
[writefreely] Fix and clarify update section

thisven:
The command for copying the `keys` directory now doesn't create a subfolder structure (`keys/keys`), which prevented writefreely from starting. In addition, some clarification in between the commands of the update procedure are introduced.


escuco:
Thanks you!
Since Writefreely v0.16.0 works fine now without the glibc issue plus this uberlab documentation fix I was able to upgrade my installation.

PR #1947, opened 2025-10-17
Update guide_pretix.rst to include a section on redis

karlquint:
When checking the system health via the proposed GET request (see here: https://docs.pretix.eu/self-hosting/maintenance/#uptime-monitoring), it returned a message that "Cache was not available".  I checked the pretix docs and adapted the section on redis in the .pretix.cfg


karlquint:
@nichtmax the instructions for pretix in the [lab guide for pretix](https://lab.uberspace.de/guide_pretix/) already state that Redis needs to be installed according to the Redis guide.

PR #1953, opened 2025-11-08
Update guide_listmonk.rst

fahrplaner:
Setting admin user and password in config is not longer advised. Postgresql user can't contain spaces or install will break


SalocinHB:
> Postgresql user can't contain spaces

Here you mention the user, but your commit says spaces in passwords are problematic. Which is it?

PR #1983, opened 2026-02-09
Update guide_grafana.rst

Ryuno-Ki:
The tar arguments need a dash instead of a dot for options. Updated for Grafana 12.3.2


Ryuno-Ki:
With the new v13 the grafana configuration changed:

https://github.com/grafana/grafana/releases

Instead of `grafana-server` it's now `grafana server`.

Once this PR is merged, I'd submit a follow-up one. If you prefer to have it as part of this PR let me know.

PR #2004, opened 2026-03-30
Update cp command to copy all contents from static.dist

karlquint:
The previous command copied the entire static.dist/ folder into the static/ folder, which is not what was intended. 

The new cp command now correctly copies the **contents** of the static.dist/ folder into the static/ folder.


SalocinHB:
The existing command will copy the static.dist folder to html/static. If, and only if, html/static already exists, then yes, that will result in html/static/static.dist. Is there any reason html/static would already exist?

karlquint:
Yes, if you upgrade pretix from an earlier version. In that case, the html/static folder will already exist and its contents need to be updated. One could - of course - delete the entire html/static folder and recreate it from the static.dist folder. But it was not self-explanatory for me...

PR #2008, opened 2026-04-16
Changed command in services.d/filebrowser.ini 

karlquint:
Fixed the supervisor start command for filebrowser by using the absolute binary path instead of relying on PATH resolution. This ensures the service starts correctly under Uberspace supervisord.

### DIFF
--- a/source/guide_filebrowser.rst
+++ b/source/guide_filebrowser.rst
@@ -104,7 +104,7 @@ Create ``~/etc/services.d/filebrowser.ini`` with the following content:
 
  [program:filebrowser]
  directory=%(ENV_HOME)s/filebrowser
- command=filebrowser
+ command=%(ENV_HOME)s/filebrowser/filebrowser
  startsecs=30
  autostart=yes
 


EV21:
Hi 👋
- The filebrowser binary file is located at `~/bin/filebrowser`. (supervisord should look for commands in `~/bin/`)
 Did you miss the following commands?:
 ```
cd ~/bin
curl --location https://github.com/filebrowser/filebrowser/releases/latest/download/linux-amd64-filebrowser.tar.gz | tar -xzf - filebrowser
```
- Your fix would break the guide as there is no filebrowser binary at `~/filebrowser/filebrowser`.
- `~/filebrowser/` contains the config files.
FYI: `directory=%(ENV_HOME)s/filebrowser` sets the working directory so filebrowser can look there for config files.

karlquint:
Hello, thank you for pointing it out. I diverted from the suggested folder structure and put the binary in a different folder. Therefore it couldn’t be found by supervisord. I assumed I need to specify the location even when it is located in the bin/ folder, hence I suggested the change. 
The issue can be closed.

PR #2021, opened 2026-06-19
fix opcache path

### LINKED ISSUE #2020: [Wordpress] opcache.file_cache must be a full path of an accessible directory
/run/user/$UID/opcache will be gone after a reboot, since /run/user/$UID is a RAM-disk. This must be a different, permanent path, such as ~/tmp/opcache.

salocinhb:
closes #2020


salocinhb:
this is identical to https://codeberg.org/uberspace/lab/pulls/63

PR #2028, opened 2026-07-15
[CastMail2List] Add new guide

mxmehl:
This PR adds a new guide for [CastMail2List](https://github.com/mxmehl/castmail2list), a light mailing list system specifically designed for use with Uberspace.


salocinhb:
does it log details for why a mail couldn't be sent? I just tried it and it receives the mails I sent in, but when it tries to send them out to the other subscribers, that doesn't work and the log doesn't provide any details.

```
Level	ERROR
Event	email_out
Message	Failed to send email to address@example.org
```

mxmehl:
Thanks for the report. Indeed, the detailed SMTP logs have been only available via the console error log stderr. I've improved that in the most recent release.

